PR #507, opened 2014-11-29
add OpenSSL support for symmetric key encryption

terrafrost:
Replaces #259

There were a few things I was wanting to do before it was merged but that I've ultimately decided against.

1. rename CRYPT_DES_MODE_3CBC and CRYPT_DES_MODE_CBC3 to CRYPT_MODE_3CBC / CRYPT_MODE_CBC3? inner chaining and outer chaining are now being tested but as far as renaming the constants goes...  that'll break BC. maybe in the php5 / 2.0 release this can be done.
2. make it so Crypt_RC2 assumes the key size is the length of the key that was passed to it? I might make this a separate and unrelated PR. I also want to look at the past discussions on the subject because I think this has been previously discussed at https://github.com/phpseclib/phpseclib/pull/124
3. test both internal modes? i'm actually leaning towards not doing this. for one, so long as their are two internal modes it'd almost make more sense to have 2x internal mode engine constants. eg. CRYPT_ENGINE_MODE_INTERNAL_INLINE and CRYPT_ENGINE_MODE_INTERNAL. But with https://github.com/phpseclib/phpseclib/pull/591 we could just remove the second mode all together (see emails to petrich from 4/16/13). Also, altho calling `$this->_setupInlineCrypt()` every time `isValidEngine()` to see if it is setting `$this->use_inline_crypt` to false (which it is in Crypt/Base.php) instead of how it is currently called would result in inefficient memory usage. ie. `isValidEngine()` is called whenever the key is set or the IV. So if you set both it is called twice whereas it only needs to be called once.

With any luck this'll make #458 more testable (ie. I get the impression mcrypt is becoming supported less and less, as the years pass, whereas openssl is becoming more and more supported; for example, doing `sudo apt-get install php5-cli` on Ubuntu yields a PHP install with OpenSSL but not mcrypt; mcrypt is still built into Windows PHP installs whereas OpenSSL is not, but who knows how long that'll continue to be the case).

Also, the following might make for some interesting reading:

https://github.com/terrafrost/phpseclib/blob/f6e0c4b506b0bdb310314aabfca8d0b7244fe8ba/tests/Unit/Crypt/RC2.php#L17

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -4,3 +4,4 @@ imports:
 tools:
     external_code_coverage:
         runs: 5 # No Code Coverage on PHP 5.2 and HHVM
+        timeout: 2700 # 45 minutes
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -3,5 +3,5 @@ imports:
 
 tools:
     external_code_coverage:
-        runs: 5 # No Code Coverage on PHP 5.2 and HHVM
+        runs: 5 # No Code Coverage and HHVM
         timeout: 2700 # 45 minutes
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ env:
 before_install: true
 
 install:
+  - sudo apt-get install parallel
   - eval `ssh-agent -s`
   - travis/setup-secure-shell.sh
   - sh -c "if [ '$TRAVIS_PHP_VERSION' != 'hhvm' ]; then travis/install-php-extensions.sh; fi"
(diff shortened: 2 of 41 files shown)


terrafrost:
Why is this failing on Travis CI on PHP 5.2 and PHP 5.3.3? I don't see any errors in the output ugh...

bantu:
@terrafrost What is the status here. Is this just for testing?

terrafrost:
It's pretty much finished. Just need to start splitting it out into multiple PR's per this post:

> I know this is based on #259, but still: This patch is huge and does too many things at once and as a result is hard to review (if at all). I would have appreciated if parts from #259 were extracted into their own PRs. E.g. why not send a pull request only implementing _setEngine? OpenSSL can hardly be required for that.

I expect it to be a bit of a PITA, however. Maybe I'll start on that next weekend idk.

PR #584, opened 2014-12-27
Add callback to SFTP.php

DavidAnderson684:
Currently, Net_SCP::put() has a callback parameter, allowing the consumer to monitor/log/report progress (or whatever they please).

This patch adds the same feature, working in the same way, to Net_SFTP::put()


DavidAnderson684:
@bantu: This was also my reason for adding the callback, when I'd read the code. Apologies for being slow to feed back.

PR #590, opened 2014-12-29
Mark Base as abstract

cnelissen:
This PR marks Crypt\Base as abstract as well as the three methods that must be implemented in derived classes. This allows the PHP engine to throw proper errors if derived classes do not implement methods as required by the Base class.

### DIFF
--- a/phpseclib/Crypt/Base.php
+++ b/phpseclib/Crypt/Base.php
@@ -44,9 +44,8 @@ use phpseclib\Crypt\Hash;
  * @package Base
  * @author  Jim Wigginton <terrafrost@php.net>
  * @author  Hans-Juergen Petrich <petrich@tronic-media.com>
- * @access  public
  */
-class Base
+abstract class Base
 {
     /**#@+
      * @access public
@@ -1170,10 +1169,7 @@ class Base
      * @param String $in
      * @return String
      */
-    function _encryptBlock($in)
-    {
-        user_error(__METHOD__ . '() must extend by class ' . get_class($this), E_USER_ERROR);
-    }
+    abstract function _encryptBlock($in);
 
     /**
      * Decrypts a block
@@ -1184,10 +1180,7 @@ class Base
      * @param String $in
      * @return String
      */
-    function _decryptBlock($in)
-    {
-        user_error(__METHOD__ . '() must extend by class ' . get_class($this), E_USER_ERROR);
-    }
+    abstract function _decryptBlock($in);
 
     /**
      * Setup the key (expansion)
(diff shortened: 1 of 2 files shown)


phpseclib/Crypt/Base.php
```diff
@@ -44,9 +44,9 @@
  * @package Base
  * @author  Jim Wigginton <terrafrost@php.net>
  * @author  Hans-Juergen Petrich <petrich@tronic-media.com>
- * @access  public
+ * @abstract
```
bantu: Is that necessary? Seems redundant.

bantu: same below

GrahamCampbell: yeh, don't need that

GrahamCampbell: tbh, i think we should get rid of the package docblocks too. totally useless as far as i can see

bantu: I agree, but it's not related to this PR. :-)

GrahamCampbell: yeh

terrafrost: It's in the PEAR coding standards, which phpseclib is based off of:

http://pear.php.net/manual/en/standards.header.php

GrahamCampbell: how about we switch to psr-2?

terrafrost: For the php5 branch / 2.0 I'm okay with that.

cnelissen:
Any other issues with this PR?

bantu:
Do we want to start using inheritdoc (possibly overall and in a new PR)?

terrafrost:
Aside from what bantu said this PR looks good to me.

cnelissen:
> Do we want to start using inheritdoc (possibly overall and in a new PR)?

+1 for that. There is no reason to duplicate documentation...

bantu:
@cnelissen Do you know how much duplication there already is. Is this the only duplication?

cnelissen:
All of the constructers for classes that extend base are duplicated currently, and are nearly verbatim from Base...

Otherwise most of the other methods do implement some specific functionality enough to warrant their own description. We could probably re-structure stuff to use inheritdoc followed by a custom message if we cared enough to do so. It might help improve documentation accuracy and changes only need to be made in one place instead of several.

bantu:
@cnelissen Thanks for your analysis.

As for the patch, I think the new variant with the empty method in RC4 is almost as weird as the previous version. Maybe an even better abstraction can be found in the future.

Issue #575, opened 2014-12-17
Add visibility statements for all class properties and methods


cnelissen:
As part of modernizing the PHP 5 branch, we should add visibility (public/protected/private) statements for every property and method on every class. This allows for a more understandable and definitive API for everyone using the library, and will aid in development since developers can be confident about which methods are being used exposed publicly versus internal methods which will not break the API.

Since this is a library, we should probably avoid using "private" in almost all circumstances and should instead use "protected" since any class should be available to be extended.

If this is agreeable I would more than happy to submit a PR.

PR #591, opened 2014-12-30
Convert Base::_createInlineCryptFunction to use PHP 5 anonymous functions instead of create_function

cnelissen:
This PR converts Base::_createInlineCryptFunction to use native PHP 5 anonymous functions instead of the PHP 4 create_function method. This is necessary for adding proper support for visibility statements throughout the codebase as the old lambda functions were not scope-aware and were causing visibility conflicts.

Additionally, the anonymous functions are showing a significant performance gain. Running the unit tests on the current version yields memory usage of 21.75Mb and running against this PR yields memory usage of 17.75Mb.

I am planning on further converting these functions to get rid of using eval statements, but I figured this should be good for a single PR.

Pre-requisite for #575

### DIFF
--- a/phpseclib/Crypt/Base.php
+++ b/phpseclib/Crypt/Base.php
@@ -395,7 +395,7 @@ class Base
     var $inline_crypt;
 
     /**
-     * Holds whether performance-optimized $inline_crypt() can/should be used.
+     * Holds whether performance-optimized $inline_crypt() can/should be used. Defaults to true.
      *
      * @see \phpseclib\Crypt\Base::encrypt()
      * @see \phpseclib\Crypt\Base::decrypt()
@@ -403,7 +403,7 @@ class Base
      * @var mixed
      * @access private
      */
-    var $use_inline_crypt;
+    var $use_inline_crypt = true;
 
     /**
      * Default Constructor.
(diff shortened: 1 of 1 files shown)


bantu:
@terrafrost Do you want to look into performance implications of this PR?

terrafrost:
I'll do that when I get home as time permits.

terrafrost:
Before this commit internal mode takes 1.4 seconds. After this commit it takes 25 seconds. Need to figure out what's going on. My test code:

```php
<?php
include('autoloader.php'); // https://raw.githubusercontent.com/composer/composer/master/src/Composer/Autoload/ClassLoader.php

$loader->addPsr4('phpseclib\\', __DIR__.'/phpseclib');
$loader->register();

use \phpseclib\Crypt\AES;

$plaintext = '';
for ($i = 0; $i < 1024*1024; $i++) {
    $plaintext.= chr(mt_rand(0, 255));
}

$start = microtime(true);

define('CRYPT_AES_MODE', AES::ENGINE_INTERNAL);
$aes = new AES();
$aes->setKey('abcdefghijklmnop');
$aes->encrypt($plaintext);
```

cnelissen:
I am seeing similar results... I will mess around with the eval statements and see what I can do.

bantu:
This will be closed as the php5 will be going away. Feel free to send a new PR.

Issue #597, opened 2015-01-06
OpenSSL library and header mismatch causing poor performance


haphan:
https://github.com/phpseclib/phpseclib/blob/master/phpseclib/Math/BigInteger.php#L286-L294

phpseclib takes advantage of openssl feature only if OpenSSL library and OpenSSL header version are identical. 

Major distributions like Apple PHP and MAMP PHP were compiled with minor difference in version. 

Apple PHP
```bash
haphan@koala:~/$ /usr/bin/php -i | grep OpenSSL
OpenSSL support => enabled
OpenSSL Library Version => OpenSSL 0.9.8za 5 Jun 2014
OpenSSL Header Version => OpenSSL 0.9.8y 5 Feb 2013
OpenSSL support => enabled
```

MAMP
```bash
haphan@koala:~/$ php -i | grep OpenSSL
SSL Version => OpenSSL/0.9.8z
OpenSSL support => enabled
OpenSSL Library Version => OpenSSL 0.9.8za 5 Jun 2014
OpenSSL Header Version => OpenSSL 0.9.8r 8 Feb 2011
Native OpenSSL support => enabled
```

I think phpseclib should only check for major version of openSSL.

terrafrost:
Overall I think it's a good commit but... 

```php
                            //Look for major version
                            $fullVersion = trim(str_replace('=>', '', strip_tags($matches[2][$i])));
                            preg_match('/(\d+\.\d+\.\d+)/i', $fullVersion, $m);

                            if (!empty($m)) {
                                $versions[$matches[1][$i]] = $m[0];
                            } else {
                                $versions[$matches[1][$i]] = $fullVersion;
                            }
```

You could do this instead:

```php
                            //Look for major version
                            $fullVersion = trim(str_replace('=>', '', strip_tags($matches[2][$i])));
                            if (!preg_match('/(\d+\.\d+\.\d+)/i', $fullVersion, $m)) {
                                $versions[$matches[1][$i]] = $fullVersion;
                            } else {
                                $versions[$matches[1][$i]] = $m[0];
                            }
```

Also, saying major version is a little misleading. With semantic versioning you have [major].[minor].[patch], so people used to versioning like that might think you're only checking to see if it's OpenSSL 1.0 headers with an OpenSSL 2.0 library. Maybe a better comment would be... "remove [letter versions](http://stackoverflow.com/a/10347301/569976)] or some such.

terrafrost:
Fixed by #598

PR #598, opened 2015-01-06
(OpenSSL Big integer) Only require major version to match

haphan:
See issue [#597](https://github.com/phpseclib/phpseclib/issues/597)

### DIFF
--- a/phpseclib/Crypt/RSA.php
+++ b/phpseclib/Crypt/RSA.php
@@ -515,9 +515,16 @@ class Crypt_RSA
 
                     $versions = array();
                     if (!empty($matches[1])) {
-                       for ($i = 0; $i < count($matches[1]); $i++) {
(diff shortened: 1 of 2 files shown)


cnelissen:
This project uses PSR-2 as a coding standard, which is why your Travis checks are failing. Take a look at http://www.php-fig.org/psr/psr-2, specifically section 5. IF statements should like like:

```
if ($expr1) {
    // if body
} elseif ($expr2) {
    // elseif body
} else {
    // else body;
}
```

haphan:
Thank you @cnelissen. I will fix and create pr again

bantu:
Can you describe pre-patch and post-patch behaviour. It is not very clear to me, even after reading #597.

bantu:
Can you please add spaces after "//" and squash all your commits?

haphan:
@bantu done

bantu:
@haphan Thanks.

PR #599, opened 2015-01-06
Require PHP 5.3.3 since that is the lowest version we are running tests ...

### LINKED ISSUE #571: Change PHP requirement from 5.3.2 to 5.3.3
As has been said elsewhere, the `stream_resolve_include_path` function is no longer used. Thus we can probably lower the PHP version to 5.3.0, unless anyone else sees a dependency only available in higher versions.

bantu:
Fix #571

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,7 @@
         }
     ],
     "require": {
-        "php": ">=5.3.2"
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
:+1:

PR #612, opened 2015-01-28
.scrutinizer.yml: Use large timeout to prevent coverage receiver from ti...

bantu:
...ming out.

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -4,3 +4,4 @@ imports:
 tools:
     external_code_coverage:
         runs: 5 # No Code Coverage on PHP 5.2 and HHVM
+        timeout: 2700 # 45 minutes


.scrutinizer.yml
```diff
@@ -4,3 +4,4 @@ imports:
 tools:
     external_code_coverage:
         runs: 5 # No Code Coverage on PHP 5.2 and HHVM
+        timeout: 2700 # 45 minutes
```
GrahamCampbell: I normally go for 30 mins. 45 mins can be a long time to wait if someone sends a bad pull request and pushes a few commit to try to fix it, they'll lock up scrutinizer for hours.

bantu: I had it at 25 mins earlier, but then realised that tests run for 20 minutes (wall time) and they might be queued for a bit too.

GrahamCampbell: Oh right. Makes sense then. My test suites are normally much smaller so take less time. 45 mins is probably good then. :)

Issue #496, opened 2014-10-29
Provide some kind of changelog, or at least API stability note


AdamWill:
So as the Fedora packager for phpseclib I swing by and note that since 0.3.5 (the version I currently have packaged) there have been three releases, several hundreds (or thousands?) of commits...and absolutely no release notes or changelog. So far as I can tell there's no formal API stability information anywhere in the source tree.

So...I can just assume blindly that 0.3.8 is backwards compatible with 0.3.5 and start packaging it as an update for stable Fedora releases, but that seems a bit of a risk. Or I can try and pick my way through every single commit since 0.3.5 and try and figure out whether it changes the API, but it seems like it'd take a while.

It might be nice if there was a reasonable high-level changelog or at least release notes between releases, or at minimum at least a note that said 'this doesn't change API from the previous release' (or, you know, it _does_). Thanks!

PR #621, opened 2015-02-08
CHANGELOG: add 0.3.10 release

terrafrost:
See #496 . I figure this should be merged into the 2.0 branch and the master branch as well. Should 2.0.0 be mentioned in the changelog? Since 2.0.0 isn't even tagged it's not entirely clear to me if that version has even been officially released. Regardless, if it has, that can be another pull request.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 0.3.10 - 2015-02-04
 
-- simplify window size handling ([#538](https://github.com/phpseclib/phpseclib/pull/538))
+- simplify SSH2 window size handling ([#538](https://github.com/phpseclib/phpseclib/pull/538))
 - slightly relax the conditions under which OpenSSL is used ([#598](https://github.com/phpseclib/phpseclib/pull/598))
 - fix issue with empty constructed context-specific tags in ASN1 ([#606](https://github.com/phpseclib/phpseclib/pull/606))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.3.10 - 2015-02-04
+
+- simplify window size handling ([#538](https://github.com/phpseclib/phpseclib/pull/538))
+- slightly relax the conditions under which OpenSSL is used ([#598](https://github.com/phpseclib/phpseclib/pull/598))
+- fix issue with empty constructed context-specific tags in ASN1 ([#606](https://github.com/phpseclib/phpseclib/pull/606))
+
 ## 0.3.9 - 2014-11-09
 
 - PHP 5.6 improvements ([#482](https://github.com/phpseclib/phpseclib/pull/482), [#491](https://github.com/phpseclib/phpseclib/issues/491))


bantu:
You should be requesting this to be merged into 1.0 then, right?

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.3.10 - 2015-02-04
+
+- simplify window size handling ([#538](https://github.com/phpseclib/phpseclib/pull/538))
```
bantu: should probably supply some context here. ssh?

bantu:
Not sure about the best way to handle changelog. Symfony uses version specific changelogs. E.g. CHANGELOG-1.0.md and CHANGELOG-2.0.md.

GrahamCampbell:
Why don't we have a 0.3 branch? Surely we should have one, even if it's not maintained?

bantu:
@GrahamCampbell 0.3.x becomes 1.0.0 when OpenSSL support is implemented. So 0.3 is in the 1.0 branch and fully BC.

GrahamCampbell:
Right, so you're just pushing out a final 0.3 release from the 1.0 branch then. Makes sense. I've done something similar before. :)

bantu:
@GrahamCampbell My understand was that terrafrost wants to maintain 1.0 for a longer time. Possibly lower than 2.0. Kind of as an LTS release.

terrafrost:
> You should be requesting this to be merged into 1.0 then, right?

Yup. idk how this should be handled for the 2.0 series.

> @GrahamCampbell My understanding is that terrafrost wants to maintain 1.0 for a longer time. Possibly longer than 2.0. Kind of as an LTS release.

Yah - pretty much.

CHANGELOG.md
terrafrost: Yah - it's for SSH. SSH2 specifically. I'll update it later today.

bantu:
> Yup. idk how this should be handled for the 2.0 series.

Rename CHANGELOG.md to CHANGELOG-1.0.md in 1.0 branch. Merge down. Add CHANGELOG-2.0 to 2.0 branch. Merge down. ?

GrahamCampbell:
> Rename CHANGELOG.md to CHANGELOG-1.0.md in 1.0 branch. Merge down. Add CHANGELOG-2.0 to 2.0 branch. Merge down. ?

:+1: :rocket:

ravage84:
http://keepachangelog.com/ has some ideas about handling change logs.

I personally would keep only one change log.
Starting a new one for each new major release (SemVer-wise).
All pre-stable releases are also kept in this change log.
So all 0.x.x releases including all beta and RC releases of 1.x would be kept in the change log which is included in the code tree of 1.x

terrafrost:
@ravage84 - you're proposing that each branch have it's own changelog independent of any other branch? eg.

CHANGELOG.md for the 2.0 branch starts at 2.0, as tho the 0.x/1.x series never happened?

@AdamWill - what are your thoughts?

AdamWill:
Not sure why you're asking me, for a distro perspective? Don't think I have a strong opinion or Fedora / distros in general have particular policies, it's not at all uncommon to have apps which don't have all their history available in current tarballs for sure. So long as it's all available *somewhere* I don't think I really have an opinion.

ravage84:
@terrafrost
The CHANGELOG.md for the 2.0 branch starts with whatever version was developed for this branch first (normally a pre stable release, such as 2.0.0-beta1). So people can see the changes that happened towards the first major stable of that branch.
I would not mind having the full back log of changes that happened before that branch, but it creates an interesting problem when you have more than one branch actively maintained.
Also I don't think there is much of a benefit of having the full change log in a branch of  a later release, because if you are using the 2.x release and really need to know what changed within the lifecycle of the 1.x release you can simple look it up in that branch.

cnelissen:
I concur with @ravage84... Each branch really only needs it's own history. If you are interested in history from a different branch, then you could go look at it.

bantu:
A single changelog per branch likely results in merge conflicts when merging e.g. 1.0 into 2.0. This is not a big problem, but should be considered.

bantu:
Replaced by #624

PR #592, opened 2014-12-30
SSH agent forwarding implementation

montdidier:
A clean single commit version of SSH agent forwarding implementation work.

PR #623, opened 2015-02-14
add setPreferredEngine()

terrafrost:
Backport of setPreferredEngine() from #507 [per bantu](https://github.com/phpseclib/phpseclib/pull/507#issuecomment-65205702):

> I know this is based on #259, but still: This patch is huge and does too many things at once and as a result is hard to review (if at all). I would have appreciated if parts from #259 were extracted into their own PRs. E.g. why not send a pull request only implementing _setEngine? OpenSSL can hardly be required for that. 

For a version of this branch that has been merged with the 2.0 branch see #643


bantu:
For some reason I missed that this PR was sent. Sorry.

phpseclib/Crypt/AES.php
```diff
@@ -107,20 +111,6 @@
...
 /**#@-*/
 
-/**#@+
- * @access private
- * @see Crypt_Base::Crypt_Base()
- */
-/**
- * Toggles the internal implementation
- */
-define('CRYPT_AES_MODE_INTERNAL', CRYPT_MODE_INTERNAL);
```
bantu: Breaks BC?

bantu: Hummm, okay, declared private.

phpseclib/Crypt/Base.php
```diff
@@ -53,9 +53,10 @@
  */
 
 /**#@+
- * @access public
+ * @access private
```
bantu: This doesn't sounds right. CRYPT_MODE_CTR etc. are part of API.

phpseclib/Crypt/Base.php
```diff
@@ -1179,16 +1150,113 @@ function disableContinuousBuffer()
...
+    function isValidEngine($engine)
+    {
+        switch ($engine) {
+            case CRYPT_ENGINE_MCRYPT:
+                return $this->cipher_name_mcrypt &&
+                       extension_loaded('mcrypt') &&
+                       in_array($this->cipher_name_mcrypt, mcrypt_list_algorithms());
+            case CRYPT_ENGINE_INTERNAL:
+                return true;
+        }
```
bantu: return false?

phpseclib/Crypt/Base.php
```diff
@@ -1179,16 +1150,113 @@ function disableContinuousBuffer()
...
+     * @see Crypt_Base::Crypt_Base()
+     * @access private
+     */
+    function _setEngine()
+    {
+        switch (true) {
+            case $this->isValidEngine($this->preferredEngine):
+                $this->engine = $this->preferredEngine;
+                break;
+            case $this->isValidEngine(CRYPT_ENGINE_MCRYPT):
```
bantu: Repeating check and set here. A loop over an array holding the order might be better.

phpseclib/Crypt/Base.php
```diff
@@ -1241,14 +1307,12 @@ function _setupKey()
...
      *
-     * Internally: _setup() is called always before(!) en/decryption.
-     *
-     * Note: Could, but not must, extend by the child Crypt_* class
-     *
      * @see setKey()
      * @see setIV()
      * @see disableContinuousBuffer()
      * @access private
+     * @internal _setup() is called always before(!) en/decryption.
```
bantu: "is always called before"

phpseclib/Crypt/Base.php
```diff
@@ -1383,14 +1445,12 @@ function _unpad($text)
...
-     * Note: Could, but not must, extend by the child Crypt_* class
-     *
      * @access public
+     * @internal Could, but not must, extend by the child Crypt_* class
      */
     function _clearBuffers()
     {
-        $this->enbuffer = array('encrypted'  => '', 'xor' => '', 'pos' => 0, 'enmcrypt_init' => true);
-        $this->debuffer = array('ciphertext' => '', 'xor' => '', 'pos' => 0, 'demcrypt_init' => true);
+        $this->enbuffer = $this->debuffer = array('ciphertext' => '', 'xor' => '', 'pos' => 0, 'enmcrypt_init' => true);
```
bantu: Not sure how this change is related to "add setPreferredEngine". Seems to be unrelated cleanup.

phpseclib/Crypt/Base.php
```diff
@@ -873,16 +850,20 @@ function encrypt($plaintext)
...
      * @see Crypt_Base::encrypt()
      * @access public
      * @param String $ciphertext
      * @return String $plaintext
+     * @internal Could, but not must, extend by the child Crypt_* class
      */
     function decrypt($ciphertext)
     {
-        if ($this->engine == CRYPT_MODE_MCRYPT) {
+        if ($this->paddable) {
```
bantu: Not sure how this change is related to "add setPreferredEngine". Seems to be unrelated cleanup.

phpseclib/Crypt/Base.php
```diff
@@ -645,16 +626,19 @@ function setPassword($password, $method = 'pbkdf2')
...
      * @access public
      * @param String $plaintext
-     * @return String $cipertext
+     * @return String $ciphertext
+     * @internal Could, but not must, extend by the child Crypt_* class
      */
     function encrypt($plaintext)
     {
-        if ($this->engine == CRYPT_MODE_MCRYPT) {
+        if ($this->paddable) {
```
bantu: Not sure how this change is related to "add setPreferredEngine". Seems to be unrelated cleanup.

phpseclib/Crypt/Base.php
```diff
@@ -2001,11 +2057,48 @@ function _createInlineCryptFunction($cipher_code)
...
+     * @param $bytes
+     * @return String
+     */
+    function _trapdoor($bytes)
+    {
+        if (!defined('CRYPT_BASE_WHIRLPOOL_AVAILABLE')) {
+            define('CRYPT_BASE_WHIRLPOOL_AVAILABLE', (bool)(extension_loaded('hash') && in_array('whirlpool', hash_algos())));
+        }
+
+        // return pack('H*', md5($bytes) . sha1($bytes) . (CRYPT_BASE_WHIRLPOOL_AVAILABLE ? hash('whirlpool', $bytes) : '')); // Alternative
```
bantu: Remove

phpseclib/Crypt/Base.php
```diff
@@ -2001,11 +2057,48 @@ function _createInlineCryptFunction($cipher_code)
...
+     * Generates a digest from $bytes
+     *
+     * @see _setupInlineCrypt()
+     * @access private
+     * @param $bytes
+     * @return String
+     */
+    function _trapdoor($bytes)
+    {
+        if (!defined('CRYPT_BASE_WHIRLPOOL_AVAILABLE')) {
```
bantu: Introduction of a non-constant constant. Use static variable or property.

phpseclib/Crypt/Blowfish.php
```diff
@@ -534,15 +519,14 @@ function _setupInlineCrypt()
...
-
-        switch (true) {
-            case $gen_hi_opt_code:
-                $code_hash = md5(str_pad("Crypt_Blowfish, {$this->mode}, ", 32, "\0") . $this->key);
-                break;
-            default:
-                $code_hash = "Crypt_Blowfish, {$this->mode}";
+        $gen_hi_opt_code = (bool)( count($lambda_functions) < 10 );
+
+        // Generation of a uniqe hash for our generated code
```
bantu: unique

phpseclib/Crypt/Base.php
```diff
@@ -2001,11 +2057,48 @@ function _createInlineCryptFunction($cipher_code)
...
+
+    /**
+     * Generates a digest from $bytes
+     *
+     * @see _setupInlineCrypt()
+     * @access private
+     * @param $bytes
+     * @return String
+     */
+    function _trapdoor($bytes)
```
bantu: This might be (likely is) inconsistent with the definition of a cryptographical trapdoor function. Use a better name. _digest_whatever_you_are_digesting()?

phpseclib/Crypt/Base.php
terrafrost: Oh right - good point.

phpseclib/Crypt/Base.php
terrafrost: Good idea - thanks!

phpseclib/Crypt/Base.php
terrafrost: That's correct - it is an unrelated cleanup.

phpseclib/Crypt/Blowfish.php
terrafrost: I'm not sure what you mean by that? FWIW some of this stuff was in petrich's orig PR. Although this PR is based off of that I can't necessarily give immediate insight into why everything was done the way it is - in some cases I'd have to do research or in some cases it might have been just arbitrary too. So for some of the stuff you mentioned I'll have to look into it before I comment on it.

bantu: I mean: "uniqe" should be "unique". Simple typo.

terrafrost: Oh lol. I see the typo now. Good catch!

phpseclib/Crypt/Base.php
terrafrost: So something like this?:

```php
$candidateEngines = array($this->preferredEngine, CRYPT_ENGINE_MCRYPT, CRYPT_ENGINE_INTERNAL);
foreach ($candidateEngines as $engine) {
    if ($this->isValidEngine($engine)) {
        $this->engine = $this->preferredEngine;
        break;
    }
}
```

It's doing one more isValidEngine() check than is needed, although I suppose there's nothing wrong with that. It's not like it's in a part of the code that needs to be high performance.

Another possibility, that's a little less readable:

```php
$engines = array($this->preferredEngine, CRYPT_ENGINE_MCRYPT);
for ($i = 0; $i < count($engines) && !$this->isValidEngine($engines[$i]); $i++);
$this->engine = $i == count($engines) ? CRYPT_ENGINE_INTERNAL : $engines[$i];
```

bantu: I'd say the for loops is rather ugly too. I'd recommend using the foreach loop and handling CRYPT_ENGINE_INTERNAL as a special case (ommiting the isValidEngine() check as it is always available).

Something like this:
```php
$candidateEngines = array(
    $this->preferredEngine,
    CRYPT_ENGINE_MCRYPT,
);
foreach ($candidateEngines as $engine) {
    if ($this->isValidEngine($engine)) {
        $this->engine = $this->preferredEngine;
        break;
    }
}
if (!$this->engine) {
    $this-engine = CRYPT_ENGINE_INTERNAL;
}
```

phpseclib/Crypt/Base.php
terrafrost: phpseclib / php4 / 1.0 already uses non-constant variables. eg. MATH_BIGINTEGER_OPENSSL_ENABLED, MATH_BIGINTEGER_MODE, CRYPT_RSA_MODE, etc.

For php5 / 2.0, a static class variable very definitely ought to be used but static variables aren't available in php4 / 1.0 and I don't think a property is a good way to go on this. I mean, if we go with a property on this for php4 / 1.0 we should probably go with properties for all the above instances where constants are already being used in this manner and even if we omitted `MATH_BIGINTEGER_*` it's still a check that'd now have to be done over and over and over again when instead it could have been done once and just once and that's it.

bantu: static properties do not exist in PHP4, but i think static variables are available.

terrafrost: I don't think they are. I mean, PHP4 supports static variables in methods but not in classes. And static variables in methods can't be seen outside of those methods.

Example: http://3v4l.org/6R7cr#v430

PHP5+ gives no output. PHP4.x gives a parse error.

bantu: I actually meant a static variable in the method.

phpseclib/Crypt/Base.php
terrafrost: I think this is a bit mincing words but, in-so-far as new names go...  how about `_generateGUID` where GUID = globally unique identifier?

That's what this function is essentially doing. It creates loop-unrolled dynamic functions that are specific to the key / block size / number rounds, etc, being used.

Rather than use the key as an index to the array, however, a hash of the key was used.

Here's what it's doing now:

```php
$code_hash = md5(str_pad("Rijndael, {$this->mode}, {$this->block_size}, ", 32, "\0") . implode(',', $w));
```

What it does after this PR:

```php
        $code_hash = "Crypt_Rijndael, {$this->mode}, {$this->Nr}, {$this->Nb}";
        if ($gen_hi_opt_code) {
            $code_hash = str_pad($code_hash, 32) . $this->_trapdoor($this->key);
        }
```

I'm just asking now because I don't want to go make a commit to change the name to something and for you to then come back and say you think that's not a good name either lol. I'd rather just figure out what a good name is and *then* make the change.

bantu: generateGUID sounds like taking something from random. This sounds like deterministic digesting/hashing, though.

> It creates loop-unrolled dynamic functions that are specific to the key / block size / number rounds, etc, being used.

_digest_unrolled_function?

terrafrost: What about `_digest_dynamic_function`? idk that I like `_digest_unrolled_function` because that implies that that's the only optimization. It's the only one I mentioned and recalled when I made the last post but it's not the only one. Quoting the header for the function that makes the dynamic functions:

```php
     *    The main reason why can speed up things [up to 50%] this way are:
     *
     *    - using variables more effective then regular.
     *      (ie no use of expensive arrays but integers $k_0, $k_1 ...
     *      or even, for example, the pure $key[] values hardcoded)
     *
     *    - avoiding 1000's of function calls of ie _encryptBlock()
     *      but inlining the crypt operations.
     *      in the mode of operation for() loop.
     *
     *    - full loop unroll the (sometimes key-dependent) rounds
     *      avoiding this way ++$i counters and runtime-if's etc...
```

The name of that method is `_createInlineCryptFunction`.

bantu: Then this should be `_hashInlineCryptFunction` or `_digestInlineCryptFunction`?

phpseclib/Crypt/Base.php
terrafrost: I guess that actually could work since the `CRYPT_BASE_WHIRLPOOL_AVAILABLE` isn't really used outside of that function, however, I'm still lacking enthusiasm about doing a static variable in the method as it's just inconsistent with the way phpseclib does other similar tests (as noted above).

And when phpseclib does do static variables in methods it's not doing them for 

I mean I'll do it if it helps get this merged but I think it is a little inconsistent for the php4 / 1.0 branch.

bantu: I guess I don't care too much. It should be fine.

phpseclib/Crypt/Base.php
terrafrost: I can do that. Altho we'd need to do `$this->engine = false` before all of that lest something like this happen.

For the initial key / IV length mcrypt is a valid engine, so `$this->engine` gets set to CRYPT_ENGINE_MCRYPT. Then you change the key length to one for which only the internal engine is valid. `$this->engine` is never changed to internal because `$this->engine` is never false (or NULL or 0).

bantu: Ok. Just add `$this->engine = false` at the top here? That should be sufficient.

bantu: Or maybe rather null.

terrafrost:
I'll try to make a commit with the changes tomorrow night. Have most of them done but I need to do some other stuff before I go to bed tonight.

Thanks for the feedback!

terrafrost:
The recommended changes have been made.

bantu:
Will try to look at this again tomorrow.

terrafrost:
Is there any way I can help speed this along?

terrafrost:
One more week of inactivity on this and I'm thinking I'm just going to merge #507 myself (which will kinda invalidate this particular PR). I like the idea of code reviews for pull requests as they demonstrate a willingness on the part of this project to employ best practices. But practicality has to be taken into consideration too.

I realize that I, myself, can take a while to code review big PR's (eg. #592), but that said, I also essentially take ownership of any big PR I approve. It'd be nice if people supported their larger code contributions after they were merged but they don't and thus the onus almost always falls on me.

The last commit for this was 19 days ago.

terrafrost:
I created a new PR that has this feature merged into the 2.0 branch, with all the merge conflicts hopefully resolved (I guess the unit tests will be the ultimate determiner). Hopefully this will remove an obstacle for the merging of this PR.

terrafrost:
A hurdle for merging #640 (aka #507) myself has been overcome with #652 targeting the 2.0 branch. The recommendations made in this PR have been incorporated into #640 and #652 so those are as up-to-date as they can be.

I'd like it if this PR and #643 were given a final sign off but in lieu of one I think I'll go ahead and merge #640 and #652 if there's no activity on this or #643 within a week.

PR #624, opened 2015-02-19
CHANGELOG: add 0.3.10 release

bantu:
Replaces #621

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 0.3.10 - 2015-02-04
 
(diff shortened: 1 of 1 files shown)


bantu:
For next time: This should obviously be included in the tag (0.3.10 in this case).

terrafrost:
Looks good to me. And you're right - it should have been in the 0.3.10 tag meh. Last time I retagged something, however, that caused a bit of an outcry, so I guess that's not an option.

bantu:
Err yes. Retagging is an absolute no-go.

Issue #625, opened 2015-02-24
Add arbitrary extension to X.509 certificate


asherkin:
I've poked around a bit in the ASN1 and X509 classes, but it looks like everything has to be hard-coded in the X509 class? _getMapping appears to have a provision for unknown extensions, but it seems to only be considered for parsing (which is also alluded to by a comment in said function).

I'm after something like https://www.openssl.org/docs/apps/x509v3_config.html#arbitrary_extensions.

terrafrost:
In theory, with phpseclib, you'd use File_ASN1_Element to create arbitrary extensions. In practice, it looks like you'll have to add one line to File/X509.php:

```
#
#-----[ OPEN ]------------------------------------------
#
File/X509.php
#
#-----[ FIND ]------------------------------------------
#
    function _mapOutExtensions(&$root, $path, $asn1)
    {
        $extensions = &$this->_subArray($root, $path);

        if (is_array($extensions)) {
            $size = count($extensions);
            for ($i = 0; $i < $size; $i++) {
#
#-----[ AFTER, ADD ]------------------------------------
#
if (is_object($extensions[$i]) && strtolower(get_class($extensions[$i])) == 'file_asn1_element') continue;
```

Once that's done you can do this:

```php
<?php

include('File/X509.php');

$x509 = new File_X509();
$cert = $x509->loadX509('-----BEGIN CERTIFICATE-----
MIIDITCCAoqgAwIBAgIQT52W2WawmStUwpV8tBV9TTANBgkqhkiG9w0BAQUFADBM
MQswCQYDVQQGEwJaQTElMCMGA1UEChMcVGhhd3RlIENvbnN1bHRpbmcgKFB0eSkg
THRkLjEWMBQGA1UEAxMNVGhhd3RlIFNHQyBDQTAeFw0xMTEwMjYwMDAwMDBaFw0x
MzA5MzAyMzU5NTlaMGgxCzAJBgNVBAYTAlVTMRMwEQYDVQQIEwpDYWxpZm9ybmlh
MRYwFAYDVQQHFA1Nb3VudGFpbiBWaWV3MRMwEQYDVQQKFApHb29nbGUgSW5jMRcw
FQYDVQQDFA53d3cuZ29vZ2xlLmNvbTCBnzANBgkqhkiG9w0BAQEFAAOBjQAwgYkC
[...]

PR #627, opened 2015-03-01
X509: make it so you can use File_ASN1_Element for custom X.509 extensions

terrafrost:
Addresses #625


bantu:
Tests?

terrafrost:
I added a unit test to this if you want to code review, bantu!

Issue #626, opened 2015-02-25
ASN.1 Parser issue detecting constructed/context specific content


metaclassing:
I have encountered an issue with the ASN.1 parser, I have example cert & code posted over at stack overflow: http://stackoverflow.com/questions/28686874/phpseclib-asn1-parser-issue-with-ms-cert-san/28710090#28710090

I would really appreciate some input as to the best resolution for this behavior, I dont quite understand why the bitwise operations to determine class/construction were selected or feel confident modifying that logic without guidance.

Thanks!

terrafrost:
https://github.com/phpseclib/phpseclib/pull/628 kinda fixes this altho not in the way suggested by the stackoverflow post.

The problem with doing what the stackoverflow post is that doing that would mix ASN1 mapped stuff (`asn1map`) and ASN1 unmapped stuff (`decodeBER`). I mean, yah, the extnValue is, in theory, supposed to be DER encoded, but for one, it can be malformed without affecting the overall parsability of the X.509 cert, and for two, without any sort of mapping, it wouldn't have the same semantic meaning that the rest of the mapped X.509 cert had.

metaclassing:
I grabbed the latest code off github and modified the line you mentioned in #628 , I guess I misunderstood something. I'm getting a binary blob at the end of the cert object:
[signatureSubject] => 0^Â^D¾ ^C^B^A^B^B^SP^@^@^@^M*^ÃÆÄè8k,^@^@^@^@^@^M0^M^F    *^ÆH^Æ÷^M^A^A^K^E^@0"1 0^^^F^CU^D^C^S^W.Secure Enterprise CA...

I can recursively iterate through the currentcert/tbscert looking for base64 strings and decoding them into nested arrays, that gets me closer to reading the extensions I guess but a little painful to decode the data that way. Maybe I am missing the big picture?

terrafrost:
You must have done something wrong. I just added unit tests to test this (didn't change any code in File/X509.php) and it works just fine.

That said I did have have to submit a new PR since #628 was against master instead of 1.0 (and idk how to change it). The new PR is #629.

Regardless, rather than modifying the line by hand how about just copy / paste the file, pre-modified?:

https://raw.githubusercontent.com/terrafrost/phpseclib/e686c095b63eadfa3ed60b1cc4c01c7e17d30b77/phpseclib/File/X509.php

[...]

PR #628, opened 2015-03-01
X509: always base64-encode extensions for which _getMapping returns a bool

terrafrost:
Addresses #626


bantu:
This should be against 1.0 instead

bantu:
Tests?

terrafrost:
> This should be against 1.0 instead

Yah - it should. Is there a way I can change that or do I need to do a whole new PR?

> Tests?

I can add one. Did you have a chance to look at any of the OpenSSL stuff btw?

PR #633, opened 2015-03-13
Banner timeout feature

sapzxc:
I bumped of situations when server sending some non ssh banner (not "SSH-...") but some another, like "MyProtoV1.1". So then we still waiting for next string and then again and so on. But any way we could not reach header need for us. For this situations we need some timeout. This patch made this timeout feature.

Function stream_set_timeout provide breaking of waiting for gets. And then we get information about socket (stream_get_meta_data) and test there "timed_out" flag.


sapzxc:
Sorry, closed, will recheck some cases.

PR #634, opened 2015-03-13
Banner timeout feature (fixed)

sapzxc:
I bumped of situations when server sending some non ssh banner (not "SSH-...") but some another, like "MyProtoV1.1". So then we still waiting for next string and then again and so on. But any way we could not reach header need for us. For this situations we need some timeout. This patch made this timeout feature.

Function stream_set_timeout provide breaking of waiting for gets. And then we get information about socket (stream_get_meta_data) and test there "timed_out" flag.

### DIFF
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -978,7 +978,6 @@ class SSH2
            MUST be able to process such lines." */
         $temp = '';
         $extra = '';
-        stream_set_timeout($this->fsock, $sec, $usec);
         while (!feof($this->fsock) && !preg_match('#^SSH-(\d\.\d+)#', $temp, $matches)) {
             if (substr($temp, -2) == "\r\n") {
                 $extra.= $temp;
@@ -988,7 +987,7 @@ class SSH2
 
             $info = stream_get_meta_data($this->fsock);
             if($info['timed_out']) {
-                user_error('Banner timeout or wrong.');
+                user_error('Banner read timeout or wrong.');
                 return false;
             }
         }
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -978,12 +978,19 @@ class SSH2
            MUST be able to process such lines." */
         $temp = '';
         $extra = '';
+        stream_set_timeout($this->fsock, $sec, $usec);
         while (!feof($this->fsock) && !preg_match('#^SSH-(\d\.\d+)#', $temp, $matches)) {
             if (substr($temp, -2) == "\r\n") {
                 $extra.= $temp;
                 $temp = '';
             }
             $temp.= fgets($this->fsock, 255);
+            
+            $info = stream_get_meta_data($this->fsock);
+	        if($info['timed_out']) {
+		        user_error('Banner timeout or wrong.');
+		        return false;
+	        }
         }
 
         if (feof($this->fsock)) {
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -985,12 +985,12 @@ class SSH2
                 $temp = '';
             }
             $temp.= fgets($this->fsock, 255);
-            
+
             $info = stream_get_meta_data($this->fsock);
-	        if($info['timed_out']) {
-		        user_error('Banner timeout or wrong.');
-		        return false;
-	        }
+            if($info['timed_out']) {
+                user_error('Banner timeout or wrong.');
+                return false;
+            }
         }
 
         if (feof($this->fsock)) {


terrafrost:
I think it's a good idea however, I think it'd be better to use `stream_select` (which, despite your third commit's commit message, you do not appear to be using). eg.

```php
                // on windows this returns a "Warning: Invalid CRT parameters detected" error
                if (!@stream_select($read, $write, $except, $sec, $usec) && !count($read)) {
                    $this->is_timeout = true;
                    return true;
                }
```

sapzxc:
Yes, you are right. I fixed this and copy-paste message from previous pull, my fault. 
But code in this pull works properly, tested several days.

terrafrost:
Does https://github.com/terrafrost/phpseclib/tree/banner-timeout work for you?

sapzxc:
Looks like it is similar fix but made by hard way using own timeout flags instead system function :)
Did not tested your tree yet.

Why you do not prefer stream_get_meta_data ?

terrafrost:
> Why you do not prefer stream_get_meta_data ?

Consistency. Also, you're not using `stream_select` so it won't time out after a pre-determined amount of time - it'll just timeout whenever.

If you do do `stream_select` then the next problem is that the timeouts aren't cumulative. So if your timeout is ten seconds and each fgets() call takes two seconds the approach that I'm using will timeout after five fgets() calls whereas using `stream_get_meta_data` won't ever timeout since technically none of the individual fgets() calls ever timed out.

sapzxc:
Not sure about ``stream_get_meta_data``. Lets read php documentation:

> timed_out (bool) - TRUE if the stream timed out while waiting for data **on the last call to fread() or fgets()**.

Need test, but I suggest that stream_get_meta_data works as we need.

terrafrost:
idk if you're definition of "as we need" accounts for cumulative timeouts. But let's test it out.

```php
<?php
set_time_limit(0);

$socket = socket_create(AF_INET, SOCK_STREAM, 0);
socket_bind($socket, '127.0.0.1', 9999) or die("Could not connect");
socket_listen($socket);

$client_connection = socket_accept($socket);

for ($i = 0; $i < 10; $i++) {
	sleep(2);
	socket_write($client_connection, "zzz");
}

socket_close($client_connection);
socket_close($socket);
```

So with a timeout of 10 how many zzz's should connecting to 127.0.0.1:9999 yield? I say 5. So let's test it out.

Several different possible clients have been proposed. Your commits include several different varieties. The code as of https://github.com/sapzxc/phpseclib/commit/63f49ac52a91e5da2ef9ee88babf6b47ed3009b4 is analogous to this:

```php
<?php
$fsock = fsockopen('127.0.0.1', 9999);
$i = 0;
while (!feof($fsock)) {
	$i++;
	echo "$i. " . fgets($fsock, 4);
	echo "\r\n";

	$info = stream_get_meta_data($fsock);
	if($info['timed_out']) {
		exit("timeout\r\n");
 	}
}
```
That shows zzz ten times.

In https://github.com/sapzxc/phpseclib/commit/63f49ac52a91e5da2ef9ee88babf6b47ed3009b4 you removed `stream_set_timeout($this->fsock, $sec, $usec);` so let's try adding that with $sec = 10:

```
<?php
$fsock = fsockopen('127.0.0.1', 9999);
stream_set_timeout($fsock, 10);
$i = 0;
while (!feof($fsock)) {
	$i++;
	echo "$i. " . fgets($fsock, 4);
	echo "\r\n";

	$info = stream_get_meta_data($fsock);
	if($info['timed_out']) {
		exit("timeout\r\n");
 	}
}
```

That shows zzz ten times.

The commit message for https://github.com/sapzxc/phpseclib/commit/63f49ac52a91e5da2ef9ee88babf6b47ed3009b4 says "using stream_select instead " even tho your actual commit doesn't do that. None-the-less let's try that:
```php
<?php
$fsock = fsockopen('127.0.0.1', 9999);
$i = 0;
while (!feof($fsock)) {
	$read = array($fsock);
	$write = $except = null;

	stream_select($read, $write, $expect, 10);

	$i++;
	echo "$i. " . fgets($fsock, 4);
	echo "\r\n";

	$info = stream_get_meta_data($fsock);
	if($info['timed_out']) {
		exit("timeout\r\n");
 	}
}
```

Still displays ten zzz's.

So let's try to make it cumulative using the same technique that I implemented in my branch but using `stream_get_meta_data`:

```php
<?php
$fsock = fsockopen('127.0.0.1', 9998);
$i = 0;
$curTimeout = 10;
while (!feof($fsock)) {
	$read = array($fsock);
	$write = $except = null;

	$start = microtime(true);
	$sec = floor($curTimeout);
	$usec = 1000000 * ($curTimeout - $sec);

	stream_select($read, $write, $expect, $sec, $usec);

	$elapsed = microtime(true) - $start;
	$curTimeout-= $elapsed;

	$i++;
	echo "$i. " . fgets($fsock, 4);
	echo "\r\n";

	$info = stream_get_meta_data($fsock);
	if($info['timed_out']) {
		exit("timeout\r\n");
 	}
}
```

It still displays zzz ten times for me but after the fifth time I start getting this error too:

```
PHP Warning:  stream_select(): The seconds parameter must be greater than 0 in /path/to/filename.php on line 13
```

Why isn't stream_get_meta_data working as expected? idk. But then again I guess I don't need to know as the following *does* work (and doesn't use stream_get_meta_data):

```php
<?php
$fsock = fsockopen('127.0.0.1', 9998);
$i = 0;
$curTimeout = 10;
while (!feof($fsock)) {
	if ($curTimeout < 0) {
		exit("timeout\r\n");
	}

	$read = array($fsock);
	$write = $except = null;

	$start = microtime(true);
	$sec = floor($curTimeout);
	$usec = 1000000 * ($curTimeout - $sec);

	stream_select($read, $write, $expect, $sec, $usec);

	$elapsed = microtime(true) - $start;
	$curTimeout-= $elapsed;

	$i++;
	echo "$i. " . fgets($fsock, 4);
	echo "\r\n";
}
```

bantu:
I believe this has been replaced with #639 .

PR #639, opened 2015-03-20
improve error handling with protocol version exchange

terrafrost:
See #634 and https://tools.ietf.org/html/rfc4253#section-4.2

### DIFF
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -991,8 +991,6 @@ class Net_SSH2
                 $temp = '';
             }
 
-            $read = array($this->fsock);
-
             if ($this->curTimeout) {
                 if ($this->curTimeout < 0) {
                     $this->is_timeout = true;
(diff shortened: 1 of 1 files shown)


bantu:
This patch is not good in so far that the connectionTimeout property is no longer used.

bantu:
Generally, I think it would make sense to have multiple / different timeout settings. E.g. different ones for connect() and read(). I am not sure whether that is already the case and didn't look too deep into the pull request.

terrafrost:
> This patch is not good in so far that the connectionTimeout property is no longer used.

That's a good point although I'm thinking it can just be deleted.

> Generally, I think it would make sense to have multiple / different timeout settings. E.g. different ones for connect() and read(). I am not sure whether that is already the case and didn't look too deep into the pull request.

By default it gets reset. So if the timeout is 10 seconds and it takes 8 seconds to connect the timeout for the read() will still be 10 seconds. If you want the timeout to be cumulative between read()'s and exec()'s you'll need to keep track of the elapsed time yourself and call `setTimeout` as appropriate before each read() / exec() call.

http://phpseclib.sourceforge.net/ssh/examples.html#timeout elaborates.

In #634 I said that that one didn't have cumulative timeout's but that one is a little different. The timeout's for this PR are cumulative between each stream_select call, so long as the stream_select call is for the constructor or for read() / exec(). #634 wasn't even cumulative even for stream_select.

PR #641, opened 2015-03-21
DES: fix fatal error that could occur under special circumstances

### LINKED ISSUE #632: PHP Fatal error: Cannot access self:: when no class scope is active
In composer requires:

```
"require": {
        "php": ">=5.3.3"
    },
```

But I have OSX:
PHP 5.5.14 (cli) (built: Sep  9 2014 19:09:25)
Copyright (c) 1997-2014 The PHP Group
Zend Engine v2.5.0, Copyright (c) 1998-2014 Zend Technologies

And have errors in some situations:

*PHP Fatal error:  Cannot access self:: when no class scope is active in /dev/vendor/phpseclib/phpseclib/phpseclib/Crypt/Base.php(1958) : runtime-created function on line 16*

Use case:
```
		$ssh = new \phpseclib\Net\SSH2($ip, $port, 3);
		if(!$ssh->login($user,$pass))
		{
			return $ssh;
		}
		$res = $ssh->exec('uname -a');

		$ssh->disconnect();
```

Errors happens in random situations.

terrafrost:
Fixes #632. No unit test is currently included. I'll add this to the openssl-support and setengine PR's however.

### DIFF
--- a/phpseclib/Crypt/DES.php
+++ b/phpseclib/Crypt/DES.php
@@ -1343,8 +1343,8 @@ class DES extends Base
                         $k[self::ENCRYPT][$i] = '$ke[' . $i . ']';
                         $k[self::DECRYPT][$i] = '$kd[' . $i . ']';
                     }
-                    $init_encrypt = '$ke = $self->keys[self::ENCRYPT];';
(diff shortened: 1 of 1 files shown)


kravco:
openssl merge reintroduced the issue

PR #648, opened 2015-03-29
Agent Forwarding for 2.0 branch

terrafrost:
#644 for 2.0


GrahamCampbell:
Are merges from 1.0 -> 2.0 not done anymore? Does that mean my docblock changes need to be submitted to both branches?

bantu:
@GrahamCampbell They are done (at least should be done). However, sometimes it is useful to have an additional PR for the merge, e.g. for complicated changes, because they run through CI.

terrafrost:
> Are merges from 1.0 -> 2.0 not done anymore? Does that mean my docblock changes need to be submitted to both branches?

I'm also doing it for bigger patches so it's less work for bantu to do the merges. ie. I'm resolving the conflicts so he doesn't have to. For smaller changes conflicts - if they occur at all - will be easier to resolve.

PR #655, opened 2015-04-08
support for callback function for SFTP::put function - in order to pipe ...

andrey012:
...data directly to remote server without putting it into file or keeping in memory. This can be useful particularly for dumping big databases directly to remote server.

### DIFF
--- a/phpseclib/Net/SFTP.php
+++ b/phpseclib/Net/SFTP.php
@@ -73,6 +73,11 @@ class SFTP extends SSH2
     // this value isn't really used anymore but i'm keeping it reserved for historical reasons
     const SOURCE_STRING = 2;
     /**
+     * Reads data from callback:
+     * function callback($length) returns string to proceed, null for EOF
+     */
+    const SOURCE_CALLBACK = 16;
+    /**
      * Resumes an upload
     */
     const RESUME = 4;
@@ -1718,6 +1723,8 @@ class SFTP extends SSH2
      * contain as many bytes as filename.ext does on your local filesystem.  If your filename.ext is 1MB then that is how
      * large $remote_file will be, as well.
      *
+     * Setting $mode to self::SOURCE_CALLBACK will use $data as callback function, which gets only one parameter -- number of bytes to return, and returns a string if there is some data or null if there is no more data
+     *
      * If $data is a resource then it'll be used as a resource instead.
      *
      * Currently, only binary mode is supported.  As such, if the line endings need to be adjusted, you will need to take
@@ -1797,7 +1804,12 @@ class SFTP extends SSH2
         }
 
         // http://tools.ietf.org/html/draft-ietf-secsh-filexfer-13#section-8.2.3
+        $callback = false;
         switch (true) {
+            case $mode & self::SOURCE_CALLBACK;
+                $callback = $data;
+                // do nothing
+                break;
             case is_resource($data):
                 $mode = $mode & ~self::SOURCE_LOCAL_FILE;
                 $fp = $data;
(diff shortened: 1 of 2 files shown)


terrafrost:
There's another PR that does something similar:

https://github.com/phpseclib/phpseclib/pull/584

I like your method better however. That said, the other PR adds a `is_callable` check. It'd be good if your PR did the same thing.

andrey012:
terrafrost,

I think my PR and #584 <https://github.com/phpseclib/phpseclib/pull/584> are different.
#584 <https://github.com/phpseclib/phpseclib/pull/584> is about reporting progress to the application.
My PR is about streaming data in the manner Curl can do - instead of giving a file or a string you can give a callback and Net will call callback to get next piece of data.
For me the reason was simple - there is a server with 80 Gb database and only 20 Gb of free space, when I try to pack the database through mysqldump | gzip  - it can hit the free space. This way I managed to stream things without taking any of disk space.

I've made cosmetic tweaks as you requested, in the same branch.

Thank you.

AG

terrafrost wrote:

terrafrost:
Good point. However, none-the-less, I still don't like how that PR add's a sixth parameter to an already overloaded method. Although that said I guess I'm willing to live with that patch with the understanding that that will be removed from later (BC breaking) versions of phpseclib. eg.

```
$sftp = new Net_SFTP(...);
$sftp->login(...); 

$fp = $sftp->open('filename.ext', 'w');
$fp->write('whatever');
$fp->close();
```

At that point what I'd probably do is remove $mode all together along with $start, $local_start and $callback.

If you want to efficiently upload from a file you can do this:

```
$local = fopen('filename.ext', 'r');
$remote = $sftp->open('filename.ext', 'w');
while (!feof($local)) {
    $remote->write(fread($local, 1024));
}
$remote->close();
fclose($local);
```

Back to this PR tho...  currently phpseclib uses user_error - not Exceptions. I think exceptions are actually superior and I want to change all of phpseclib's user_error's to exceptions but I also want phpseclib to be consistent.

I suppose that's a change I can make just as easily myself as well. I'll try to do this and two other PR's tomorrow or the day after.

andrey012:
terrafrost wrote:
>
> However, none-the-less, I still don't like how that PR add's a sixth parameter to an already overloaded method.
>
I think I didn't add any new parameters to any function.
Do you mean SOURCE_CALLBACK constant?
>
> Although that said I guess I'm willing to live with that patch with the understanding that that will be removed from later (BC breaking) versions of phpseclib. eg.
>
> |$sftp = new Net_SFTP(...);
> $sftp->login(...);
>
> $fp = $sftp->open('filename.ext', 'w');
> $fp->write('whatever');
> $fp->close();|
|This way is better, of course. |As for me - I'd leave BC where possible.

AG

terrafrost:
> I think I didn't add any new parameters to any function.
> Do you mean SOURCE_CALLBACK constant?

I was talking about the other PR.

> |This way is better, of course. |As for me - I'd leave BC where possible.

phpseclib 0.x / 1.x would retain BC. phpseclib 2.0.x would retain BC as well. Maybe the version of phpseclib where this new API lived would be 2.1.x or 3.x or something idk.

Basically, phpseclib has multiple concurrent versions that are being supported. Kinda like how PHP was concurrently supporting PHP 4 / 5 for a time and how phpBB is concurrently supporting 3.0 and 3.1. Although that said I'm tentatively thinking I'll support phpseclib 0.x / 1.x a good deal longer but who knows. I'm also not going to come up with a ten year roadmap for phpseclib either.

andrey012:
Sorry, I didn't notice, that there is 1.0 version.
I've made same changes there and will create another PR for 1.0 branch.

Also, I've changed throw with user_error as you recommended.

AG

terrafrost wrote:

PR #668, opened 2015-04-27
Boolean comparisons was missing an equal sign

e3betht:
I ran the code through Code Climate and saw that there were three Boolean comparisons that were not using the identical operator (which is always recommended to avoid unexpected or incorrect results). I updated this to use the identical operator and added the Code Climate badge to the readme.

### DIFF
--- a/phpseclib/Crypt/RSA.php
+++ b/phpseclib/Crypt/RSA.php
@@ -2464,7 +2464,7 @@ class RSA
         // be output.
 
         $emLen = ($emBits + 1) >> 3; // ie. ceil($emBits / 8)
-        $sLen = $this->sLen == false ? $this->hLen : $this->sLen;
+        $sLen = $this->sLen === false ? $this->hLen : $this->sLen;
 
         $mHash = $this->hash->hash($m);
         if ($emLen < $this->hLen + $sLen + 2) {
@@ -2502,7 +2502,7 @@ class RSA
         // be output.
 
         $emLen = ($emBits + 1) >> 3; // ie. ceil($emBits / 8);
-        $sLen = $this->sLen == false ? $this->hLen : $this->sLen;
+        $sLen = $this->sLen === false ? $this->hLen : $this->sLen;
 
         $mHash = $this->hash->hash($m);
         if ($emLen < $this->hLen + $sLen + 2) {
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # phpseclib - PHP Secure Communications Library
 
-[![Build Status](https://secure.travis-ci.org/phpseclib/phpseclib.png?branch=php5)](http://travis-ci.org/phpseclib/phpseclib)
+[![Build Status](https://secure.travis-ci.org/phpseclib/phpseclib.png?branch=php5)](http://travis-ci.org/phpseclib/phpseclib) [![Code Climate](https://codeclimate.com/github/phpseclib/phpseclib/badges/gpa.svg)](https://codeclimate.com/github/phpseclib/phpseclib)
 
 MIT-licensed pure-PHP implementations of an arbitrary-precision integer
 arithmetic library, fully PKCS#1 (v2.1) compliant RSA, DES, 3DES, RC4, Rijndael,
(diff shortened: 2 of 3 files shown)


terrafrost:
It is unclear to me why this is failing the unit tests. I'll have to do some experimentation before I consider merging it.

e3betht:
Let me know if I need to fix anything. Is it possible that one of your Boolean compares is not being compared against a Boolean and that is causing it to fail? I can cast the values as Booleans before they are compared if that would help.

bantu:
The tests failure seem to have been introduced by #655
The main question here is whether we want to add the Code Climate badge. ;-)

e3betht:
I'm glad it wasn't my changes, and I think you know my vote on the badge already. ;)

terrafrost:
I'm good with it being merged!

terrafrost:
I merged this but then I undid it.

I guess I should have checked more thoroughly before I merged but...

consider Crypt/DES.php for instance. It's dinging us for having excessively complex code and while I agree that it kinda is, the code also yields performance gains of 25%. See the discussion here:

https://github.com/phpseclib/phpseclib/pull/94#discussion_r4013686

And a lot of the variables come from the reference implementations for DES / whatever. eg.

http://csrc.nist.gov/publications/fips/fips46-3/fips46-3.pdf#page=14

Altho it'd be nice if they used more descriptive variables sometimes people just don't and if what you're doing is mostly an adaptation of the algorithm's base implementation than idk that changing variable names is a good idea.

I think static code analysis can be a good thing but it's results have to be taken in stride with everything else and in lieu of that idk that I like the idea of displaying a red (ie. broken badge vs. a green "working" badge) when some of the things it flags are arguably not even issues that ought to be "fixed".

That said, I did send codeclimate.com an email tho about it misidentifying similar pieces of code as being identical when they're not. Hopefully it helps :)

e3betht:
I understand what you mean. If I could help you update the code to 
improve your score, would you put the badge back? Or is it less the 
grade and more the false positives that are affecting this decision?


On 5/7/15 9:21 AM, terrafrost wrote:
>
> I merged this but then I undid it.
>
> I guess I should have checked more thoroughly before I merged but...
>
> consider Crypt/DES.php for instance. It's dinging us for having 
> excessively complex code and while I agree that it kinda is, the code 
> also yields performance gains of 25%. See the discussion here:
>
> #94 (comment) 
> <https://github.com/phpseclib/phpseclib/pull/94#discussion_r4013686>
>
> And a lot of the variables come from the reference implementations for 
> DES / whatever. eg.
>
> http://csrc.nist.gov/publications/fips/fips46-3/fips46-3.pdf#page=14
>
> Altho it'd be nice if they used more descriptive variables sometimes 
> people just don't and if what you're doing is mostly an adaptation of 
> the algorithm's base implementation than idk that changing variable 
> names is a good idea.
>
> I think static code analysis can be a good thing but it's results have 
> to be taken in stride with everything else and in lieu of that idk 
> that I like the idea of displaying a red (ie. broken badge vs. a green 
> "working" badge) when some of the things it flags are arguably not 
> even issues that ought to be "fixed".
>
> That said, I did send codeclimate.com an email tho about it 
> misidentifying similar pieces of code as being identical when they're 
> not. Hopefully it helps :)
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/phpseclib/phpseclib/pull/668#issuecomment-99884099>.
>

-- 
Beth Tucker Long
Treeline Design, LLC
807 Arbor Vitae Place
Verona, WI 53593
608-770-6677
http://www.TreelineDesign.com

phpseclib/Crypt/RSA.php
```diff
@@ -2464,7 +2464,7 @@ function _emsa_pss_encode($m, $emBits)
         // be output.
 
         $emLen = ($emBits + 1) >> 3; // ie. ceil($emBits / 8)
-        $sLen = $this->sLen == false ? $this->hLen : $this->sLen;
+        $sLen = $this->sLen === false ? $this->hLen : $this->sLen;
```
bantu: One of our users reported at #768 that this may have actually broken functionality. It sounds like, logically, this should have been `$sLen = $this->sLen ? $this->sLen : $this->hLen;` instead. Note that s and h was swapped.

PR #672, opened 2015-05-02
Fix callback parameter conflicts of SFTP put() method.

andrey012:


bantu:
This looks good to me. Thanks a bunch.

bantu:
@terrafrost There seems to be something wrong with the branching. 1.0 is not merged into 2.0 and produces a conflict. Please avoid this in the future. 1.0 needs to be merged into 2.0 at all times, 2.0 needs to be merged into master at all times.

PR #673, opened 2015-05-04
RSA: Add support for calculating a public key's fingerprint

Lukas-W:
This adds the function `getPublicKeyFingerprint()` to `Crypt_RSA`:
```php
$rsa = new Crypt_RSA();
$rsa->loadPublicKey(file_get_contents('rsa.pub'));
echo $rsa->getPublicKeyFingerprint('md5');
echo $rsa->getPublicKeyFingerprint('sha256');
```
returns something like
```
c1:b1:30:29:d7:b8:de:6c:97:77:10:d7:46:41:63:87
N9sV2uSNZEe8TITODku0pRI27l+Zk0IY0TrRTw3ozwM
```
Calling this function outputs the same as running `ssh-keygen -lE sha256 f id.pub` or `ssh-keygen -lE md5 -f id.pub`.

The MD5 format is specified in https://tools.ietf.org/html/rfc4716#section-4.

### DIFF
--- a/phpseclib/Crypt/RSA.php
+++ b/phpseclib/Crypt/RSA.php
@@ -1706,6 +1706,42 @@ class RSA
     }
 
     /**
+     * Returns the public key's fingerprint
+     *
+     * The public key's fingerprint is returned, which is equivalent to running `ssh-keygen -lf rsa.pub`. If there is
+     * no public key currently loaded, false is returned.
+     * Example output (md5): "c1:b1:30:29:d7:b8:de:6c:97:77:10:d7:46:41:63:87" (as specified by RFC 4716)
+     *
+     * @access public
+     * @param String $algorithm The hashing algorithm to be used. Valid options are 'md5' and 'sha256'. False is returned
+     * for invalid values.
+     */
+    public function getPublicKeyFingerprint($algorithm = 'md5')
+    {
+        if (empty($this->modulus) || empty($this->publicExponent)) {
+            return false;
+        }
+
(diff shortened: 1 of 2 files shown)


terrafrost:
I like it with the caveats I mentioned in-line in the diff. If you could address those that'd be awesome - thanks!

Lukas-W:
Sure - I addressed your points in c46890b00cd3c2258896af84ab6b1156e65d1fa1.

terrafrost:
Err...  I forgot the md5 lol. Should be this:

```php
substr(chunk_split(md5($RSAPublicKey), 2, ':'), 0, -1);
```

Time permitting I may do this myself later tonight.

Lukas-W:
Ooops… Fixed. :bike:

bantu:
Please document the return value.

bantu:
Why only md5 and sha256?

Lukas-W:
> Why only md5 and sha256?

Well originally I only planned to implement md5, as specified in RFC 4716, see https://tools.ietf.org/html/rfc4716#section-4. Then I noticed `ssh-keygen -lf` also supports sha256, and even uses it by default in newer releases, so I included that as well. 

What do you suggest? Should I just pass the `$algorithm` to the `Hash` constructor and support every algorithm `Hash` supports?

terrafrost:
The md5 and sha256 formats differ already so idk that adding more would be a good thing, lest we just be guessing as to which format the new hashes ought to be in.

Anyway, I've merged this.

Thanks!

PR #683, opened 2015-05-07
README: rm codeclimate badge

terrafrost:
it's giving us a low score for code who's complexity can't easily be changed
and dinging us for "short variables" that are based on the RFC's of the undyling
encryption algorithms


bantu:
;-)

PR #689, opened 2015-05-23
ANSI updates for 1.0 branch

terrafrost:


phpseclib/File/ANSI.php
```diff
@@ -213,6 +181,16 @@ class File_ANSI
...
     function File_ANSI()
     {
+        $attr_cell = new stdClass();
+        $attr_cell->bold = false;
+        $attr_cell->underline = false;
+        $attr_cell->blink = false;
+        $attr_cell->background = 'black';
+        $attr_cell->foreground = 'white';
+        $attr_cell->reverse = false;
+        $this->base_attr_cell = clone($attr_cell);
```
bantu: clone is PHP5 only? and an operator

terrafrost: Yup - but it's one that can be remedied with PHP_Compat:

https://github.com/pear/PHP_Compat/blob/master/Compat/Function/clone.php

In the 2.0 branch I did `clone $attr_cell` instead of `clone($attr_cell)` since I think that looks better.

bantu: We need to update https://github.com/phpseclib/phpseclib/blob/1.0/composer.json#L60 then.

terrafrost: Is https://github.com/terrafrost/phpseclib/commit/3cb685337a871577cb20fa482efa077645025b3a what you had in mind?

bantu: Yes

phpseclib/File/ANSI.php
```diff
@@ -416,28 +372,22 @@ function appendString($source)
                                                 case 47: $back = 'white'; break;
 
                                                 default:
-                                                    user_error('Unsupported attribute: ' . $mod);
+                                                    //user_error('Unsupported attribute: ' . $mod);
```
bantu: ?

phpseclib/File/ANSI.php
```diff
@@ -416,28 +372,22 @@ function appendString($source)
...
-                                            if ($this->color) {
-                                                $this->attrs[$this->y][$this->x] = '</span>' . $this->attrs[$this->y][$this->x];
-                                            }
-                                            $this->color = true;
                                     }
                                 }
                                 break;
                             default:
-                                user_error("{$this->ansi} unsupported\r\n");
+                                //user_error("{$this->ansi} is unsupported\r\n");
```
bantu: ?

phpseclib/File/ANSI.php
```diff
@@ -445,12 +395,32 @@ function appendString($source)
...
+                            $this->x,
+                            1
+                        );
+                    }
+                    break;
                 case "\x0F": // shift
                     break;
                 case "\x1B": // start ANSI escape code
+                    $this->tokenization[count($this->tokenization) - 1] = substr($this->tokenization[count($this->tokenization) - 1], 0, -1);
+                    //if (!strlen($this->tokenization[count($this->tokenization) - 1])) {
```
bantu: ?

phpseclib/File/ANSI.php
terrafrost: I commented it out because, honestly, trying to figure out what all the ANSI escape codes do is difficult. Some of them I just do `break` and that's it because they don't apply (eg. show the cursor position, etc) but others...  I'm not sure what they do. I don't want to add a `case` statement for them that I `break` since I don't know what they do but neither do they seem necessary to create the desired visual effects.

I didn't delete the line because it could still be useful for diagnostic purposes.

phpseclib/File/ANSI.php
terrafrost: See above.

phpseclib/File/ANSI.php
terrafrost: It's useful for diagnostic purposes. If I want to see what ANSI escape codes are being utilized I can either try to figure that out by hand or I can utilize the built in ANSI parsing routines in this class and have it spit them out to me, interleaved with the non-ANSI escape codes.

It's not documented anywhere and doesn't have a public API method for it. Pretty much it's not a feature that's of much benefit unless you're intending on developing this file. I suppose I could remove it along with the user_error's and put them in some sort of private notes of mine but that would make my life harder cause I'd have to consult those notes to copy / paste the code (it's easier to remember that I have `$this->tokenization` in the file than it is for me to remember that whole line) and it's not like removing that line would yield significant performance gains or anything.

So, basically, it's not hurting anyone and it's helping me. And the fact that it isn't documented means that it's not an officially supported feature and that semantic versioning can be maintained without doing a new "major" release if if this "feature" is removed or changed.

PR #692, opened 2015-05-26
add "none" encryption mode for 1.0 branch

### LINKED ISSUE #687: Crypt_RSA disable padding
Crypt_RSA does padding automatically, even with: setEncryptionMode(CRYPT_RSA_ENCRYPTION_PKCS1);
disablePadding() isn't available under Crypt_RSA.

How do I disable padding?

terrafrost:
Fixes #687

### DIFF
--- a/tests/Unit/Crypt/RSA/ModeTest.php
+++ b/tests/Unit/Crypt/RSA/ModeTest.php
@@ -38,6 +38,6 @@ U9VQQSQzY1oZMVX8i1m5WUTLPz2yLJIBQVdXqhMCQBGoiuSoSjafUhV7i1cEGpb88h5NBYZzWXGZ
         $this->assertEquals($result, $expected);
 
         $rsa->loadKey($privatekey);
-        $this->assertEquals(trim($rsa->decrypt($a), "\0"), $plaintext);
+        $this->assertEquals(trim($rsa->decrypt($result), "\0"), $plaintext);
     }
 }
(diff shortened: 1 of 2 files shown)


tests/Unit/Crypt/RSA/ModeTest.php
```diff
@@ -0,0 +1,30 @@
...
+        $rsa = new Crypt_RSA();
+
+        extract($rsa->createKey());
+        $rsa->loadKey($publickey);
+
+        $rsa->setEncryptionMode(CRYPT_RSA_ENCRYPTION_NONE);
+        $a = $rsa->encrypt($plaintext);
+        $b = $rsa->encrypt($plaintext);
+
+        $this->assertEquals($a, $b);
```
bantu: Please consider just fixing the expected value (first parameter) here.

tests/Unit/Crypt/RSA/ModeTest.php
```diff
@@ -0,0 +1,30 @@
...
+
+class Unit_Crypt_RSA_ModeTest extends PhpseclibTestCase
+{
+    public function testEncryptionModeNone()
+    {
+        $plaintext = 'a';
+
+        $rsa = new Crypt_RSA();
+
+        extract($rsa->createKey());
```
bantu: Not a fan of this at all as it hides data flows. Assign the result to an array instead?

tests/Unit/Crypt/RSA/ModeTest.php
terrafrost: In order to do that the key can't be random, which it currently is. I can hard-code the key tho. I guess I'll do that in another commit shortly.

Issue #706, opened 2015-06-18
Delete empty folder works but does not update stat cache


PVince81:
See https://github.com/owncloud/core/issues/17019#issuecomment-113188650
This happens in phpseclib 0.3.9.

When deleting an empty folder, it is properly deleted remotely, but for some reason the call returns 'NET_SFTP_STATUS_FAILURE' so it will skip updating the stat cache.

This doesn't happen for non-empty folders.

@bantu

PR #709, opened 2015-06-21
SFTP: stat cache updates

terrafrost:
Two new unit tests have been added that show the conditions under which phpseclib failed prior to this commit.


bantu:
This is supposed to fix #706 right?

terrafrost:
No - this fixes some issues that I was emailed about - #706 still needs to be looked at.

PR #714, opened 2015-06-25
Add support for RFC 4419: diffie-hellman-group-exchange-sha{1,256}

### LINKED ISSUE #609: Add diffie-hellman-group-exchange-sha256 key exchange algorithm
It seems that some server configurations are now upgraded to use only diffie-hellman-group-exchange-sha256 or some more advanced key exchange algorithms than the diffie-hellman-group1-sha1 or diffie-hellman-group14-sha1 currently supported by phpseclib.
Would it be possible to add diffie-hellman-group-exchange-sha256 ?

bantu:
Fix #609

### DIFF
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -207,6 +207,33 @@ class Net_SSH2
     var $kex_algorithms = false;
 
     /**
+     * Minimum Diffie-Hellman Group Bit Size in RFC 4419 Key Exchange Methods
+     *
+     * @see Net_SSH2::_key_exchange()
+     * @var Integer
+     * @access private
+     */
+    var $kex_dh_group_size_min = 1536;
+
+    /**
+     * Preferred Diffie-Hellman Group Bit Size in RFC 4419 Key Exchange Methods
+     *
+     * @see Net_SSH2::_key_exchange()
+     * @var Integer
+     * @access private
+     */
+    var $kex_dh_group_size_preferred = 2048;
+
+    /**
+     * Maximum Diffie-Hellman Group Bit Size in RFC 4419 Key Exchange Methods
+     *
+     * @see Net_SSH2::_key_exchange()
+     * @var Integer
+     * @access private
+     */
+    var $kex_dh_group_size_max = 4096;
+
+    /**
      * Server Host Key Algorithms
      *
      * @see Net_SSH2::getServerHostKeyAlgorithms()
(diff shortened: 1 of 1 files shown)


bantu:
@terrafrost This should be ready for review.

bantu:
This does not cleanly merge into 2.0. I can prepare a patch for 2.0 after this has received initial review.

terrafrost:
Added some feedback. Otherwise the change looks good!

bantu:
> Maybe move diffie-hellman-group-exchange-sha1 and diffie-hellman-group-exchange-sha256 up at the top? The order determines which algorithms are used and it's my understanding group exchange is supposed to be more secure than either diffie-hellman-group1-sha1 or diffie-hellman-group14-sha1.

In that case, should group1 and group14 also swap places? This would yield
```php
        static $kex_algorithms = array(
            'diffie-hellman-group-exchange-sha256', // RFC 4419
            'diffie-hellman-group-exchange-sha1', // RFC 4419
            'diffie-hellman-group14-sha1', // REQUIRED
            'diffie-hellman-group1-sha1', // REQUIRED
        );
```

terrafrost:
It's probably doing diffie-hellman-group1-sha1 preferentially over diffie-hellman-group14-sha1 for speed purposes. Sure, group14 is faster but if it takes 4x as long (ie. 10 seconds to 40 seconds) it might be worthwhile to do the less secure approach.

I guess what'd be good is benchmarks of each of the dh methods with the worst possible BigInteger setup (eg. internal mode BigInteger without the OpenSSL enhancements). If one is slower than the other by 10% it's prob not a huge issue but 400%...  that's a bit harder to miss.

bantu:
The same speed/security tradeoff applies to the newly introduced ones here. Basically, assuming we get the group size we request, all sizes besides diffie-hellman-group1-sha1 are 2048 bits, which is 1024 bits.

Maybe the order should be changed based on the BigInteger engine. More complex, but probably more secure in the default case.

bantu:
Your comments here were lost because you made them on the commit instead of the PR and I rebased/amended the commit.

bantu:
The constants are defined individually now. I've left the KEX order as is. We can do benchmarks later and swap them around. Let's get this merged and then release 0.3.11?

terrafrost:
> Maybe the order should be changed based on the BigInteger engine. More complex, but probably more secure in the default case.

I agree - that would be good.

> Let's get this merged and then release 0.3.11?

The next release of the 1.0 release is 1.0 since OpenSSL support has been developed into the symmetric key algorithms.

I'd like to do a concurrent release of 2.0 as well. Last time I did a release I tagged the 2.0 branch with a release version but that tag got undone. Are we going to start doing releases of 2.0 and tagging them accordingly? Or is the idea that the 2.0 branch will be tagless?

bantu:
Ah yeah. 1.0.0 and 2.0.0 is fine too. Lets switch to PSR2 before that then.
Is this PR fine now?

terrafrost:
Yah - this PR is fine.

For PSR2 (this might be better in a separate discussion) maybe we can replace the var's with public's instead of protected / private.

http://aperiplus.sourceforge.net/visibility.php explains why private shouldn't be used and calls into question the virtue of using protected (eg. it's solving a problem that was never a problem to begin with). Plus, in the case of phpseclib, it could actually break things. Net/SCP.php calls methods that are prefixed with underscores and denoted as being private via the phpdoc comments. And sometimes I'll tell people to use "private" methods as workarounds. Asking people to modify Net/SSH2.php as a prerequisite to run diagnostic code would slow things down and in some cases could make it so I couldn't get any sort of feedback at all.

bantu:
Will send a PR for PSR2 shortly.

bantu:
Not sending a 2.0 PR for this since the conflict seems rather easy to resolve.

PR #726, opened 2015-07-03
SFTP: delete stat cache for recursive deletes

### LINKED ISSUE #706: Delete empty folder works but does not update stat cache

terrafrost:
Fixes #706

### DIFF
--- a/tests/Functional/Net/SFTPUserStoryTest.php
+++ b/tests/Functional/Net/SFTPUserStoryTest.php
@@ -543,13 +543,23 @@ class Functional_Net_SFTPUserStoryTest extends PhpseclibFunctionalTestCase
 
     /**
      * @depends testRmDirScratch
+     * @group github706
      */
     public function testDeleteRecursiveScratch($sftp)
     {
+        $this->assertInternalType(
+            'array',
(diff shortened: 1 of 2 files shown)


tests/Functional/Net/SFTPUserStoryTest.php
```diff
@@ -543,14 +543,24 @@ public function testRmDirScratch($sftp)
 
     /**
      * @depends testRmDirScratch
+     * @group github706
```
bantu: This test does not seem to be a regression test for #706 as it does not fail when the changes above are undone.

terrafrost: It does for me on Travis CI:

https://travis-ci.org/terrafrost/phpseclib/builds/69507514

terrafrost:
I meant to do this for the 1.0 branch - not for the master branch.

PR #730, opened 2015-07-04
SFTP: don't update stat cache when it's been disabled

### LINKED ISSUE #712: disableStatCache
Hi,

I downloaded the latest 1.0 branch and found a bug, but this bug depends on disableStatCache.

Although my program disables the cache with ($sftp->disableStatCache),
the internal function Net_SFTP::_update_stat_cache/_remove_from_stat_cache
is called very often.

$rcode = $sftp->disableStatCache();
$files = $sftp->rawlist();
foreach($files as $name => $stat) {    ...    }

1.
Either this is by design. 
Then I will send you the bug report as soon as you will decide if this issue is a bug or by design.

2.
Or this is a bug, because if I disable the cache, the cache should be frozen
and no update (_update_stat_cache) or remove is allowed.
In this case my program runs without errors because of my workaround:
function _update_stat_cache($path, $value)
{
    if ($this->use_stat_cache===false) return; #wordaround disableStatCache
    ....

3.
I personally vote for point 2.

Best Regards
Michael

terrafrost:
Fixes #712

### DIFF
--- a/phpseclib/Net/SFTP.php
+++ b/phpseclib/Net/SFTP.php
@@ -1062,6 +1062,10 @@ class Net_SFTP extends Net_SSH2
      */
     function _update_stat_cache($path, $value)
     {
+        if ($this->use_stat_cache === false) {
+            return;
+        }
+
         // preg_replace('#^/|/(?=/)|/$#', '', $dir) == str_replace('//', '/', trim($path, '/'))
         $dirs = explode('/', preg_replace('#^/|/(?=/)|/$#', '', $path));
 


bantu:
Can we test this easily? Maybe using another instance of Net_SFTP?

bantu:
Patch looks good to me. It's also described in #712. Merging. Feel free to make another PR with tests, though.

terrafrost:
I'm not really sure how we'd test this without calling "private" members of Net_SFTP...

PR #732, opened 2015-07-08
[1.0] Code Sniffer: Switch from PEAR with exceptions to PSR2 with exceptions

bantu:


terrafrost:
I'm okay with most of the changes but...  some of the changes take lines like this:

```php
define('CRYPT_HASH_MODE_INTERNAL', 1);
define('CRYPT_HASH_MODE_MHASH',    2);
define('CRYPT_HASH_MODE_HASH',     3);
```

...and replace them with this:

```php
define('CRYPT_HASH_MODE_INTERNAL', 1);
define('CRYPT_HASH_MODE_MHASH', 2);
define('CRYPT_HASH_MODE_HASH', 3);
```

It does that in a number of places and I don't think that's an improvement. I mean, I don't think it's a huge imparement, either, but I do think it's a mild imparement to the readibility of the code.

Also, stuff like this:

```php
                    case "\x1B[H": // Move cursor to upper left corner
                        $this->old_x = $this->x;
                        $this->old_y = $this->y;
                        $this->x = $this->y = 0;
                        break;
```

Changing that to this:

```php
                    case "\x1B[H":
                        // Move cursor to upper left corner
                        $this->old_x = $this->x;
                        $this->old_y = $this->y;
                        $this->x = $this->y = 0;
                        break;
```

I don't think that's a good change. Previously it was clear that "move cursor to upper left corner" applied to the whole case statement. Now it's unclear. Does it apply t othe whole case statement or just the `$this->old_x = $this->x` line?

If comments need to be on their own line this would be better:

```php
                    // Move cursor to upper left corner
                    case "\x1B[H":
                        $this->old_x = $this->x;
                        $this->old_y = $this->y;
                        $this->x = $this->y = 0;
                        break;
```

Assuming it's even necessary to begin with.

Also...

https://github.com/bantu/phpseclib/commit/edc948c317deed1b506066f3973a5b3016ca3843#diff-6fc70cd3cd3475a879577a0757ead792R376

idk that that's a good change. Sure, it may be bringing it more in-line with PSR2 but it's blowing up 16 lines into 48. ie. it's tripling the size. And altho I don't think concise code is always an improvement I think sometimes it is. Larger code snippets are more likely to intimidate people and make them gloss over that whole block of code. And in the process they may bulk ignore other stuff too.

terrafrost:
Old:
```php
                    case $decoded['type'] < 18: // FILE_ASN1_TYPE_NUMERIC_STRING == 18
                    case $decoded['type'] > 30: // FILE_ASN1_TYPE_BMP_STRING == 30
```

New:
```php
                    case $decoded['type'] < 18:
                        // FILE_ASN1_TYPE_NUMERIC_STRING == 18
                    case $decoded['type'] > 30:
                        // FILE_ASN1_TYPE_BMP_STRING == 30
```
Do not want.

Old:
```php
                for ($j = 1; ($temp >> $j) & 1; ++$j);
```

New:
```php
                for ($j = 1; ($temp >> $j) & 1;
                ++$j) {
                }
```
idk that I'm a big fan of adding {}'s for otherwise empty for loops but I can live with that (hence why I didn't mention it in my last post). But having the ++$j on it's own line? That doesn't even make any sense.

Old:
```php
        for ($i = 0; $i < count($encryption_algorithms) && !in_array($encryption_algorithms[$i], $this->encryption_algorithms_server_to_client); $i++);
```

New:
```php
        for ($i = 0; $i < count($kex_algorithms) && !in_array($kex_algorithms[$i], $this->kex_algorithms);
        $i++) {
        }
```
If PSR2 has the third part of for loops be on it's own line I'd rather us just remove all for loops and replace them all with while loops.

bantu:
The for loop thing is weird and probably a bug in the automatic syntax changer. The validator still accepts it after changing it back.

terrafrost:
Any update on this? Per https://github.com/phpseclib/phpseclib/pull/714#issuecomment-119139352 it seemed like we were on track for a new release pending this.

So to recap, there are three types of changes I had issues with.

1. Some white space changes. To remedy this we can either (1) add `@codingStandardsIgnoreStart` before and after as is done at https://github.com/phpseclib/phpseclib/blob/d75f703c0ae93dbd99dab5c259c98f47759c87af/phpseclib/Math/BigInteger.php#L3293 or we could (2) make an exception to this in code-sniffer-ruleset.xml. I think the first option would be the best.
2. Comments on their own lines. We can (1) add an exception to this in code-sniffer-ruleset.xml or we can (2) move comments to the line above instead of the line after. I'm in favor of option # 1.
3. For loops having new lines introduced in the middle of them. This is probably a bug in the syntax changer so we just gotta remove it from the commit!

bantu:
About 2. https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md says that "Comments and documentation blocks" are not covered.

bantu:
The for loops in SSH2 we should probably replace with a method as all loops basically do the same thing and this likely is not performance-critical.

bantu:
@terrafrost Updated.

phpseclib/Crypt/Hash.php
```diff
@@ -278,20 +278,20 @@ function setHash($hash)
             case 'md2':
                  $this->b = 16;
```
bantu: This actually looks wrong too.

terrafrost:
> The for loops in SSH2 we should probably replace with a method as all loops basically do the same thing and this likely is not performance-critical.

I have no objection to that. Feel free to do a PR?

Regarding https://github.com/bantu/phpseclib/commit/8524560ccafc9db3b9a999a7538584a1576cba0e ...  see point # 1 in https://github.com/phpseclib/phpseclib/pull/732#issuecomment-121436452 .

phpseclib/Crypt/Hash.php
terrafrost: Good catch. It seems to be adding in an extra space. This can be seen more clearly in the full code snippet:

```php
        switch ($hash) {
            case 'md2':
                 $this->b = 16;
                 $this->hash = array($this, '_md2');
                break;
```

bantu:
@terrafrost Ah yeah, I made two extra commits for the whitespace changes, so we can more easily (re-)evaluate whether we want to have them. My recommendation would be to have both.

affe884c70b6047f4413b5044469953f06e751f9 is rather small. Could use an ignore statement if we want to keep rightRotate and rightShift aligned.

8524560ccafc9db3b9a999a7538584a1576cba0e seems acceptable to me as well. Keeping things aligned with doc blocks or comments in between does not seem to make too much sense. Also, it's probably a tradeoff. Specifically, you will probably have to respace all existing spacing when adding another line that is longer (e.g. longer constant name). Maybe we could also add an ignore statement for values like 0x00000001.

terrafrost:
So I guess `PSR2.Methods.FunctionCallSignature.SpaceAfterOpenBracket` is making it so this is invalid?:

```php
for (...){
```

...and instead this has to be done?:

```php
for (...) {
```

I have no issue with that one.

It seems like `Generic.Functions.FunctionCallArgumentSpacing.TooMuchSpaceAfterComma` is the one I'm disagreeing with more. Mainly for the define()'s tho. I don't think it's a huge issue for rightRotate and rightShift since, in those cases, the only difference is a single space. Doing `@codingStandardsIgnoreStart` would be good for that.

bantu:
@terrafrost No, `PSR2.Methods.FunctionCallSignature.SpaceAfterOpenBracket` seems to only affect what is in the relevant commit. It does not seem to be related to for loops.

bantu:
The for loops example you made is covered by `Squiz.ControlStructures.ControlSignature.SpaceAfterCloseParenthesis`.

terrafrost:
Looks good!

bantu:
Rebased.

PR #737, opened 2015-07-15
SSH2: limit the size of data that expect() looks at

### LINKED ISSUE #670: SSH2->read performance question
Good afternoon, I have a device (Cisco ASA firewall) giving me quite a bit of grief. I found that commands with short output (60kb) were returning very quickly, but longer output (600kb) took quite a bit longer than expected to come back.

I added some debugging to record performance of the function and loaded the numbers into Excel to graph. I'm trying to attach a picture here:
netsshread

The output returned from the device is a sequence of ~3400 lines, each 175 bytes in length.

The Net/SSH2.php read function performs a while loop reading data from _get_channel_packet and adding the results to $this->interactiveBuffer.

The first thousand iterations through this loop take about 4 seconds. If the performance were linear I would expect a total run time of <15 seconds. What I am seeing are run times > 3 times that expected value, as with each iteration through the loop things get slower and slower.

Because every line of text returned from the device is ~ the same length (174-175 bytes per line), the amount of data moved in every iteration is very comprable. The only difference seems to be the amount of time taken to process the same data increases.

I have tested OTHER devices (Cisco IOS-XE routers, Nexus NXOS switches) and seen many magnitudes of performance better than these firewalls. The issue appears to be platform specific, and possibly related to how data is being sent from the ASA to the Net_SSH2 library.

Are there any settings or tweaks regarding the SSH connection or channel itself, or changes I could make to the library to wait for a larger buffer of input to return instead of 175 bytes at a time, or anything else you could think of to possibly alleviate my challenge?

I really love this library and use it extensively, thanks for your time and help.

terrafrost:
Fixes #670.

### DIFF
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -2670,7 +2670,7 @@ class Net_SSH2
         $match = $expect;
         while (true) {
             if ($mode == NET_SSH2_READ_REGEX) {
-                preg_match($expect, $this->interactiveBuffer, $matches);
+                preg_match($expect, substr($this->interactiveBuffer, -1024), $matches);
                 $match = isset($matches[0]) ? $matches[0] : '';
             }
             $pos = strlen($match) ? strpos($this->interactiveBuffer, $match) : false;


bantu:
#670 talks about a disclaimer. It also lists some concerns, e.g. "My fear in suggesting such a change is unintended consequences for other users with vastly different use cases." I assume those have been considered.

terrafrost:
It's hard to imagine a use case where people would need to do an expect() on > 1,024 characters. Per the benchmarks in #670 without a fix like this it takes 30 seconds for 500,000 of data without this commit vs. 1 second with it. So if people were using expect() on more than the last > 1,024 characters you'd expect there to be more bug reports about phpseclib / expect being slow but there aren't.

It is a bit of a guess tho. I figure with this if people are using it with more than the last 1,024 characters and they've just been tolerating slowness they can file bug reports after the feature is in all the branches.

PR #740, opened 2015-07-16
X509: use a random serial number for 1.0 branch

terrafrost:
If you re-issue a cert for a website if the serial number isn't different, as well, some browsers may reject the cert. This commit eliminates the need to keep track of serial numbers. It also provides an extra layer of protection to X509 certs per http://crypto.stackexchange.com/q/257/4520

### DIFF
--- a/phpseclib/File/X509.php
+++ b/phpseclib/File/X509.php
@@ -317,6 +317,10 @@ class File_X509
             include_once 'Math/BigInteger.php';
         }
 
+        if (!function_exists('crypt_random_string')) {
+            include_once 'Crypt/Random.php';
+        }
+
         // Explicitly Tagged Module, 1988 Syntax
         // http://tools.ietf.org/html/rfc5280#appendix-A.1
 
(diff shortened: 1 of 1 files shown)


phpseclib/File/X509.php
```diff
@@ -3277,7 +3277,17 @@ function sign($issuer, $subject, $signatureAlgorithm = 'sha1WithRSAEncryption')
...
+            if (!empty($this->serialNumber)) {
+                $serialNumber = $this->serialNumber;
+            } else {
+                if (!function_exists('crypt_random_string')) {
+                    include_once 'Crypt/Random.php';
+                }
+                // "The serial number MUST be a positive integer"
+                // "Conforming CAs MUST NOT use serialNumber values longer than 20 octets."
+                //  -- https://tools.ietf.org/html/rfc5280#section-4.1.2.2
+                $serialNumber = new Math_BigInteger(crypt_random_string(20) & ("\x7F" . str_repeat("\xFF", 19)), 256);
```
bantu: Might be a good idea to explain (in a comment) what this statement does, especially the & part.

terrafrost: Done. The following code better demonstrates what the comments try to explain:

```php
<?php
include('Math/BigInteger.php');

$a = new Math_BigInteger(str_repeat("\xFF", 20) & ("\x7F" . str_repeat("\xFF", 19)), 256);
$b = $a->toBytes(true);
echo bin2hex($b) . "\r\n";
echo strlen($a->toBytes(true)) . "\r\n";
```

bantu:
Thanks for the explanation. I think the application of the bitmap is a bit weird or at least complex. Couldn't the random method on BigInteger be used to do this in a less complex way? https://github.com/phpseclib/phpseclib/blob/master/phpseclib/Math/BigInteger.php#L3061

terrafrost:
It'd be less complex from the perspective of File/X509.php but it'd be more complex overall. Math_BigInteger::random() uses a while loop to loop until a random number within the appropriate range is found. The fact that min and max for random() don't have to be multiple's of two necessitate bitmasks not be used even if, computationally, they are less expensive. In this case, however, the max is a multiple of two, so some computational complexity can be avoided.

PR #739, opened 2015-07-15
Exception support

terrafrost:
Crypt/Base.php now uses exceptions.

To accomodate this decrypt() can no longer return false. If padding is enabled and the padding length is invalid (the only condition under which false was returned) an InvalidInputException will be thrown.

Also, previously, when encrypting, if the padding was disabled and the plaintext wasn't a multiple of the block size a E_USER_NOTICE would be issued via user_error() and then padding would be enabled. Now it's not enabled - an exception is thrown suggesting that you might want to enable padding and that's that.

I figure I'll add exceptions to more classes based on the feedback I get to this commit.

This commit should not be merged into either 1.0 or 2.0 - just master and maybe then a new 2.1 branch when the time is right.

I like exceptions over user_error's because they show the stack trace. With user_error's you just see the line the error was issued on but that's not that helpful since the relevant line for the user is the line where they called the method that triggered the error - not a line in a method that was called by the method that was called by the method that the user called.

Also, in some cases, I think catching errors can be less cumbersome than trying to use an error handler. I mean, if you just want to blanket suppress all errors then adjusting the error_reporting level is even easier than throwing the whole code into a giant try / catch block but idk...  mixing user_error's and try / catch seems...  tacky. Seems like it's better to use all one or all of the other - not both.

PR #749, opened 2015-07-18
Hash: rm internal and mhash support (5.3 has hash ext built-in)

terrafrost:
Per http://php.net/manual/en/hash.installation.php PHP 5.1.2 included hash by default and per http://php.net/manual/en/function.hash-algos.php md2 support was added in PHP 5.3. Since the 2.0 branch requires namespaces it has a PHP 5.3 dependency already so we can remove the other fallback options for Crypt/Hash.php. eg. mhash and the internal mode.

Sure, hash can be disabled with `--disable-hash` at compile time but then again, up until PHP 5.3, pcre could be disabled at compile time as well.

Also, this change will cause fatal errors for scripts calling Hash::MODE_INTERNAL or whatever but aside from the unit tests I'd be surprised if anyone was doing that anyway. But, none-the-less, if that's a concern, this can be put in master and not 2.0. It's not a big deal either way!


terrafrost:
We could also remove Crypt/Hash.php all together, I suppose, and replace all the calls to it with hash() and hash_hmac() as appropriate. One thing Crypt/Hash.php does is keep track of the hash length of the selected hashing algorithm but that could be incorporated into the classes that use it easily enough as well.

All in all, I don't necessarily see doing this as a big benefit. I think my time could probably be better spent doing other things but if others think it should be done it can be..

bantu:
Hummm. The Hash class isn't really a wrapper around hash() post-patch. For example `new Hash('whirlpool');` will silently do sha1, which is just weird.

Seeing how hash() does not support truncated hashes such as sha512-96 (at least hash_algos() does not return them in the list of supported hash functions) I think it makes sense to keep the Hash class. However, here are some improvements that are kind if necessary IMO:

1. If the specified hash is not supported, throw an \InvalidArgumentException. This can be checked using hash_algos() and keeping in mind the following point.

1. More generic support for truncated hashing, e.g. whirlpool-96 should return the first 96 bits of the full whirlpool hash. Maybe also check that the lengths are within bounds, e.g. throw an Exception for sha256-384

terrafrost:
I guess I could make it so that it just assumes that if `-(\d+)` is present at the end that it does `substr($hash, 0, $size / 8)`. Altho at that point if PHP itself added a hash like "sha-256" instead of just "sha256" there'd be problems.

It's also not quite as simple as just checking to see if the hash algorithm is in hash_algos() - the  hash length would still need to be hard-coded for each individual hash. The length is used in, for example, Crypt/RSA.php and Net/SSH2.php. So a switch statement is still necessary. I mean, I guess whenever you change the hash the length could be taken by actually creating a hash of an empty string and doing an strlen() on that but idk...  that seems a bit inelegant.

Throwing an exception of the hash algorithm is unsupported is fine.

bantu:
@terrafrost If there is an exact match for sha256-96 in hash_algos(), use that hash instead of applying our own truncation logic.

bantu:
Maybe keep the lenghts of well-known hash algorithms such as SHA and call strlen(hash('')) otherwise.

terrafrost:
Good ideas. I'll update the PR as time permits.

bantu:
Remaining question is whether it is a good idea to always truncate the right hand side of the output hash.

terrafrost:
I've made the suggested updates.

Thanks!

bantu:
Let's not silently fall back to sha1 but rather present some form of error.

bantu:
I think some of the existing hash tests should be combined into a new Hash Test testing the functionality that is left, e.g. truncated hashes, hash with key, hash without key, known lenght (e.g. sha256), unknown length (e.g. whirlpool).

bantu:
Looks like this misses 2.0?

terrafrost:
Yah. You had suggested adding test cases and I didn't have those in time for the release and I've been wanting to do a release for a while and didn't want to hold it up anymore.

Anyway, I had some other stuff come up today that took longer than expected and I need to be going to bed so I'll do the tagging and update the CHANGELOG.md and README.md file tomorrow.

terrafrost:
This is ready to be merged.

bantu:
This should be against master now.

bantu:
Don't fall back to sha1 but throw an exception instead?

bantu:
Also remove the sha1 default argument from the constructor?

bantu:
Add tests for getHash() and getLength()?

phpseclib/Crypt/Hash.php
```diff
@@ -213,79 +150,24 @@ function setHash($hash)
...
-        switch ($hash) {
-            case 'md2':
-                $mode = CRYPT_HASH_MODE == self::MODE_HASH && in_array('md2', hash_algos()) ?
-                    self::MODE_HASH : self::MODE_INTERNAL;
-                break;
-            case 'sha384':
-            case 'sha512':
-                $mode = CRYPT_HASH_MODE == self::MODE_MHASH ? self::MODE_INTERNAL : CRYPT_HASH_MODE;
                 break;
             default:
```
bantu: This block is not covered by tests. E.g. add a single test for whirlpool (no truncation) and one for tiger128,3-96 (truncated)?

bantu:
This should indeed almost be ready. Feel free to reopen this against master for correct bookkeeping, though.

terrafrost:
> Don't fall back to sha1 but throw an exception instead?

Well in that case I guess I'll finish up #739 before working on this. I figure we're either doing all user_error's or all exceptions. Not both.

bantu:
@terrafrost I don't think the order in which exceptions are introduced matter too much with regards to the unstable development branch `master`. In fact, the other PR looks good too.

@terrafrost sha1 should no longer be used (use sha2 or even sha3). Same for AES-CBC (in most cases, one should rather use AES-GCM or another AEAD cipher. As such I don't think those defaults are usuful in general. If you want a default, I'd rather change it to sha256.

PR #751, opened 2015-07-21
Work in Progress: PHP7 support for 2.0

bantu:


.travis.yml
```diff
@@ -8,6 +8,7 @@ php:
   - 5.4
   - 5.5
   - 5.6
+  - 7
```
GrahamCampbell: 7.0 would be better

bantu: Changed

bantu:
Sending this to master instead and allowing failure.

PR #754, opened 2015-07-24
[1.0] SSH2: Add encryption_algorithm_to_key_size method.

bantu:

PR #757, opened 2015-07-28
[2.0] Add _encryption_algorithm_to_crypt_instance method.

bantu:


terrafrost:
Looks good to me! Altho it does seem like this could target 1.0 instead of 2.0? #754 targeted 1.0 for instance.

bantu:
It was intentional due to object handling (copying) in PHP4.

terrafrost:
Ah okay. Well, looks good to me!

phpseclib/Net/SSH2.php
```diff
@@ -1894,6 +1802,50 @@ function _encryption_algorithm_to_key_size($algorithm)
...
+            case 'twofish128-ctr':
+            case 'twofish192-ctr':
+            case 'twofish256-ctr':
+                return new Twofish(Base::MODE_CTR);
+            case 'arcfour':
+            case 'arcfour128':
+            case 'arcfour256':
+                return new RC4();
+        }
+        return null;
```
GrahamCampbell: This shouldn't be needed, if a method has no return statement, PHP treats it as `null`.

bantu: True. Is there a sniff for that?

GrahamCampbell: Not sure, I don't use phpcs, sorry.

PR #758, opened 2015-07-28
SSH2: Remove all static variables as loadable classes may change at runtime

bantu:


bantu:
Har. Wrong target branch.

PR #759, opened 2015-07-28
[2.0] SSH2: Remove all static variables as loadable classes may change at runtime.

bantu:


terrafrost:
How would you unload a class at run time?

bantu:
@terrafrost It's more about adding classes by adding an additional autoloader. Also, the next step would be to make the supported algorithms changeable via method calls.

terrafrost:
Works for me. Go for it.

phpseclib/Net/SSH2.php
```diff
@@ -1167,90 +1167,87 @@ function _key_exchange($kexinit_payload_server)
...
-                );
-            }
-            if (class_exists('\phpseclib\Crypt\TripleDES') === false) {
-                $encryption_algorithms = array_diff(
-                    $encryption_algorithms,
-                    array('3des-ctr', '3des-cbc')
-                );
-            }
-            $encryption_algorithms = array_values($encryption_algorithms);
+        if (class_exists('\phpseclib\Crypt\RC4') === false) {
```
GrahamCampbell: no leading slash should be used - infact it's wrong to include it here

phpseclib/Net/SSH2.php
```diff
@@ -1167,90 +1167,87 @@ function _key_exchange($kexinit_payload_server)
...
+            $encryption_algorithms = array_diff(
+                $encryption_algorithms,
+                array('arcfour256', 'arcfour128', 'arcfour')
+            );
+        }
+        if (class_exists('\phpseclib\Crypt\Rijndael') === false) {
```
GrahamCampbell: same

phpseclib/Net/SSH2.php
```diff
@@ -1167,90 +1167,87 @@ function _key_exchange($kexinit_payload_server)
...
+            $encryption_algorithms = array_diff(
+                $encryption_algorithms,
+                array('aes128-ctr', 'aes192-ctr', 'aes256-ctr', 'aes128-cbc', 'aes192-cbc', 'aes256-cbc')
+            );
+        }
+        if (class_exists('\phpseclib\Crypt\Twofish') === false) {
```
GrahamCampbell: in all places

phpseclib/Net/SSH2.php
bantu: @GrahamCampbell I am going to merge this now. I would appreciate if you could file a new PR with information explaining why it's wrong.

PR #763, opened 2015-08-03
Add bootstrap.php checking environment (MB_OVERLOAD_STRING).

### LINKED ISSUE #762: Handling mbstring.func_overload
PHP has this really stupid feature called `mbstring.func_overload` which affects the behavior of `strlen()` and `substr()`.

Normally, these functions operate on a per-byte basis, but if you enable `mbstring.func_overload` it adopts multibyte character support. The consequences can be insane (e.g. string lengths being calculated based off a UTF-8 interpretation of a string, which can degrade security).

defuse/php-encryption solves this by using `mb_strlen($string, '8bit')` and `mb_substr($string, $start, $length, '8bit');` if these functions exist, and falls back to the defaults if they do not.

https://github.com/defuse/php-encryption/blob/d2c33279e31d5f3d4b74a82139b6ce56d551965b/src/Crypto.php#L654-L709

It might be worthwhile to adopt a similar approach in phpseclib.

bantu:
Fix #762

### DIFF
--- a/phpseclib/bootstrap.php
+++ b/phpseclib/bootstrap.php
@@ -1,0 +1,16 @@
+<?php
+/**
+ * Bootstrapping File for phpseclib
+ *
+ * @license http://www.opensource.org/licenses/mit-license.html MIT License
+ */
+
+if (extension_loaded('mbstring')) {
+    // 2 - MB_OVERLOAD_STRING
+    if (ini_get('mbstring.func_overload') & 2) {
+        throw new \UnexpectedValueException(
+            'Overloading of string functions using mbstring.func_overload ' .
+            'is not supported by phpseclib.'
+        );
+    }
+}
--- a/composer.json
+++ b/composer.json
@@ -61,6 +61,9 @@
     },
     "include-path": ["phpseclib/"],
     "autoload": {
+        "files": [
+            "phpseclib/bootstrap.php"
+        ],
         "psr-4": {
             "phpseclib\\": "phpseclib/"
         }
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "b24ab20be15b6312e532ee2ffa18d5fa",
+    "hash": "98a14889a81b2d6ecbb029483e954dec",
     "packages": [],
     "packages-dev": [
         {


terrafrost:
I guess the problem I have with this approach is that phpseclib doesn't have to be installed via composer.

bantu:
Let's just make everything else unsupported.

terrafrost:
I don't know that I completely support that.

Last month 2,447 came from sourceforge.net per http://sourceforge.net/projects/phpseclib/files/stats/timeline?dates=2011-07-29+to+2015-08-04

3.8 million came from packagist.org but that's all time and a lot of that is likely people who didn't intentionally download it.

A lot of PHP libraries can be downloaded and utilized without Composer. They may not be very good libraries but I think none-the-less a good number of PHP devs (altho perhaps not very good ones) do expect to be able to use stuff without Composer.

That said, I suppose it is easy enough to get this working without composer (for the 2.0.0 branch and later). ie.

```php
<?php
include('autoload.php');
include('bootstrap.php');

use phpseclib\Net\SSH2;

$ssh = new SSH2();
```

And I guess people can still download the 1.0 branch. Altho after X509 and SSH2 have had an API revamp I plan on redoing the website to promote whatever the latest version is (be it 2.1, 2.2 or whatever) I do figure I'll stop mentioning the 1.0 branch on the website.

Long story short, I guess I'm okay making other things unsupported.

Altho that said, wouldn't making everything else unsupported be a BC breaking change? ie. maybe this should go in master and not 2.0? And fwiw this proposed fix wouldn't fix it in the 1.0 branch, altho tbh, I don't think that's a huge issue anyway. Fact is SSH2.php (and probably the symmetric ciphers too) wouldn't even work with MB_OVERLOAD_STRING anyway.

bantu:
phpseclib does not support MB_OVERLOAD_STRING in any version, so this is clearly not a BC break.

First of all, we need to document this incompatibility, which I'd say is even sufficient to deal with this problem.

Then, loading the bootstrap file via composer is an optional precaution to make it easier for most of our users. This optional precaution could potentially also go into the 1.0 branch.

phpseclib/bootstrap.php
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Bootstrapping File for phpseclib
+ *
+ * @license http://www.opensource.org/licenses/mit-license.html MIT License
+ */
+
+if (extension_loaded('mbstring')) {
+    // 2 - MB_OVERLOAD_STRING
+    if (ini_get('mbstring.func_overload') & 2) {
+        throw new \UnexpectedValueException(
```
GrahamCampbell: don't need the slasg

phpseclib/bootstrap.php
```diff
@@ -0,0 +1,16 @@
+<?php
+/**
+ * Bootstrapping File for phpseclib
```
GrahamCampbell: missing fullstop

terrafrost:
> phpseclib does not support MB_OVERLOAD_STRING in any version, so this is clearly not a BC break.

No, but not supporting anything other than composer installs - as you proposed - is essentially a BC break.

I don't have a problem with it being a little bonus compatibility check for people using composer but to say that the only way phpseclib will be supported hence-forth is by composer installs...  that is essentially a BC break.

bantu:
@terrafrost I see.

phpseclib/bootstrap.php
paragonie-scott: Since this is not in a namespace, I agree, but there's no harm in being explicit.

Joey3000:
> A lot of PHP libraries can be downloaded and utilized without Composer. They may not be very good libraries but I think none-the-less a good number of PHP devs (altho perhaps not very good ones) do expect to be able to use stuff without Composer.

:+1: Phpseclib can be used e.g. in plugins for CMS/Apps potentially used on shared hosts. E.g., [for Joomla](http://extensions.joomla.org/extension/encrypt-configuration) and [for Piwik](https://github.com/Joey3000/piwik-LoginEncrypted).

phpseclib/bootstrap.php
GrahamCampbell: but that's not being explicit. class names are defined without a slash

PR #765, opened 2015-08-05
composer: Remove include path. No longer necessary due to PSR4 autoloading

bantu:


GrahamCampbell:
:+1:

bantu:
@GrahamCampbell Thanks.

PR #769, opened 2015-08-11
RSA: fix regression for PSS signatures

### LINKED ISSUE #768: RSA PSS default salt length handling changed since v1.0.0
These lines set `$sLen = NULL` if  `$this->sLen === NULL`.
https://github.com/phpseclib/phpseclib/blob/master/phpseclib/Crypt/RSA.php#L2573
https://github.com/phpseclib/phpseclib/blob/master/phpseclib/Crypt/RSA.php#L2611

Before https://github.com/phpseclib/phpseclib/commit/2fee2d4f98761b6543d467166ad5143780aa0830, it was `$this->sLen == false` and used `$this->hLen` as default
(I mean default as `$this->sLen === NULL`)

Is this change intended?
Then do we have to call `setSaltLength(false)` explicitly?

terrafrost:
Fix #768 

A unit test would be nice. Maybe I can do one today but maybe not.

### DIFF
--- a/phpseclib/Crypt/RSA.php
+++ b/phpseclib/Crypt/RSA.php
@@ -2637,7 +2637,7 @@ class Crypt_RSA
         // be output.
 
         $emLen = ($emBits + 1) >> 3; // ie. ceil($emBits / 8)
-        $sLen = $this->sLen === false ? $this->hLen : $this->sLen;
+        $sLen = $this->sLen ? $this->sLen : $this->hLen;
 
         $mHash = $this->hash->hash($m);
         if ($emLen < $this->hLen + $sLen + 2) {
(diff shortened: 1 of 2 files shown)


terrafrost:
Unit test has been added!

bantu:
You came up with this on your own? Becuase I was guessing in https://github.com/phpseclib/phpseclib/pull/668/files#r36621109

terrafrost:
Yah. You should be able to apply the unit test (without the corresponding code change) to 1.0 / 2.0 / master and see that it fails whereas it doesn't with the change.

Also, logically, it makes sense. Previously it was `$this->sLen == false`. `NULL == false` so `$this->hLen` was being used. But now it's `NULL === false`, which is false, so `$this->hLen` is /not/ being used. Since sLen isn't explicitly defined by default it's set to NULL so this change changed the default behavior of phpseclib.

PR #767, opened 2015-08-08
[2.0] Static Code Analysis with Php Inspections (EA Extended) (splitting #752)

kalessil:
@bantu : #753 + SCA specific for 2.0

Loops related changes I'd provide only for v2.0 to avoid conflicts merging 1.0 into 2.0.

Issue #337, opened 2014-05-06
Compatibility with PHP 5.3/5.4


splufer:
This is the output of  PHPCompatibility (https://github.com/wimg/PHPCompatibility) 

phpseclib0.3.6$ phpcs -p  --standard=PHPCompatibility .
FILE: phpseclib0.3.6/Crypt/Hash.php

FOUND 2 ERROR(S) AFFECTING 2 LINE(S)

323 | ERROR | Extension 'mhash' is removed since PHP 5.3 - use hash instead.
346 | ERROR | Extension 'mhash' is removed since PHP 5.3 - use hash instead.

FILE: phpseclib0.3.6/File/X509.php
FOUND 4 ERROR(S) AFFECTING 4 LINE(S)

 4232 | ERROR | Using a call-time pass-by-reference is prohibited since php 5.4
 4258 | ERROR | Using a call-time pass-by-reference is prohibited since php 5.4
 4325 | ERROR | Using a call-time pass-by-reference is prohibited since php 5.4
 4397 | ERROR | Using a call-time pass-by-reference is prohibited since php 5.4

PR #772, opened 2015-08-13
Tests/X509: add unit test showing need for return by reference

terrafrost:
The fact that return by references have been done has resulted in a number of tickets.

#337 
http://sourceforge.net/p/phpseclib/discussion/705284

Also, it was one of the things #767 attempted to change.

In light of this I think a unit test showing why return by references are done would be beneficial.


bantu:
@terrafrost This should still go in? Can you show me how to break this test?

terrafrost:
Here's what loadX509() currently returns:

http://pastebin.com/2v9gGBgn

Here's what loadX509() would return if `$extensions = &$this->_subArray($root, $path);` in `_mapInExtensions` had the `&` removed:

http://pastebin.com/3U0FSRUA

In particular look at the extnValue for the id-ce-cRLDistributionPoints extension. In the first URL it's an array. In the second it's a base64-encoded string.

PR #774, opened 2015-08-14
Fixed problem that command execution fails with timeout 

torinaki:
Fixed problem that command execution fails with timeout even if command return data.

### DIFF
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -3287,6 +3287,7 @@ class SSH2
                     }
 
                     if ($client_channel == $channel) {
+                        $this->curTimeout = $this->timeout;
                         return $data;
                     }
                     if (!isset($this->channel_buffers[$channel])) {
@@ -3308,6 +3309,7 @@ class SSH2
                         break;
                     }
                     if ($client_channel == $channel) {
+                        $this->curTimeout = $this->timeout;
                         return $data;
                     }
                     if (!isset($this->channel_buffers[$channel])) {


terrafrost:
Your commit breaks this:

```php
<?php
include('Net/SSH2.php');

$ssh = new Net_SSH2('domain.tld');
$ssh->login('user', 'pass');
$ssh->setTimeout(5);
echo $ssh->exec('ping google.com');
```

ie. without your commit that command times out after five seconds (as it should). With your commit it doesn't ever return.

What exactly were you trying to do that caused you problems?

torinaki:
So the case is following. I'm trying to run `composer install` on remote host that takes a while. `exec` method return only part of output and interrupts command execution on the middle of progress because of timeout and sets `is_timeout` to true. Reasonable question is why? Take into account that command always retuning some output.
After investigation I have found that any `exec` method execution should not be long that 10 seconds by default. I see few weird moments:
1. Why 10 seconds is enough for any command?
2. Why do we need limit max command execution time and how we ever know what is best max execution time for any command(for example for KDE compilation)? If someone need such functionality it can be achieved using callback. 
3. Timeout in most cases describes a time period during what some process do not response that helps us to understand that process hung.
4. I can be wrong, but we can't disable timeout at all. If we set `false` or `0` command execution will be interrupted right after start. 

Please point where I'm wrong and what/how I should use to solve my problem. Thanks in advance

terrafrost:
You're right - 10 seconds might not be enough time. You can change the timeout by calling `$ssh->setTimeout($x)` where $x is 30, 60, or whatever. You can also disable timeout's all together by doing `$ssh->setTimeout(0)`. Feel free to try it out.

Issue #775, opened 2015-08-15
Timeout cannot be disabled


torinaki:
I have faced with problem that behaviour of SSH2 client timeout is too weird and cannot be disabled.
My thoughts about timeout functionality is described in PR #774 where I was tried to fix incorrect(from my point of view) timeout behaviour. 

The current problem is that timeout cannot be disabled is you are setting `0` into `setTimeout`. I was tried to fix it and dive into to the deep of SSH2 class and timeout implementation and found it is more weird that I was thought before(see #774 [comment](https://github.com/phpseclib/phpseclib/pull/774#issuecomment-131275931))

I found the main cause of problem and it lives here:
    
    function _connect()
    {
        ...
        if (!is_resource($this->fsock)) {
            $start = microtime(true);
            $this->fsock = @fsockopen($this->host, $this->port, $errno, $errstr, $this->curTimeout);
            if (!$this->fsock) {
                $host = $this->host . ':' . $this->port;
                user_error(rtrim("Cannot connect to $host. Error $errno. $errstr"));
                return false;
            }
            $elapsed = microtime(true) - $start;
            $this->curTimeout-= $elapsed;
            if ($this->curTimeout <= 0) {
                $this->is_timeout = true;
                return false;
            }
        }
    
If we set timeout to `0` than `if ($this->curTimeout <= 0) {`will be true and timeout happen.
More interesting that as we know SSH2 timeout is using to limit command execution time not for determination is connection was hung or not, but in code above we can see that `$this->curTimeout` also passed as socket timeout configuration. This approach finally have blowed my mind and I gave up. 

Please help!

terrafrost:
It is not at all clear to me what you're wanting. Do you want to be able to make it so that the initial connection doesn't timeout? Unfortunately, I don't think that's possible.

Per your code snippet phpseclib uses `fsockopen()`. Even if you don't pass a timeout parameter to `fsockopen()` it still has one. Quoting http://php.net/fsockopen, the function definition for `fsockopen()` is as follows:

```
resource fsockopen ( string $hostname [, int $port = -1 [, int &$errno [, string &$errstr [, float $timeout = ini_get("default_socket_timeout") ]]]] )
```

ie. `$timeout` is equal to `ini_get("default_socket_timeout")` when it's not explicitely set.

Per http://php.net/manual/en/filesystem.configuration.php it is 60 by default and per http://php.net/manual/en/filesystem.configuration.php#ini.default-socket-timeout it's an integer.

[...]

PR #776, opened 2015-08-15
Timeout cannot be disabled(only refactoring)

torinaki:


torinaki:
Here currently only small refactoring of timeout functionality to investigate #775

terrafrost:
If you're wanting to test your changes against Travis CI you don't need to make a PR to do so - you can just create an account on travis-ci.org and then go to https://travis-ci.org/profile/torinaki , make sure that your fork of phpseclib is switched on. Then go to https://travis-ci.org/torinaki/phpseclib and check out the status.

torinaki:
@terrafrost yeah I know possibility to run tests on my own branch. I have prepared this PR just for refactoring purpose. As I said didn't changed anything in this PR, because just done small refactoring. Currently SSH2 class methods has very high complexity and it's too hard to read/understand code.

terrafrost:
My personal opinion: replacing one line of code with one function call is a little silly. Whereas before you could just trace through the code now, once you get to that function, you have to look that function up, and then go back to where you were at. And for what? Just to replace one line with another line?

I mean, if you're removing a lot of duplicated or re-usable code into a function that can be useful. But this doesn't really meet the "a lot of" test imho.

PR #777, opened 2015-08-17
Use placeholder variables in list() calls

duncan3dc:
PHP7 does not support empty parameters in calls to list()

http://php.net/manual/en/migration70.incompatible.php

### DIFF
--- a/phpseclib/Crypt/RSA.php
+++ b/phpseclib/Crypt/RSA.php
@@ -1182,7 +1182,7 @@ class RSA
                                 return false;
                             }
                             $this->_decodeLength($temp);
-                            list(, $iterationCount) = unpack('N', str_pad($temp, 4, chr(0), STR_PAD_LEFT));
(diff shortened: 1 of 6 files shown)


bantu:
I am afraid I'll have to fix the unit tests some more.

bantu:
Actual test failures introduced by your changes can be found here: https://travis-ci.org/phpseclib/phpseclib/jobs/75941904#L1461

bantu:
On that page you linked to, I see a "Empty list() assignments have been removed" section. But most of the list() calls you are changing in phpseclib are in fact not empty, they are only missing some variables. Are you sure this is an issue with PHP 7.0? https://travis-ci.org/phpseclib/phpseclib/jobs/76047515 does not seem to complain about this either.

duncan3dc:
Sorry for wasting your time, looks like I was misunderstanding an error and thought the cause was the missing `list()` parameter. You're right though, it seems to only affect missing parameters if there are __no__ parameters present

PR #782, opened 2015-08-20
Use Hoa\Math\Combinatorics\Combination\CartesianProduct

bantu:


bantu:
Argh. This needs PHP 5.4.

GrahamCampbell:
Maybe target at master and bump the version there to 5.4?

GrahamCampbell:
TBH, I think we really should bump 3.x to at least php 5.4.

composer.json
```diff
@@ -49,6 +49,7 @@
         "php": ">=5.3.3"
     },
     "require-dev": {
+        "hoa/math": "0.15.05.29",
```
GrahamCampbell: Shame we have to depend on a package with non-standard versioning conventions.

bantu: Absolute non-issue for a dev component I'd say. Also, there is maybe a de
facto standard at best.

GrahamCampbell: Oh yeh. I thought we were going to make it an optional dep to use all phpseclib features, but, yeh, it really is only used in the tests. ;)

bantu:
This patch is generally useful to have because CartesianProduct is an iterator which means we do not have to precompute all the data. On the other hand, however, this size of the dependency added is rather large for potentially little gain.

PR #783, opened 2015-08-25
Switch to extension_loaded instead of function_exists in most places

cweagans:
I'm working on a library called mcrypt-polyfill. The idea is to provide the same function names and signatures as the mcrypt extension, but backed by something more actively maintained than mcrypt (a combination of phpseclib, openssl, random_bytes() in PHP7 and possibly libsodium if it's present (and if phpseclib doesn't implement it first)).

In any case, function_exists() is not a good way to check that the mcrypt extension is loaded, because in the instance where I want to create a function with the same name, that changes behaviors in other code and that's not a desirable behavior. It also creates the possibility for infinite loops in some cases, which is also obviously not desirable.

To that end, I've switched phpseclib to use extension_loaded() throughout the code base, except for a few scenarios where function_exists really does make sense.

After applying this patch, these are the remaining instances of function_exists();

```
$ grep -r function_exists *
phpseclib/Crypt/Base.php:        if ($this->use_inline_crypt !== false && function_exists('create_function')) {
phpseclib/Crypt/Base.php:                    case !function_exists('hash_pbkdf2'):
phpseclib/Crypt/Base.php:                    case !function_exists('hash_algos'):
phpseclib/Crypt/Random.php:            // method 1. prior to PHP 5.3 this would call rand() on windows hence the function_exists('class_alias') call.
phpseclib/Crypt/Random.php:            if (extension_loaded('mcrypt') && function_exists('class_alias')) {
phpseclib/Math/BigInteger.php:        if (MATH_BIGINTEGER_MODE == self::MODE_GMP && function_exists('gmp_nextprime')) {
```

phpseclib/Crypt/Base.php:
* create_function makes sense
* The hash_* checks didn't make sense from grep, but in that context, I think it's a good idea to check it that way, because otherwise it'll be much more verbose, as I'd have to check for the extension being loaded AND a PHP version for the second one.

phpseclib/Crypt/Random.php:
* class_alias isn't part of a php ext, so this is the most straightforward check

phpseclib/Math/BigInteger.php:
* I went back and forth on this one. I don't have a strong opinion, so if you'd prefer `extension_loaded('gmp') && version_compare(PHP_VERSION, "5.2.0", "<=")`, I can change it. Probably more simple to do it this way though.

### DIFF
--- a/phpseclib/Crypt/RSA.php
+++ b/phpseclib/Crypt/RSA.php
@@ -473,11 +473,7 @@ class RSA
                 case defined('MATH_BIGINTEGER_OPENSSL_DISABLE'):
                     define('CRYPT_RSA_MODE', self::MODE_INTERNAL);
                     break;
-                // openssl_pkey_get_details - which is used in the only place Crypt/RSA.php uses OpenSSL - was introduced in PHP 5.2.0
-                case extension_loaded('openssl') && version_compare(PHP_VERSION, '5.2.0', "<="):
-                    define('CRYPT_RSA_MODE', self::MODE_INTERNAL);
-                    break;
-                case extension_loaded('openssl') && version_compare(PHP_VERSION, '4.2.0', '>=') && file_exists($this->configFile):
+                case extension_loaded('openssl') && file_exists($this->configFile):
                     // some versions of XAMPP have mismatched versions of OpenSSL which causes it not to work
                     ob_start();
                     @phpinfo();
--- a/phpseclib/Crypt/RSA.php
+++ b/phpseclib/Crypt/RSA.php
@@ -474,7 +474,7 @@ class RSA
                     define('CRYPT_RSA_MODE', self::MODE_INTERNAL);
                     break;
                 // openssl_pkey_get_details - which is used in the only place Crypt/RSA.php uses OpenSSL - was introduced in PHP 5.2.0
-                case !function_exists('openssl_pkey_get_details'):
+                case extension_loaded('openssl') && version_compare(PHP_VERSION, '5.2.0', "<="):
                     define('CRYPT_RSA_MODE', self::MODE_INTERNAL);
                     break;
                 case extension_loaded('openssl') && version_compare(PHP_VERSION, '4.2.0', '>=') && file_exists($this->configFile):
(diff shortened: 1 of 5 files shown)


GrahamCampbell:
This is much better security wise. Prevents bad bad dependencies from overriding these functions on systems that don't have the extension installed.

terrafrost:
Works for me! Altho that said I think this PR would be good for the 1.0 and 2.0 branches. And if those versions are going to be targeted then...  well I've made some in-line comments.

cweagans:
@terrafrost Okay, so the todo list here as I see it is this, if I'm not mistaken:

* For the master branch PR:
    * Remove the version_compare() call in `phpseclib/Crypt/RSA.php`
* Submit PR for 2.0:
    * Remove the version_compare() as noted above.
    * Basically identical to the master pr
* Submit PR for 1.0
    * @terrafrost is this a prerequisite for landing the PR in master & 2.0 ? I don't have much interest in 1.0, but it's a requirement to get this change into 2.x, I'll do it.

cweagans:
Also, I think the function_exists() call in phpseclib/Math/BigInteger.php can be replaced with `extension_loaded('gmp')` in master and 2.0 if we know for sure that the PHP version is >= 5.3.0

terrafrost:
> @terrafrost is this a prerequisite for landing the PR in master & 2.0 ? I don't have much interest in 1.0, but it's a requirement to get this change into 2.x, I'll do it.

I'll go ahead and say yes. If non-BC breaking improvements aren't going to be incorporated in the 1.0 branch then it needs to be abandoned and deprecated. Then people using 1.0 who have issues will have to rewrite their application to use 2.0 or whatever the latest version is. And how long do versions then stay valid for?

I don't have a problem abandoning 1.0 but, among other things, the website would need to be updated to be documenting the latest version instead of 1.0. And it's something I plan on doing (unless someone else wants to step up to the plate) but I want to redo File_X509 (making a lot of BC breaking changes to that so it'd go in the master branch only) and then later redo a lot of Net_SSH2.

Right now the only reason people have to use 2.0 instead of 1.0 is if they're wanting to avoid namespace conflicts. Or if they want to avoid E_DEPRECTED errors in sufficiently new versions of PHP. There just aren't compelling reasons to switch over, right now, and although the BC breaking changes I plan on introducing will hopefully change that those changes haven't been implemeted yet.

cweagans:
@terrafrost Personally, I wouldn't be too torn up about abandoning 1.0. The target for that branch is a version of PHP that's no longer supported. Not supporting the branch won't make all the existing code stop working, but hopefully it will push people to upgrade their setups. In any case, I've submitted #790 (PR for 2.0) and #791 (PR for 1.0) so that this is unblocked, as well as a couple of extra commits here that resolve some other issues noted above. Please let me know if you need anything else to get this in.

cweagans:
@terrafrost @bantu Any update on this, #790, and #791?

bantu:
This patch is identical to #790 which applies cleanly to master as well.

PR #789, opened 2015-08-30
X509: add getOID() method

### LINKED ISSUE #770: Add more OID's to X509 signature algorithm
The x509 parser is missing some OID for the mapping of the signature algorithm.

A list of OID's can be found in the following project:
https://golang.org/src/crypto/x509/x509.go (Line 232)
https://cryptography.io/en/latest/x509/reference/#cryptography.x509.oid.SignatureAlgorithmOID

An other problem is, that OID are replaced with the NOID mapping. If no NOID was found, the signature algorithm will contain the OID. This makes it very hard to identify the used algorithm.
It would be nice to have separation of the NOID and the OID.

terrafrost:
Fixes #770

### DIFF
--- a/phpseclib/File/X509.php
+++ b/phpseclib/File/X509.php
@@ -4614,4 +4614,27 @@ class File_X509
         $temp = preg_match('#^[a-zA-Z\d/+]*={0,2}$#', $temp) ? base64_decode($temp) : false;
         return $temp != false ? $temp : $str;
     }
+
+    /**
+     * Returns the OID corresponding to a name
+     *
+     * What's returned in the associative array is either a name or an OID if no OID to name mapping is available.
+     * The problem with this is that what may be an unmapped OID in one version of phpseclib may not be unmapped
+     * in the next version, so apps that are looking at this OID may not be able to work from version to version.
+     *
+     * This method will return the OID if a name is passed to it and if no mapping is avialable it'll assume that
+     * what's being passed to it already is an OID and return that instead. A few examples.
+     *
+     * getOID('2.16.840.1.101.3.4.2.1') == '2.16.840.1.101.3.4.2.1'
+     * getOID('id-sha256') == '2.16.840.1.101.3.4.2.1'
+     * getOID('zzz') == 'zzz'
+     *
+     * @access public
+     * @return String
+     */
+    function getOID($name)
+    {
+        $idx = array_search($name, $this->oids);
+        return $idx === false ? $name : $idx;
+    }
 }
--- a/phpseclib/File/X509.php
+++ b/phpseclib/File/X509.php
@@ -4618,9 +4618,10 @@ class File_X509
     /**
      * Returns the OID corresponding to a name
      *
-     * What's returned in the associative array is either a name or an OID if no OID to name mapping is available.
-     * The problem with this is that what may be an unmapped OID in one version of phpseclib may not be unmapped
-     * in the next version, so apps that are looking at this OID may not be able to work from version to version.
+     * What's returned in the associative array returned by loadX509() (or load*()) is either a name or an OID if
+     * no OID to name mapping is available. The problem with this is that what may be an unmapped OID in one version
+     * of phpseclib may not be unmapped in the next version, so apps that are looking at this OID may not be able
+     * to work from version to version.
      *
      * This method will return the OID if a name is passed to it and if no mapping is avialable it'll assume that
      * what's being passed to it already is an OID and return that instead. A few examples.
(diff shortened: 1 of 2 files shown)


phpseclib/File/X509.php
```diff
@@ -4614,4 +4614,27 @@ function _extractBER($str)
         $temp = preg_match('#^[a-zA-Z\d/+]*={0,2}$#', $temp) ? base64_decode($temp) : false;
         return $temp != false ? $temp : $str;
     }
+
+    /**
+     * Returns the OID corresponding to a name
+     *
+     * What's returned in the associative array is either a name or an OID if no OID to name mapping is available.
```
bantu: associative array? this method returns a string

phpseclib/File/X509.php
```diff
@@ -4614,4 +4614,27 @@ function _extractBER($str)
...
+     * getOID('2.16.840.1.101.3.4.2.1') == '2.16.840.1.101.3.4.2.1'
+     * getOID('id-sha256') == '2.16.840.1.101.3.4.2.1'
+     * getOID('zzz') == 'zzz'
+     *
+     * @access public
+     * @return String
+     */
+    function getOID($name)
+    {
+        $idx = array_search($name, $this->oids);
```
bantu: Linear search. Could setup an array_flipped version of $this->oids and use isset().

phpseclib/File/X509.php
```diff
@@ -4614,4 +4614,27 @@ function _extractBER($str)
...
+
+    /**
+     * Returns the OID corresponding to a name
+     *
+     * What's returned in the associative array is either a name or an OID if no OID to name mapping is available.
+     * The problem with this is that what may be an unmapped OID in one version of phpseclib may not be unmapped
+     * in the next version, so apps that are looking at this OID may not be able to work from version to version.
+     *
+     * This method will return the OID if a name is passed to it and if no mapping is avialable it'll assume that
+     * what's being passed to it already is an OID and return that instead. A few examples.
```
bantu: I am not sure this is useful API. Why not return null?

bantu: Maybe have two methods, getOID and mapOID.

terrafrost: > I am not sure this is useful API. Why not return null or false?
The current behaviour can be easily kept by calling
`$x509->getOID($name) ?: $name`

Because `switch ($x509->getOID($name))` looks better than `switch ($x509->getOID($name) ?: $name)`? I figure most people using `$x509->getOID($name)` will be using it in switch statements, trying to make their switch statements future proof and with that as a use case most people are going to have to do `$x509->getOID($name) ?: $name` anyway. And if most people are going to have to do that outside of the function then why not do that inside of the function instead?

It's also not clear to me what `mapOID` would do? I guess that'd do something like...  `getOID('2.16.840.1.101.3.4.2.1') == 'id-sha256'`? If so I can add that.

So I guess lmk about `mapOID`. I'll give you a chance to further convince me about `getOID`. In the mean time I'll do the other changes.

Thanks!

phpseclib/File/X509.php
terrafrost: Well what's returned by `loadX509()` is an associative array with either a name or OID.

bantu:
Your last commit seems to have added some log file. Please amend/rebase.

phpseclib/File/X509.php
bantu: I don't think the switch statements you proposed makes sense. Either you work on IDs, or you work on names. If you work on IDs, you don't have to translate to names. If you work on names, you don't handle IDs anyway, so returning an ID is pointless.

mapOID does what you implemented, getOID does what I described (return null/false if not found). Not sure this is consistent with existing methods though.

terrafrost: > If you work on IDs, you don't have to translate to names. If you work on names, you don't handle IDs anyway, so returning an ID is pointless.

I don't agree. Presumably you should always want to work on names over IDs as they're more human readable. But sometime's a name just isn't going to be available and in that case you don't have a choice - you have to work with an ID. But, with the way phpseclib currently works, although `$cert['tbsCertificate']['extensions'][0]['extnId']` may be an ID in v1.0.0 it might not be an ID in v1.0.1 - it might be a name.

Per your proposal I'd rename `getOID`, as implemented in this PR, to `mapOID` and then implement `getOID` such that it returns null if the string that's passed to it doesn't exist as  a name?

Personally, I'm thinking `getOID` is a better function name for the function as implemented in this PR. `getOID` doesn't imply that a mapping exists whereas `mapOID` does...

terrafrost:
> Your last commit seems to have added some log file. Please amend/rebase.

Done. Thanks for the heads up.

phpseclib/File/X509.php
```diff
@@ -4614,4 +4614,31 @@ function _extractBER($str)
...
+     *
+     * getOID('2.16.840.1.101.3.4.2.1') == '2.16.840.1.101.3.4.2.1'
+     * getOID('id-sha256') == '2.16.840.1.101.3.4.2.1'
+     * getOID('zzz') == 'zzz'
+     *
+     * @access public
+     * @return String
```
GrahamCampbell: wrong

PR #793, opened 2015-09-01
Use random_bytes() when appropriate.

cweagans:


phpseclib/Crypt/Random.php
```diff
@@ -53,6 +53,20 @@ class Random
      */
     public static function string($length)
     {
+        if (version_compare(PHP_VERSION, '7.0.0', '>=')) {
+            try {
+                $string = random_bytes($length);
```
GrahamCampbell: don't need the variable asignment

phpseclib/Crypt/Random.php
```diff
@@ -53,6 +53,20 @@ class Random
...
+                return $string;
+            }
+            catch (\EngineException $e) {
```
GrahamCampbell: there is no such class as EngineException - it was renamed to "Error"

GrahamCampbell: see https://3v4l.org/jfh8F

phpseclib/Crypt/Random.php
```diff
@@ -53,6 +53,19 @@ class Random
      */
     public static function string($length)
     {
+        if (version_compare(PHP_VERSION, '7.0.0', '>=')) {
+            try {
+                return random_bytes($length);
+            }
```
bantu: The catch statement need to go here I think

cweagans: Fixed.

bantu:
This is no longer required after #794

cweagans:
#794 is against a different branch.

bantu:
@cweagans We merge 1.0 into 2.0 and 2.0 into master.

cweagans:
Ah, I see. :+1:

PR #797, opened 2015-09-01
[1.0] Test on php 5.5.9 too

GrahamCampbell:
This is the version of php that ships with ubuntu 14.04, so it's probably important to be 100% sure that phpseclib works fine on that version of php.


bantu:
While this might be the version that the PHP binary on Ubuntu 14.04 reports, it is not exactly the same binary since the Ubuntu version is probably (heavily) patched while the version provided by Travis is probably directly from php.net. Nevertheless, I am fine with having this.

GrahamCampbell:
Yeh, sure. :)

PR #798, opened 2015-09-01
[1.0] Test on php 7.0

GrahamCampbell:
Can we please test on php 7.0 on every version?


GrahamCampbell:
And, yeh, I know the tests will fail, but I'm pretty sure we can just skip a few for now and we'll be good (that is, we can open an issue and fix them later). I don't think we should use allow failures, because that'll just hide issues. Better to skip the tests for the stuff that we know about, rather than hide eveything.

bantu:
1.0 is PHP4 compatible, e.g. due to class name constructors. Those do not work on PHP7 as far as I remember. As such, 1.0 will never be PHP7 compatible. Hence, having this in 1.0 does not make sense.

GrahamCampbell:
Ok, right. I'll look at 2.0 and 3.0 then. :)

PR #801, opened 2015-09-01
[1.0] PHPDoc Fixes

GrahamCampbell:
Made possible by https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1373, where case sensitive is set to false.

### DIFF
--- a/phpseclib/Crypt/AES.php
+++ b/phpseclib/Crypt/AES.php
@@ -124,7 +124,7 @@ class Crypt_AES extends Crypt_Rijndael
      * The namespace used by the cipher for its constants.
      *
      * @see Crypt_Base::const_namespace
-     * @var String
+     * @var string
      * @access private
      */
     var $const_namespace = 'AES';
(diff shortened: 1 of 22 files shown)


bantu:
Hell yeah!

GrahamCampbell:
Can we get this merged asap to stop other changes changes with wrong code going in.

bantu:
@GrahamCampbell Yes

bantu:
grep says this needs slightly more work, but these PRs are okay to merge IMO.

bantu:
Specifially:
```
phpseclib/File/X509.php:     * @return String
phpseclib/Net/SSH2.php:     * @var mixed false or Array
phpseclib/Net/SSH1.php:     * @return string or Array
phpseclib/Net/SFTP.php:     * @return string or Array
phpseclib/Math/BigInteger.php:     * @param $input Array
phpseclib/Math/BigInteger.php:     * @param $x String
phpseclib/Math/BigInteger.php:     * @param $shift Integer
phpseclib/Math/BigInteger.php:     * @param $x String
phpseclib/Math/BigInteger.php:     * @param $shift Integer
```

GrahamCampbell:
> @return String

That's odd.

> @param $shift Integer

Yeh - that needs manually fixing because the order is wrong. You've sad shift is the type an integer is the variable

GrahamCampbell:
> @return string or Array

and that isn't valid syntax, so i couldn't fix it

GrahamCampbell:
that is, automatially

PR #804, opened 2015-09-02
added passthru option to read function

zolex:
This small feature adds the ability to output the response from the read method "in realtime" which is good for commands which take a while and send the output while running.


phpseclib/Net/SSH2.php
```diff
@@ -2664,7 +2664,7 @@ function _get_open_channel()
      * @throws \RuntimeException on connection error
      * @access public
      */
-    function read($expect = '', $mode = self::READ_SIMPLE)
+    function read($expect = '', $mode = self::READ_SIMPLE, $passthru = false)
```
GrahamCampbell: phpdoc needed

phpseclib/Net/SSH2.php
```diff
@@ -2694,6 +2694,12 @@ function read($expect = '', $mode = self::READ_SIMPLE)
                 $this->in_request_pty_exec = false;
                 return $response ? $this->_string_shift($this->interactiveBuffer, strlen($this->interactiveBuffer)) : false;
             }
+            
+            if ($passthru === true) {
+
```
GrahamCampbell: cs

phpseclib/Net/SSH2.php
zolex: added

phpseclib/Net/SSH2.php
zolex: hm?

GrahamCampbell: no random extra line

GrahamCampbell:
Don't change the file permissions.

phpseclib/Net/SSH2.php
zolex: removed

zolex:
file permissions reverted

terrafrost:
Why not just do `if ($passthru)`? Doing `if ($passthru === true)` will make it so $passthru can't be 1. That's not a huge issue, I suppose, but is there an overwhelming reason to require that it be bool(true)?

Like `sha1`'s second parameter is a boolean but `sha1('', true)` and `sha1('', 1)` output the same thing.  With `sha1` the second parameter will be treated as `true` so long as it's castable to `true`. But with this PR we're not going to be doing that?

Also, this should be against the 1.0 branch. There's no reason this feature can't exist in that branch. It's not a BC break or anything.

PR #805, opened 2015-09-02
Backport to 2.0: Add PHP7 to Travis CI.

bantu:
Cherry pick of 2813e2cd3b829b3fd3e89bb66762f7b38e234ce6.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ php:
   - 5.5.9
   - 5.5
   - 5.6
+  - 7.0
   - hhvm
 
 env:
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
:-1: for allow failures

GrahamCampbell:
Skipping the broken tests would be better, because right now, we have the equivalent of skipping every test.

bantu:
@GrahamCampbell Don't care. This is for consistency with master. Feel free to actually fix the failures.

GrahamCampbell:
> This is for consistency with master. Feel free to actually fix the failures.

Yeh, I was actually going to send a PR - that's why i sent to 1.0 - i was planning to send to 1.0, deal with the issues, then send equivalent PRs to 2.0/master. Then you pointed out that 1.0 was totally broken. I've not had time to do anything more yet,

bantu:
@GrahamCampbell Someone will have to take the time to look at what tests are failing, what tests can be made to work on php 7.0, what tests need to be skipped. Simply allowing tests to fail allows us to see what tests are currently failing on PHP 7 (which may even change based on PHP 7 development).

@GrahamCampbell I am not sure 2.0 can be made 7.0 compatible. It might be possible. I am pretty sure 1.0 can't. Initially it looked like 2.0 can not be made compatible (see #777), which is why this ended up in master only. Anyway, thanks for your help.

Issue #760, opened 2015-07-31
remove SSH1.php from master branch


terrafrost:
It would be kept in the 1.0 and 2.0 branches but the overall impression I get is that people don't really use SSH1.php so it's just an extra file that needs to be maintained for no real reason what-so-ever.

If people do want it it can be re-added. Removing is defintely a BC breaking change - adding it back...  not so much.

bantu:
Fine with me.

metaclassing:
While I agree SSHv1 needs to disappear, I am one of the few unlucky users that is stuck with 1-2% of legacy network devices (embedded operating systems) that are v1 only with no upgrade path (thankyou vendors)

Most of my code is using the master branch for obvious performance improvements and features/bugfixes. I am willing to manually include the older file pulled from the 2.0 branch but wonder what kind of structural changes will be required on my end to keep it working? OO/namespace conversion etc.

terrafrost:
I can keep it too lol. SSH1.php is already namespaced - the only thing I'd need to add it to it to bring it in line with where master will be exceptions.  But then again maybe in your case using phpseclib v1 vs v2.1 or something would be good too? I intend to main to maintain v1 for a long time and adding non-BC breaking features to it.

I want to totally redo the X509 API but that'll be a huge BC break so that'll be master branch only. But for something like adding SFTP v4/5/6 support...  that can be in v1 and v2 and master.

metaclassing:
If you merge exception implementation for SSHv2 I can try to duplicate them in v1 for you and send a PR for review? My sucky 3rd world equipment shouldn't mean more wasted time and work for others lol.

terrafrost:
> If you merge exception implementation for SSHv2 I can try to duplicate them in v1 for you and send a PR for review? My sucky 3rd world equipment shouldn't mean more wasted time and work for others lol.

Works for me!

terrafrost:
@metalclassing - if you want to go ahead and start adding exceptions to SSHv1 that'd be great. SSHv2 is pretty much done. See #739 for more info.

Thanks!

metaclassing:
@terrafrost Hi, sorry for the slow response - its been a busy week. I will take a stab at adding exceptions and have something for you to review after the holiday. Have a good weekend!

metaclassing:
FYI, I hope to have the last couple dozen SSHv1 devices removed from my infrastructure by Q4 sometime... After that hopefully any vested interest I have for that 20 year old protocol will vanish lol

PR #806, opened 2015-09-04
Exception support for SSHv1

metaclassing:
I took a stab at replicating the SSHv2 exception support from #739 into SSHv1. As they are somewhat different, I targeted the user_error() calls for replacement, however some additional return false; statements may warrant a runtime exception?

Just as with the other exception support, the target is master and not 1.0 or 2.0. Please let me know the changes you would like me to make prior to accepting this PR.

### DIFF
--- a/phpseclib/Net/SSH1.php
+++ b/phpseclib/Net/SSH1.php
@@ -537,14 +537,15 @@ class SSH1
      * Connect to an SSHv1 server
      *
      * @return bool
+     * @throws \UnexpectedValueException on receipt of unexpected packets
+     * @throws \RuntimeException on other errors
      * @access private
      */
     function _connect()
     {
         $this->fsock = @fsockopen($this->host, $this->port, $errno, $errstr, $this->connectionTimeout);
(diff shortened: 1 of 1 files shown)


terrafrost:
Looks good to me!

GrahamCampbell:
Wait, but wasn't this class getting removed from the master branch anyway, so why upgrade it only to delete it?

GrahamCampbell:
Or is that not happening now? ;)

terrafrost:
metalclassing volunteered to at least do Exceptions to keep it in master:

https://github.com/phpseclib/phpseclib/issues/760#issuecomment-126701616

Just because I don't want to maintain it doesn't mean that I'm opposed to other people doing it if they're up to the task lol.

Plus, speaking practically, I want people to contribute and if adding Exceptions to SSH1.php can segway metalclassing into making other contributions (altho he's made other contributions already lol) then it's inclusion is valuable imho.

GrahamCampbell:
Sure. :)

PR #809, opened 2015-09-05
Update requirements to PHP 5.5.0

### LINKED ISSUE #792: Drop PHP 5.3
I think it would be a good idea to drop php 5.3 support for 3.x given that it's been EOL for quite a while. Maybe you should consider dropping php 5.4 too since that'll be EOL before you ship 3.x.

5.4 might be a long shot, but ditching 5.3 would be a good idea imo.

bantu:
Fix #792


GrahamCampbell:
We can remove all the commented out "random" stuff too can't we?

terrafrost:
I think this is a premature PR. Quoting my response in #792...

>> If the release is in Q3 2016, I'd say it would be reasonable to drop both php 5.3 and 5.4 in 3.0.

> I think we should cross that bridge when we get there.

> Q3 2016 is my best guess but what if it gets done Q4 2015 or Q1 2016? Let's not commit to anything until we need to. For the time being I think any such discussion is premature.

terrafrost:
> We can remove all the commented out "random" stuff too can't we?

I'm not sure what you mean?

Random.php has comments but I believe they're useful and don't believe they should be deleted? But then I'm not even sure that's what you're talking about?

GrahamCampbell:
All of this stuff: https://i.starbs.net/gYKY.

GrahamCampbell:
That's what I was referring to, for reference, but on a proper read, I can see it doesn't need removing. Ignore me. :)

GrahamCampbell:
> Q3 2016 is my best guess but what if it gets done Q4 2015 or Q1 2016? Let's not commit to anything until we need to. For the time being I think any such discussion is premature.

PHP 5.4 is still EOL though. It's EOL this weekend.

terrafrost:
> PHP 5.4 is still EOL though. It's EOL this weekend.

A lot of people are still using 5.3 per the link cweagans posted in #792:

https://wordpress.org/about/stats/

And quoting what I said in #792...

> I mean, I can only assume that phpseclib has achieved the level of popularity that it has due to the fact that it's server requirements are pretty minimal. I'm not a big fan of upending that approach when I believe it's that approach that has gotten phpseclib where it is today.

There's actually a Wordpress plugin that uses phpseclib. If only 15.2% of servers running Wordpress are running it on PHP 5.5 or later then a plugin that only works on PHP 5.5 or later isn't going to be a plugin most people are going to be able to use.

If in a year 99% of people are using PHP 5.6 then that's different but we're not there yet.

GrahamCampbell:
I don't really agree that people using wordpress represent the high end php community. Anyone writing any serious PHP app will be using a maintained version of php. I do understand what you are saying though. Maybe we should instead just set the minimum version to 5.4 for now, then look at the situation again in a year's time? It may be possible that some vulnerabilities in important parts of php 5.4 have been uncovered and not been fixed this time last year, which could alter the situation.

terrafrost:
I guess I'm okay with doing it for the time being. And so long as we don't actually write code that requires anything greater than 5.3. ie. no `[]`-style arrays, no traits (not that I have any idea how those would actually be beneficial for phpseclib), etc. We can revisit when it's time to release.

GrahamCampbell:
k ;)

paragonie-scott:
5.5 is reaching EOL soon; would it be wise to change direction to 5.6? :)

terrafrost:
I've been thinking about upping the master branch to 5.6 actually. There are some specific features of it that I want to use the X509 stuff. But I figure I'll do that when I make those changes lol.

When I do make those changes, tho, I'll probably replace all instances of `array()` with `[]`. Again, master branch only however!

bantu:
I believe this is no longer relevant.

PR #810, opened 2015-09-06
PHPDoc Fixes And Others

GrahamCampbell:


phpseclib/File/X509.php
```diff
@@ -4631,7 +4631,7 @@ function _extractBER($str)
      * getOID('zzz') == 'zzz'
      *
      * @access public
-     * @return String
+     * @return string
```
GrahamCampbell: this wasn't caught by my original PR because it was merged into the codebase after I send the PR

phpseclib/File/X509.php
```diff
@@ -3794,7 +3794,7 @@ function &_subArray(&$root, $path, $create = false)
      * @param string $path optional absolute path with / as component separator
      * @param bool $create optional
      * @access private
-     * @return array ref or false
```
GrahamCampbell: there is no such type as ref, so mixed is all we can use here

bantu: The return value may have a description though, right? So one option would be to describe in the text that the return value is a reference.

bantu: Why `mixed` instead of `array|false` like you have below?

bantu: How about using `&array` as the return type for reference to array? Not sure I've seen that used before.

phpseclib/File/X509.php
bantu: Yeah, I remember merging that.

phpseclib/File/X509.php
GrahamCampbell: Hmmmm. `&array` is invalid syntax as far as I'm aware. At least, I've never seen it used before, and I know of phpdoc parsers that would fail to read that correctly.

> Why mixed instead of array|false like you have below?

My bad there. I'll change that now.

GrahamCampbell:
Regarding documenting `&`, there's no need. The only point of phpdoc is to document stuff that can't be natively expressed by PHP. That's why the `access` annotation is pointless for example, because once you're using `public`, or `private`, or whatever, that' self documenting, of you see what I mean?

What I'm getting at, is the function definition in itself already documents the fact that it returns a pointer, so no need to document that again. ;)

phpseclib/Math/BigInteger.php
```diff
@@ -1855,7 +1854,8 @@ function _slidingWindow($e, $n, $mode)
                     }
                 }
 
-                for ($k = 0; $k <= $j; ++$k) {// eg. the length of substr($e_bits, $i, $j+1)
+                for ($k = 0; $k <= $j; ++$k) {
+                    // eg. the length of substr($e_bits, $i, $j+1)
```
bantu: This probably should be before the for loop.

GrahamCampbell: sure, i'll move it, i'll also add the missing spaces around the `+`

phpseclib/Crypt/Rijndael.php
```diff
@@ -755,7 +755,7 @@ function _subWord($word)
     {
         static $sbox;
         if (empty($sbox)) {
-            list(,,,, $sbox) = $this->_getTables();
+            list(, , , , $sbox) = $this->_getTables();
```
GrahamCampbell: this syntax is invalid on php 7 btw

GrahamCampbell: you need to provide all args as of 7.0

GrahamCampbell: It's listed in their upgrading guide btw: http://php.net/manual/en/migration70.incompatible.php, under "Empty list() assignments have been removed".

bantu: This (or something similar) was previously claimed at #777 and apparently turned out to be wrong.

GrahamCampbell: Ahhh, hmmmm, this issue came up in Laravel the other day, which is why I suggested it. I'll let Taylor know.

bantu: In particular, I'd argue that this call is not empty because it contains the `$sbox` variable. Not 100% sure. Maybe the manual/documentation should be a bit more clear.

bantu: Again: I might be wrong.

GrahamCampbell: Yeh, I think you're right. It's not "empty", and yeh, php's docs are a little ambiguous here.

GrahamCampbell:
Is this good to go then?

terrafrost:
Looks good to me..

GrahamCampbell:
:ship: :P

PR #811, opened 2015-09-06
Hash: rm internal and mhash support (5.3 has hash ext built-in)

terrafrost:
Replaces #749.

To address some of the comments in that ticket...

> This block is not covered by tests. E.g. add a single test for whirlpool (no truncation) and one for tiger128,3-96 (truncated)?

The way the tests work now is that if you're testing one algorithm you're testing it with and without the `-96` truncation. Should I create a whole new method - just for whirlpool - that doesn't do the truncation that I'm doing for every other hash algorithm?

Also, for HMAC's, I was unable to find official test vectors for whirlpool:

https://www.google.com/search?q=whirlpool+hmac+test+vector

In light of that I just made up my own with hash_hmac. I feel this merits mentioning because all the other hmac tests do use official algorithms.

I had a test for both whirlpool and tiger128,3, but tiger128,3's implementation is buggy in PHP 5.3, per this:

http://php.net/ChangeLog-5.php

In particular, PHP 5.4.0 has the following entry:

- Fixed bug #60221 (Tiger hash output byte order).


phpseclib/Crypt/Hash.php
```diff
@@ -213,79 +152,22 @@ function setHash($hash)
...
-                    case 'sha1':
-                    default:
-                        $this->hash = 'sha1';
+                // if the hash algorithm doens't exist maybe it's a truncated hash. eg. md5-96 or some such
+                if (preg_match('#(-\d+)$#', $hash, $matches) && in_array($hash = substr($hash, 0, -strlen($matches[1])), hash_algos())) {
+                    $this->l = abs($matches[1]) >> 3;
+                    break;
                 }
-                return;
+                throw new UnsupportedAlgorithmException("$hash is not a supported algorithm");
```
bantu: Why not \UnexpectedValueException ?

bantu: Please also add a test case for this. See https://phpunit.de/manual/current/en/appendixes.annotations.html#appendixes.annotations.expectedException

tests/Unit/Crypt/HashTest.php
```diff
@@ -0,0 +1,220 @@
...
+            sprintf(
+                "Failed asserting that '%s' HMACs to '%s' with key '%s'.",
+                $message,
+                $expected,
+                $key
+            )
+        );
+    }
+
+    public static function hashData()
```
bantu: Add sha1 and sha386 here to get to 100% code coverage.

tests/Unit/Crypt/HashTest.php
```diff
@@ -0,0 +1,220 @@
...
+     * @dataProvider hashData()
+     */
+    public function testHash96($hash, $message, $result)
+    {
+        $this->assertHashesTo($hash . '-96', $message, substr($result, 0, 24));
+    }
+
+    public function testGetHash()
+    {
+        $hash = new Hash();
```
bantu: If you're bored, please try make individual test methods and describe what you're testing. E.g.
```
public function testConstructorDefault()
{
    $hash = new Hash();
    $this->assertEquals($hash->getHash(), 'sha256');
}
```

bantu:
This looks pretty good. I'm nitpicking anyway. Nice catch on the tiger thing.

phpseclib/Crypt/Hash.php
terrafrost: > Why not \UnexpectedValueException ?

I didn't think of it lol. phpseclib has no way of knowing if "zzz" is a valid but unsupported algorithm or not so I was just kinda giving the input the benefit of the doubt.

PR #813, opened 2015-09-07
[master] More work on Hash.php

bantu:
Follows #811

### DIFF
--- a/phpseclib/Crypt/Hash.php
+++ b/phpseclib/Crypt/Hash.php
@@ -62,7 +62,7 @@ class Hash
      * @var int
      * @access private
      */
-    var $l = false;
+    var $length = false;
 
     /**
      * Hash Algorithm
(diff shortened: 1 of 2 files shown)


phpseclib/Crypt/Hash.php
```diff
@@ -173,27 +170,27 @@ function setHash($hash)
...
-        return strlen($output) > $this->l ? substr($output, 0, $this->l) : $output;
+        return strlen($output) > $this->length
+            ? substr($output, 0, $this->length)
+            : $output;
     }
 
     /**
      * Returns the hash length (in bytes)
      *
-     * @access public
```
GrahamCampbell: access was public anyway - i think we should remove these annotations on 2.x too

GrahamCampbell: that is, for all code

terrafrost:
Looks good to me with the exception of the public / protected thing...

bantu:
@terrafrost Removed the commit in question.

terrafrost:
Looks good to me!

PR #815, opened 2015-09-13
make select methods static

### LINKED ISSUE #556: make Crypt_RSA::createKey a static method
loadKey can't be static but it seems like createKey ought to be for the php5 / 2.0 branch.

terrafrost:
Fixes #556

### DIFF
--- a/phpseclib/Crypt/RSA.php
+++ b/phpseclib/Crypt/RSA.php
@@ -462,13 +462,15 @@ class RSA
     /**
      * Initialize static variables
      *
-     * @access public
+     * @access private
      */
-    static function init()
+    static function _initialize_static_variables()
     {
-        self::$zero= new BigInteger(0);
-        self::$one = new BigInteger(1);
-        self::$configFile = __DIR__ . '/../openssl.cnf';
+        if (!isset(self::$zero)) {
+            self::$zero= new BigInteger(0);
+            self::$one = new BigInteger(1);
+            self::$configFile = __DIR__ . '/../openssl.cnf';
+        }
     }
 
     /**
@@ -483,6 +485,8 @@ class RSA
      */
     function __construct()
     {
+        self::_initialize_static_variables();
+
         if (!defined('CRYPT_RSA_MODE')) {
             switch (true) {
                 // Math/BigInteger's openssl requirements are a little less stringent than Crypt/RSA's. in particular,
@@ -554,9 +558,10 @@ class RSA
      */
     static function createKey($bits = 1024, $timeout = false, $partial = array())
     {
+        self::_initialize_static_variables();
+
         if (!isset(self::$instance)) {
-            $class = __CLASS__;
-            self::$instance = new $class;
+            self::$instance = new static();
         }
 
         if (!defined('CRYPT_RSA_EXPONENT')) {
@@ -3049,5 +3054,3 @@ class RSA
         return $temp != false ? $temp : $str;
     }
 }
-
-RSA::init();
--- a/phpseclib/Crypt/RSA.php
+++ b/phpseclib/Crypt/RSA.php
@@ -2010,7 +2010,7 @@ class RSA
      * @param mixed $arg1
      * @param string $arg2 optional
      */
-    function setPassword($arg1 = false, $arg2 = false)
+    static function setPassword($arg1 = false, $arg2 = false)
     {
         if ($arg1 instanceof RSA) {
             $arg1->password = $arg2;
--- a/phpseclib/Crypt/RSA.php
+++ b/phpseclib/Crypt/RSA.php
@@ -470,6 +470,7 @@ class RSA
             self::$zero= new BigInteger(0);
             self::$one = new BigInteger(1);
             self::$configFile = __DIR__ . '/../openssl.cnf';
+            self::$instance = new static();
         }
     }
 
@@ -486,6 +487,12 @@ class RSA
     function __construct()
     {
         self::_initialize_static_variables();
+
+        if (isset(self::$instance)) {
+            $this->password = self::$instance->password;
+            $this->privateKeyFormat = self::$instance->privateKeyFormat;
+            $this->publicKeyFormat = self::$instance->publicKeyFormat;
+        }
 
         if (!defined('CRYPT_RSA_MODE')) {
             switch (true) {
@@ -559,10 +566,6 @@ class RSA
     static function createKey($bits = 1024, $timeout = false, $partial = array())
     {
         self::_initialize_static_variables();
-
-        if (!isset(self::$instance)) {
-            self::$instance = new static();
-        }
 
         if (!defined('CRYPT_RSA_EXPONENT')) {
             // http://en.wikipedia.org/wiki/65537_%28number%29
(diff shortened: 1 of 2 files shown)


phpseclib/Crypt/RSA.php
```diff
@@ -3030,3 +3049,5 @@ function _extractBER($str)
         return $temp != false ? $temp : $str;
     }
 }
+
+RSA::init();
```
bantu: Side effect. Can't have that.

phpseclib/Crypt/RSA.php
```diff
@@ -250,12 +250,20 @@ class RSA
     /**#@-*/
 
     /**
+     * Singleton instance
+     *
+     * @var \phpseclib\Crypt\RSA
+     * @access private
```
GrahamCampbell: though it's not, it's public?

terrafrost: Correct. Half the point is to hide it from phpdoc builds. See https://github.com/phpseclib/phpseclib/pull/813#commitcomment-13197976 .

GrahamCampbell: ah, right :)

terrafrost:
Here's the way this currently works (unit tests will need to be created):

```php
<?php
include('autoloader.php');

$loader->addPsr4('phpseclib\\', __DIR__ . '/phpseclib2.0');
$loader->register();

use \phpseclib\Crypt\RSA; 

RSA::setPrivateKeyFormat(RSA::PRIVATE_FORMAT_PUTTY);
RSA::setPassword('zzz');

// by default createKey() creates keys in the RSA::PRIVATE_FORMAT_PKCS1 format.
// the above makes createKey()'s private key be in the RSA::PRIVATE_FORMAT_PUTTY format.
// the above will also change the default private key format for all newly initialized
// RSA objects to RSA::PRIVATE_FORMAT_PKCS1. old RSA objects will not be changed.
extract(RSA::createKey());

$rsa = new RSA();
$rsa->loadKey($privatekey);
// the following will override the "global' defaults
RSA::setPrivateKeyFormat($rsa, RSA::PRIVATE_FORMAT_PKCS1);
RSA::setPassword($rsa); // reset password

echo $rsa;
```

Personally, I'd prefer it if the API could be more like this:

```php
<?php
include('autoloader.php');

$loader->addPsr4('phpseclib\\', __DIR__ . '/phpseclib2.0');
$loader->register();

use \phpseclib\Crypt\RSA; 

RSA::setPrivateKeyFormat(RSA::PRIVATE_FORMAT_PUTTY);
RSA::setPassword('zzz');

// by default createKey() creates keys in the RSA::PRIVATE_FORMAT_PKCS1 format.
// the above makes createKey()'s private key be in the RSA::PRIVATE_FORMAT_PUTTY format.
// the above will also change the default private key format for all newly initialized
// RSA objects to RSA::PRIVATE_FORMAT_PKCS1. old RSA objects will not be changed.
extract(RSA::createKey());

$rsa = new RSA();
$rsa->loadKey($privatekey);
// the following will override the "global' defaults
$rsa->setPrivateKeyFormat(RSA::PRIVATE_FORMAT_PKCS1);
$rsa->setPassword(); // reset password

echo $rsa;
```

The problem is...  if `setPassword` is `static` it won't have access to `$this` even if you call it by doing `$rsa->setPassword()`. And if `setPassword` isn't `static` you'll get an E_STRICT notice:

> Strict Standards: Non-static method phpseclib\Crypt\RSA::setPassword() should not be called statically in \path\to\test.php on line x

One alternative would be to just remove (for example) `setPassword` and handle it via `__call` and `__callStatic` but if that's done then calling non-existant functions won't result in fatal errors. Like right now if you do `$rsa->zzz()` you'll get the following error message:

> Fatal error: Call to undefined method phpseclib\Crypt\RSA::zzz() in /path/to/test.php on line x

That won't happen if `__call` or `__callStatic` exist.

I suppose `__call` / `__callStatic` could do `user_error` or throw an exception but idk...  at that point it's a distinctly different error that's being output. Plus, `user_error` won't show the line that `$rsa->zzz()` was called on but instead the line that the `user_error` is on.

So in light of all that I think the API utilized in the first code snippet is the best that we can do. Unless someone has any ideas?

bantu:
@terrafrost Without looking to deep into it: Your problem with static/non-static is probabably a result of the RSA class doing too many things for starters. I don't think your new API is very good. First of all, I very much dislike the extract() call. And then having the following static does not make sense to me:
```
RSA::setPrivateKeyFormat(RSA::PRIVATE_FORMAT_PUTTY);
RSA::setPassword('zzz');
```
How do I generate many keys of type A with password X and many keys of type B with password Y? That state really should not belong to the RSA class, it should belong to an instance. Consider introducing more classes, e.g. RSA_Key and RSA_KeyGenerator.

terrafrost:
> How do I generate many keys of type A with password X and many keys of type B with password Y? That state really should not belong to the RSA class, it should belong to an instance. Consider introducing more classes, e.g. RSA_Key and RSA_KeyGenerator.

You'd just call `setPassword` again. eg.

```php
<?php
RSA::setPrivateKeyFormat(RSA::PRIVATE_FORMAT_PUTTY);
RSA::setPassword('zzz');
for ($i = 0; $i < 10; $i++) {
    extract(RSA::createKey());
    file_put_contents("key$i.pem", $privatekey);
}

RSA::setPrivateKeyFormat(RSA::PRIVATE_FORMAT_PKCS1);
RSA::setPassword('yyy');
for ($i = 10; $i < 20; $i++) {
    extract(RSA::createKey());
    file_put_contents("key$i.pem", $privatekey);
}
```

terrafrost:
> First of all, I very much dislike the extract() call.

So right now `RSA::createKey()` (or `$rsa->createKey()` in every branch other than this PR lol) returns an array with three variables - `privatekey`, `publickey` and `partialkey`.

I could make it so that `privatekey` and `publickey` were RSA objects instead of strings. That'd eliminate the need for a static `setPassword` and `setPublicKeyFormat`. But `extract` would still be a valid thing to do. To rewrite the code in my last post to work in the manner that I just proposed...

```php
<?php
for ($i = 0; $i < 10; $i++) {
    extract(RSA::createKey());
    $privatekey->setPrivateKeyFormat(RSA::PRIVATE_FORMAT_PUTTY);
    $privatekey->setPassword('zzz');
    file_put_contents("key$i.pem", $privatekey);
}

for ($i = 10; $i < 20; $i++) {
    extract(RSA::createKey());
    $privatekey->setPassword('yyy');
    file_put_contents("key$i.pem", $privatekey);
}
```

The fact that RSA has a `__toString` method should result in the RSA object being converted to a string during the `file_put_contents` call.

Thinking about it I think that's a lot more preferable to what I implemented.

As for doing RSA_Key and RSA_KeyGenerator...  idk...  what methods would RSA_KeyGenerator have? createKey and that's it? Or would createKey not return anything and instead have  getPublicKey and getPrivateKey static methods as well? If all it had was createKey and we kept the current behavior of it returning an array with objects (instead of strings) then idk what the point of having a whole separate class would be?

And aside from RSA_KeyGenerator...  what would RSA_Key do? Would I just remove createKey and rename RSA to RSA_Key?

I guess it's hard to comment on your proposal without seeing it more fleshed out.

I think I'll create a new branch specific for RSA changes instead of trying to bundle all "making methods static" stuff into a single branch. So that branch will contain a new createKey and maybe input plugins as well (or maybe not - we'll see - createKey for sure it'll have). I'll also create a dedicated branch for BigInteger changes (which'll make random and randomPrime static and will do type hinting [which I had planned on initially doing in a separate "type hinting" PR]).

PR #820, opened 2015-09-14
[2.0] Fixed phpdoc

GrahamCampbell:


phpseclib/Crypt/Rijndael.php
```diff
@@ -190,29 +190,6 @@ class Rijndael extends Base
     var $kl;
 
     /**
-     * Default Constructor.
```
GrahamCampbell: this looks like it's left over from a bad branch merge.

bantu: ugh

GrahamCampbell: well, it's not documenting anything - it just randomly stops, and another docblock started

bantu: I'd be interested in knowing which one.

GrahamCampbell: Before my changes:

```php

    /**
     * Default Constructor.
     *
     * Determines whether or not the mcrypt extension should be used.
     *
     * $mode could be:
     *
     * - \phpseclib\Crypt\Base::MODE_ECB
     *
     * - \phpseclib\Crypt\Base::MODE_CBC
     *
     * - \phpseclib\Crypt\Base::MODE_CTR
     *
     * - \phpseclib\Crypt\Base::MODE_CFB
     *
     * - \phpseclib\Crypt\Base::MODE_OFB
     *
     * If not explictly set, \phpseclib\Crypt\Base::MODE_CBC will be used.
     *
     * @see \phpseclib\Crypt\Base::Crypt_Base()
     * @param int $mode
     * @access public

    /**
     * Sets the key.
     *
     * Keys can be of any length.  Rijndael, itself, requires the use of a key that's between 128-bits and 256-bits long and
     * whose length is a multiple of 32.  If the key is less than 256-bits and the key length isn't set, we round the length
     * up to the closest valid key length, padding $key with null bytes.  If the key is more than 256-bits, we trim the
     * excess bits.
     *
     * If the key is not explicitly set, it'll be assumed to be all null bytes.
     *
     * Note: 160/224-bit keys must explicitly set by setKeyLength(), otherwise they will be round/pad up to 192/256 bits.
     *
     * @see \phpseclib\Crypt\Base:setKey()
     * @see setKeyLength()
     * @access public
     * @param string $key
     */
    function setKey($key)
    {
        if (!$this->explicit_key_length) {
            $length = strlen($key);
            switch (true) {
                case $length <= 16:
                    $this->key_size = 16;
                    break;
                case $length <= 20:
                    $this->key_size = 20;
                    break;
                case $length <= 24:
                    $this->key_size = 24;
                    break;
                case $length <= 28:
                    $this->key_size = 28;
                    break;
                default:
                    $this->key_size = 32;
            }
        }
        parent::setKey($key);
    }
```

bantu: I think we may have a misunderstanding here. I was not trying to say you're wrong or anything similar. I was surprised to see this change actually made it into the repository. I agree it was likely caused by a merge. I'd be interested in knowing which merge. Specifically, whether I am to blame for this. :-P

GrahamCampbell: lol, ok ;)

GrahamCampbell: Hmmm, well, it's not broken at this point in the history: https://github.com/GrahamForks/phpseclib/blob/3f3e0cac98a479da04174c7805e4bcc390a27f0c/phpseclib/Crypt/Rijndael.php.

GrahamCampbell: Good here too: https://github.com/GrahamForks/phpseclib/blob/2013a31ecd70fc37c5413c4379b8a32922cbae26/phpseclib/Crypt/Rijndael.php.

GrahamCampbell: But it is broken here: https://github.com/GrahamForks/phpseclib/blob/d9e9504fbacc20502e4b456891b6f310e8022e91/phpseclib/Crypt/Rijndael.php.

bantu:
@GrahamCampbell Can we add stuff to CI ensuring all doc blocks have the expected format?

GrahamCampbell:
> Can we add stuff to CI ensuring all doc blocks have the expected format?

Yes. I'd recommend enabling StyleCI on this repo.

GrahamCampbell:
https://styleci.io/.

bantu:
We can probably do that. I generally prefer stuff that can be run offline, though.

GrahamCampbell:
You'd want to use StyleCI's PHP-CS-FIXER fork for that then. :)

PR #825, opened 2015-09-16
RSA: make createKey static and make it return RSA objects

terrafrost:
Replaces #815. I renamed loadKey to load to make RSA have an API more similar to the API I have planned for X509 (and CSR and SPKAC and all the classes I plan to split X509 up into).

Next step for RSA: splitting out all the key format handling stuff into their own individual class files.

### DIFF
--- a/phpseclib/Crypt/RSA.php
+++ b/phpseclib/Crypt/RSA.php
@@ -15,10 +15,10 @@
  *
  *    $plaintext = 'terrafrost';
  *
- *    $rsa->loadKey($privatekey);
+ *    $rsa->load($privatekey);
  *    $ciphertext = $rsa->encrypt($plaintext);
  *
- *    $rsa->loadKey($publickey);
+ *    $rsa->load($publickey);
  *    echo $rsa->decrypt($ciphertext);
  * ?>
  * </code>
@@ -33,10 +33,10 @@
  *
  *    $plaintext = 'terrafrost';
  *
- *    $rsa->loadKey($privatekey);
+ *    $rsa->load($privatekey);
  *    $signature = $rsa->sign($plaintext);
  *
- *    $rsa->loadKey($publickey);
+ *    $rsa->load($publickey);
  *    echo $rsa->verify($plaintext, $signature) ? 'verified' : 'unverified';
  * ?>
  * </code>
@@ -255,7 +255,7 @@ class RSA
      * @var array
      * @access private
      */
-    var $zero;
+    static $zero;
 
     /**
      * Precomputed One
@@ -263,7 +263,7 @@ class RSA
      * @var array
      * @access private
      */
-    var $one;
+    static $one;
 
     /**
      * Private Key Format
@@ -437,11 +437,12 @@ class RSA
      * OpenSSL configuration file name.
      *
      * Set to null to use system configuration file.
+     *
      * @see \phpseclib\Crypt\RSA::createKey()
      * @var mixed
-     * @Access public
+     * @access public
      */
-    var $configFile;
+    static $configFile;
 
     /**
      * Public key comment field.
@@ -450,6 +451,20 @@ class RSA
      * @access private
      */
     var $comment = 'phpseclib-generated-key';
+
+    /**
+     * Initialize static variables
+     *
+     * @access private
+     */
+    static function _initialize_static_variables()
+    {
+        if (!isset(self::$zero)) {
+            self::$zero= new BigInteger(0);
+            self::$one = new BigInteger(1);
+            self::$configFile = __DIR__ . '/../openssl.cnf';
+        }
+    }
 
     /**
      * The constructor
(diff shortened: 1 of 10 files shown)


terrafrost:
> Next step for RSA: splitting out all the key format handling stuff into their own individual class files.

The next commit will depend on both this PR and on #829 . I figure it'd be best if we went ahead and merged this to facilitate an easier to review PR for the next commit.

tests/Unit/Crypt/RSA/LoadKeyTest.php
```diff
@@ -7,7 +7,7 @@
 
 use phpseclib\Crypt\RSA;
 
-class Unit_Crypt_RSA_LoadKeyTest extends PhpseclibTestCase
+class Unit_Crypt_RSA_loadTest extends PhpseclibTestCase
```
bantu: This needs to be kept.

phpseclib/Crypt/RSA.php
```diff
@@ -250,12 +250,20 @@ class RSA
     /**#@-*/
 
     /**
+     * Singleton instance
```
bantu: Why?

tests/Unit/Crypt/RSA/CreateKeyTest.php
```diff
@@ -0,0 +1,37 @@
...
+ * @author    Jim Wigginton <terrafrost@php.net>
+ * @copyright 2013 Jim Wigginton
+ * @license   http://www.opensource.org/licenses/mit-license.html  MIT License
+ */
+
+use phpseclib\Crypt\RSA;
+
+class Unit_Crypt_RSA_CreateKeyTest extends PhpseclibTestCase
+{
+    static protected $privatekey;
```
bantu: not static

tests/Unit/Crypt/RSA/CreateKeyTest.php
```diff
@@ -0,0 +1,37 @@
...
+
+    public function testCreateKey()
+    {
+        extract(RSA::createKey(512));
+        $this->assertInstanceOf('\phpseclib\Crypt\RSA', $privatekey);
+        $this->assertInstanceOf('\phpseclib\Crypt\RSA', $publickey);
+        $this->assertNotEmpty("$privatekey");
+        $this->assertNotEmpty("$publickey");
+
+        self::$publickey = $publickey;
```
bantu: Return an array instead.

phpseclib/Crypt/RSA.php
terrafrost: It's a holdover from #815. Initially an RSA instance was being used to return a string but I made it return objects instead and I guess I forgot to remove this bit of the code.

tests/Unit/Crypt/RSA/CreateKeyTest.php
terrafrost: I'm not sure what you mean. Do you mean having `testEncryptDecrypt` use a dataprovider that generates the keys?

terrafrost:
I've made all  the suggested updates with the exception of the last one. I could use some clarification for that one:

https://github.com/phpseclib/phpseclib/pull/825#discussion_r40284084

tests/Unit/Crypt/RSA/CreateKeyTest.php
```diff
@@ -0,0 +1,37 @@
...
+        $this->assertNotEmpty("$publickey");
+
+        $this->publickey = $publickey;
+        $this->privatekey = $privatekey;
+    }
+
+    /**
+     * @depends testCreateKey
+     */
+    public function testEncryptDecrypt()
```
bantu: This method should better receive it's dependencies as an argument instead of having them passed via properties.

```public function testEncryptDecrypt($args)
{
     list($publickey, $privatekey) = $args;
```

The function above needs to return those arguments:
`return array($publickey, $privatekey);`

phpseclib/Crypt/RSA.php
```diff
@@ -452,6 +453,21 @@ class RSA
...
+     *
+     * @access private
+     */
+    static function _initialize_static_variables()
+    {
+        if (!isset(self::$zero)) {
+            self::$zero= new BigInteger(0);
+            self::$one = new BigInteger(1);
+            self::$configFile = __DIR__ . '/../openssl.cnf';
+            self::$instance = new static();
```
bantu: this needs to go

tests/Unit/Crypt/RSA/CreateKeyTest.php
```diff
@@ -0,0 +1,37 @@
+<?php
+/**
+ * @author    Jim Wigginton <terrafrost@php.net>
+ * @copyright 2013 Jim Wigginton
```
bantu: 2015?

terrafrost:
All the suggested changes have been implemented.

bantu:
This looks okay to me now. RSA could use some more code coverage though.

PR #826, opened 2015-09-18
Replace extension_loaded('mhash')

DavidAnderson684:
In PHP 7, mhash is not an extension. See: https://github.com/php/php-src/blob/php-7.0.0RC2/UPGRADING

### DIFF
--- a/phpseclib/Crypt/Hash.php
+++ b/phpseclib/Crypt/Hash.php
@@ -158,7 +158,7 @@ class Crypt_Hash
                 case extension_loaded('hash'):
                     define('CRYPT_HASH_MODE', CRYPT_HASH_MODE_HASH);
                     break;
-                case extension_loaded('mhash'):
+                case function_exists('mhash'):
                     define('CRYPT_HASH_MODE', CRYPT_HASH_MODE_MHASH);
                     break;
                 default:


phpseclib/Crypt/Hash.php
```diff
@@ -158,7 +158,7 @@ function Crypt_Hash($hash = 'sha1')
                 case extension_loaded('hash'):
                     define('CRYPT_HASH_MODE', CRYPT_HASH_MODE_HASH);
                     break;
-                case extension_loaded('mhash'):
+                case function_exists('mhash'):
```
GrahamCampbell: this is a security bug

GrahamCampbell: instead we should use extension_loaded and then check the php version

GrahamCampbell:
:-1: phpseclib 1.x isn't meant to work on php 7.

GrahamCampbell:
Only 2.x is meant to be php 7 compatible.

terrafrost:
1.x can work on PHP 7 - you'll just get strict warnings or deprecated warnings or some such.

That said, I don't see what the point of this PR is. If mhash isn't available then mhash won't be used. That's the way it currently works. Why would you have the mhash function defined if you didn't have the mhash extension available? Are you using some sort of BC layer? Well, even if you are, you'd be better off using hash - not mhash. PHP 5.3 introduced hash. So why would you want to be using mhash (or a compatibility layer) on PHP 7 anyway?

DavidAnderson684:
@terrafrost : mhash *is* still in PHP 7 (find it in ext/hash/hash.c in the source). It's just not an extension. Hence the advice given in the PHP 7 release notes to replace extension_exists('mhash') with function_exists('mhash'), as the former will return false, even though mhash is still there.

@GrahamCampbell : The PHP 7 release notes recommend replacing extension_exists('mhash') with function_exists('mhash'). Why do you say that doing so is a security issue?

terrafrost:
> @terrafrost : mhash is still in PHP 7 (find it in ext/hash/hash.c in the source). It's just not an extension. Hence the advice given in the PHP 7 release notes to replace extension_exists('mhash') with function_exists('mhash'), as the former will return false, even though mhash is still there.

In lieu of having looked at the sources, here's what http://php.net/manual/en/mhash.installation.php says:

> As of PHP 5.3.0, the mhash extension is emulated thru the Hash extension. This makes the mhash installation directory have no effect, and requires the hash extension enabled to make the mhash support available.

GrahamCampbell:
When I say 1.x doesn't work on php 7, I didn't mean this thing, I meant the whole code base. We don't even test 1.x on php 7.

bantu:
@terrafrost The 1.0 code base is not expected to work on PHP7 because of PHP4 style class constructors, which are (to my knowledge) no longer supported in PHP7.

DavidAnderson684:
@bantu: PHP4-style class constructors are deprecated in PHP 7.0, so throw a deprecation warning, but still work; see section 4: https://github.coma/php/php-src/blob/php-7.0.0RC3/UPGRADING

The meta-problem I'm facing is that WordPress's next release will still officially support PHP 5.2, so I'm restricted to the 1.0 code base. Unfortunately there are still plenty of (yes, horrible) hosting companies running 5.2. I am a developer for a WordPress plugin with 500,000 live installs, that uses phpseclib; we will need to keep the plugin operational on PHP 5.2 - 7.0. (When we start new plugins now, we're find with dropping 5.2 - but with large existing install bases, it's a different ball-game). So, I need to either a) review and patch the code of the 1.0 base for issues (time-consuming) on 7.0, or b) bundle both 1.0 and 2.0 and build some compatibility layer on top of the two (ugly).

GrahamCampbell:
Just because they support 5.2 it doesn't meant your plugin has to?

DavidAnderson684:
@GrahamCampbell : Depends what you mean by "has to". It wouldn't be fun to try telling that to 50,000 current PHP 5.2 users who will respond by 1-star-reviewing your product into oblivion, and tie up your support desk for weeks, when your next release stops their previously-working backups from working any more. (The wordpress.org plugin directory doesn't have a way for selectively pushing updates to some users and not to others). There are trade-offs on every side, and it is one of our selling points that our plugin supports what WordPress supports: that you don't need to be a tech-savvy user who can navigate other technical requirements in order to use it. As a business requirement we have to find a technical solution that keeps currently working features still working on PHP 5.2. That sucks, but we've evaluated it and it sucks less than the alternatives.

GrahamCampbell:
Ok, then. You need two versions of the plugin then.

DavidAnderson684:
It would be lovely if the workings of the wordpress.org plugins directory + update system made that an option.

Anyway, now that our options are clearer, we'll have a think.

bantu:
@GrahamCampbell If PHP4 style constructors are only deprecated we should make 1.0 PHP7 compatible too.

GrahamCampbell:
Sure. I've got no objections to that at all. Just when I wanted to test on php 7 on 1.x, the impression I got was that it wasn't worth it and that we should just focus on 2.x.

terrafrost:
1.0 is PHP7 compatible. I've tested it on PHP7. It just issues deprecated notices. #817 addressed the PHP7 compatibility issues 1.0 had.

Hopefully PHP 7 won't drop PHP4-style constructor support until after Wordpress drops PHP 5.2 support.

In the mean time, I still don't believe this PR is needed per my earlier comments.

DavidAnderson684:
This is great news.

"In the mean time, I still don't believe this PR is needed per my earlier comments." - I think you're right; the condition can't ever fire, given that extension_loaded('hash') is tested before it.

PR #827, opened 2015-09-19
Crypt: make it so the key length can be explicitly set and retrieved for all ciphers / 1.0

terrafrost:
Still need to add unit tests for the new functions..

PR #828, opened 2015-09-21
Crypt: make it so the key length can be explicitly set and retreived for all ciphers / 2.0

terrafrost:
#827 for the 2.0 / master branches


bantu:
Something went wrong in the merge as this reintroduces const_namespace.

PR #833, opened 2015-09-28
SFTP.php : prevent endless loop when upload completed with non-zero $start or $local_start for 1.0

terrafrost:
Replaces #830.

#830 is for the master branch but the issue is present in the 1.0 branch as well. I've also added a unit test that demonstrates the issue.

### DIFF
--- a/phpseclib/Net/SFTP.php
+++ b/phpseclib/Net/SFTP.php
@@ -1917,10 +1917,12 @@ class Net_SFTP extends Net_SSH2
 
             if ($local_start >= 0) {
                 fseek($fp, $local_start);
+                $size-= $local_start;
             } elseif ($mode & NET_SFTP_RESUME_START) {
                 // do nothing
             } else {
                 fseek($fp, $offset);
+                $size-= $offset;
             }
         } elseif ($dataCallback) {
             $size = 0;
--- a/phpseclib/Net/SFTP.php
+++ b/phpseclib/Net/SFTP.php
@@ -1918,11 +1918,6 @@ class Net_SFTP extends Net_SSH2
             if ($local_start >= 0) {
                 fseek($fp, $local_start);
                 $size-= $local_start;
-            } elseif ($mode & NET_SFTP_RESUME_START) {
-                // do nothing
-            } else {
-                fseek($fp, $offset);
-                $size-= $offset;
             }
         } elseif ($dataCallback) {
             $size = 0;
--- a/tests/Functional/Net/SFTPUserStoryTest.php
+++ b/tests/Functional/Net/SFTPUserStoryTest.php
@@ -649,5 +649,18 @@ class Functional_Net_SFTPUserStoryTest extends PhpseclibFunctionalTestCase
         $this->assertSame($stat['type'], NET_SFTP_TYPE_SYMLINK);
 
         $sftp->enableStatCache();
+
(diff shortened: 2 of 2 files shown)


tests/Functional/Net/SFTPUserStoryTest.php
```diff
@@ -649,5 +649,22 @@ public function testStatVsLstat($sftp)
...
         $sftp->enableStatCache();
+
+        return $sftp;
+    }
+
+    /**
+     * @depends testStatVsLstat
+     * @group github830
+     */
+    public function testEndlessLoopOnUpload($sftp)
```
bantu: This test case only covers the first hunk above. The second hunk is covered by testResourceXfer already, but I am not sure it is sufficient. Can you explain/confirm/add another test?

terrafrost: I can do that...

Although tbh I think the code ought to be updated such that the "second hunk" isn't even possible. Quoting from the code comments:

```php
     * $start and $local_start give you more fine grained control over this process and take precident over self::RESUME
     * when they're non-negative. ie. $start could let you write at the end of a file (like self::RESUME) or in the middle
     * of one. $local_start could let you start your reading from the end of a file (like self::RESUME_START) or in the
     * middle of one.
     *
     * Setting $local_start to > 0 or $mode | self::RESUME_START doesn't do anything unless $mode | self::SOURCE_LOCAL_FILE.
```

Based on that it doesn't seem like it should be doing `fseek($fp, $offset)`, since `$offset` is based on `$start`. ie. I think they ought to be completely disjoint.

Here's the commit that added it:

https://github.com/phpseclib/phpseclib/commit/cbe156f0cbc7e715731b841482a1b1625df31250

Truth by told, I don't really like how the `put` function has so many different parameters and it's something I'd like to change, as elaborated upon here:

https://github.com/phpseclib/phpseclib/pull/655#issuecomment-92600348

bantu: @terrafrost Are you saying that #830 was not acceptable after all? If yes, why are you including the second hunk?

terrafrost: The second hunk is already there and no one has created a bug report for it. It's possible people like the current behavior even tho it, imho, doesn't mesh with the documentation. It could be changed but idk...  atm it's not a priority of mine.

bantu: With "the second hunk" I mean the addition of `$size-= $offset;`, not the code that is already there in the else block.

terrafrost: SFTP.php is still doing the fseek(). Or at least was. But regardless, the latest commit removes that. If I'm going to have a unit test for that portion of SFTP.php then that portion of SFTP.php ought to be behaving as per the documentation.

phpseclib/Net/SFTP.php
```diff
@@ -1917,10 +1917,7 @@ function put($remote_file, $data, $mode = NET_SFTP_STRING, $start = -1, $local_s
 
             if ($local_start >= 0) {
                 fseek($fp, $local_start);
-            } elseif ($mode & NET_SFTP_RESUME_START) {
-                // do nothing
```
bantu: I hated that anyway. Good it's gone. :-P

PR #837, opened 2015-10-01
RSA: move key format handling to plugins

terrafrost:
For the master branch only. Depends on #828.

Instead of doing `$rsa->setPrivateKeyFormat(RSA::FORMAT_PKCS1)` or whatever you now do `$rsa->setPrivateKeyFormat('PKCS1')`. In this way it's kinda like [mb_convert_encoding](http://php.net/manual/en/function.mb-convert-encoding.php), which takes in strings as the to and from character formats instead of constants. Also, just like with `mb_convert_encoding`, key formats are case-insensitive.

The supported formats are:

- OpenSSH
- Raw
- PKCS1
- PKCS8
- XML
- PuTTY
- Microsoft BLOB - a new format not supported by phpseclib v1.0 or v2.0.

Adding a new plugin is simple. Plugins basically need to define between 1-3 functions:

- load
- savePublicKey
- savePrivateKey

You can look at the existing plugins to get an idea of how they should work - of the parameters they require and what they need to return. I guess the XML one would probably be the easiest to understand.

The plugins add some new support too. Like for PKCS1 you can now change the encryption algorithm that's being used. Previously it was DES-EDE3-CBC and that was that. Mainly because that's what puttygen produced. Now, it still defaults to DES-EDE3-CBC (because I don't know how widely supported other algorithms are) but you can change it by doing `PKCS1::setEncryptionAlgorithm('AES-128-CBC')` or whatever.

You can't currently change the encryption algorithm for PKCS8. Among other things I've never seen a PKCS8 encrypted key that used an algorithm other than PBE-MD5-DES. OpenSSL discusses the state of affairs for PKCS8 encryption algorithm support:

https://www.openssl.org/docs/manmaster/apps/pkcs8.html#PKCS-5-v1.5-and-PKCS-12-algorithms

I feel more comfortable changing the algorithm for PKCS1 because I have seen keys with different algorithms before for that.

Comments can be set in those plugins that support them by doing `OpenSSH::setComment('whatever')` or `PuTTY::setComment('whatever')`.

You can see what key formats are supported by doing `RSA::getSupportedFormats()`. You can see what format a particular key is in by doing `$rsa->getLoadedFormat()`. You can dynamically add a new format by doing `RSA::addFileFormat('...')`.


terrafrost:
I checked `mb_convert_encoding` and the case names for that are not case sensitive so I've updated RSA.php to be similarly non-case-sensitive.

I also added a new function:

```php
RSA::addFileFormat('myformat');
```

Any non-abstract file in Crypt/RSA/ will be loaded as a plugin but if you have a custom plugin that doesn't live in that directory you can now add your own.

bantu:
I am not a fan of those static calls and would (probably, as always) prefer a more object oriented way, say
```
$someRsaObject = new FooRSA(new ClassNameForFormat)
```
or
```
$someRsaObject = new FooRSA;
$someRsaObject->generateSomething(new ClassNameForFormat);
```
but I do not have time to further look into this.

This way, the whole plugin-loading would be done via class autoloading and would trivially allow extension with new formats/classes.

terrafrost:
It's hard to comment on an API that doesn't exist. An implementation of that API wouldn't even be necessary - just a example of how to use it or something.

That said, even if we did have an API it's possible I just might not like it as much as I like this one (and I do like the API that I've come up with). And even if you and I come to an agreement there are liable to still be other people out there who don't like the new API. Not everyone is going to see eye to eye and you can't make everyone happy - that's just life.

But that said, I do like this API. It's basically "you load an RSA key and you do stuff with it". This "stuff" can include encryption, signature verification, or outputting the key you loaded in whatever format you so desire.

The only thing that doesn't completely fit with that paradigm is `createKey` but the fact that it's static somewhat removes it from the regular API anyway. And I don't see a lot of point in creating a class just for key generation. So instead of doing `\phpseclib\Crypt\RSA::createKey` you do `\phpseclib\Crypt\RSA\KeyGen::createKey`? I just don't like it.

> This way, the whole plugin-loading would be done via class autoloading and would trivially allow extension with new formats/classes.

It was pretty trivial before until I made it class insensitive. At least if you were trying to echo out a key in a specified format. It'd basically do `if (class_exists(...))` which would trigger the auto loader.

And I don't see how you can use autoloading for `load()` method. I mean, sure, you could do so if you **required** the format be passed to it as a second parameter but I think that's silly. The fact that it'll essentially auto-detect the format - that it's able to guess it as it were - is one of RSA.php's big strengths. I mean, sure, we could auto-load, but then where do you get the list of file format classes to auto load? Do I hard-code it into RSA.php instead of generating it dynamically from the directory? Maybe the person using phpseclib has to specify each and every format they want to support in the RSA constructor and a class_exists is done on each format they name to autoload the class if it exists? Sorry, but I think the method I'm employing is better..

terrafrost:
Another advantage of the case-insensitive approach that's being used...  let's say straight up auto-loading were being used. At that point you could do something like `$rsa->getPrivateKey('\DOMDocument')`. phpseclib would then try to call `load()`, which would succeed, would try to read an XML file off of the local file system and then idk.

Another alternative design would have been to objects instead of static calls. eg. loading a `PKCS1` key results in `new PKCS1()` being called. With an approach like this you could do `getPrivateKey(PKCS1(array('password' => 'whatever')))` or some such. ie. instead of having global format settings you could have per key format settings. But idk I like this approach better..

terrafrost:
So one thought I had...  we could give the RSA object methods depending on the key that was loaded. Like if it's a PuTTY key you can do `$rsa->getComment()`.

The problem with that is that we'd have to do `__call`, which I'm not a huge fan of since doing so will change what'll happen when you call an undefined method.

Here's the normal error in PHP5:

> Fatal error: Call to undefined method a::zzz()

If you implement `__call` that no longer happens. Instead you get no error.

PHP7 changes this around somewhat. Instead you get this:

> Fatal error: Uncaught Error: Call to undefined method a::zzz() on line 11

That can be duplicated by doing `throw new Error('Call to undefined method ' . __CLASS__ . '::' . $name)` but the stacktrace and line numbers are different.

So I guess one option would be to not care about the stacktrace and line numbers and just throw an exception - even on PHP5. But idk.

And you can currently get comments, anyway, by doing `$key = PuTTY::load('...'); echo $key['comment']; $rsa->load($key)` (vs. `$rsa->load('...');`)

Of course if we're doing this for `get*()` should this also be done for `set*()`? Well one problem with `set*()` is that you won't necessarily know the private key format. eg. you could do `getPrivateKey('PuTTY')` or some such. Maybe that method could do `func_get_args()` and then call the plugin via `call_user_func_array()` but idk.

I suppose the API could be completely redesigned as well. It'd have been interesting to see the API bantu was considering. But idk...  I suppose every API will have it's advantages and disadvantages and altho a hypothetical API might be better than the one presented in this PR I do believe that this API is none-the-less an improvement from the current API and it has the advantage of actually having an implementation.

metaclassing:
While on the topic of loading cert(s) and encoding, have you given any more thought to desired behavior when parsing a blob of ascii that contains multiple certificates (and other assorted throw-away bits)? I originally thought just return the first matching key and chuck the rest, but could see a use-case in returning multiple RSA objects in an array for every cert extracted from the input. That would fit the RSA::load() paradigm more than the $RSA->load() approach, but i suppose there are trade-off's either way.

terrafrost:
> That would fit the RSA::load() paradigm more than the $RSA->load() approach, but i suppose there are trade-off's either way.

Yah that's true. And no - that's not a use case I had considered. It seems like that use-case is probably more common for X509 certs than it is for RSA keys anyway? Like with X509 certs you might have a certificate validation chain. Not really sure how to deal with that atm.

> Second thought: Maybe I should think about writing a File\PEM handler that returns multiple RSA and KEY objects after digging through the input?

That might not be a bad idea. One thing I'm thinking of doing, long term, is to make a File/PublicKey.php file that'll return RSA or DSA or ECDSA objects depending on the key. Like it'll auto detect the format and return the appropriate object. Reason being that sometimes it might not be obvious or you might not even know or something idk lol.

But that isn't super helpful for X509 certs either..

PR #838, opened 2015-10-05
phpDoc changes for 1.0 branch

terrafrost:
A lot of the in-line comments for #827 had to do with the phpDoc syntax so this makes all current phpDoc calls consistent.

It also adds missing @return statements to four functions in Crypt/RSA.php.

### DIFF
--- a/phpseclib/Crypt/Base.php
+++ b/phpseclib/Crypt/Base.php
@@ -296,9 +296,9 @@ class Crypt_Base
     /**
      * Does internal cipher state need to be (re)initialized?
      *
-     * @see setKey()
-     * @see setIV()
-     * @see disableContinuousBuffer()
+     * @see Crypt_Base::setKey()
+     * @see Crypt_Base::setIV()
+     * @see Crypt_Base::disableContinuousBuffer()
      * @var bool
      * @access private
      */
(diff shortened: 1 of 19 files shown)


terrafrost:
Let's get #827 merged before merging this.

phpseclib/Crypt/Base.php
```diff
@@ -296,9 +296,9 @@ class Crypt_Base
     /**
      * Does internal cipher state need to be (re)initialized?
      *
-     * @see setKey()
-     * @see setIV()
-     * @see disableContinuousBuffer()
+     * @see Crypt_Base::setKey()
```
GrahamCampbell: what about static::setKey()

GrahamCampbell: or self, for php 5.2-

terrafrost: static, to me, implies that the method needs to be called statically, which is not the case here.

I could do self but the 1.0 branch is supposed to be compatible with 4.x (with PHP_Compat). Of course I guess making changes to phpDoc comments doesn't really change the versioning requirements so I guess I can do it.

I guess phpDoc just kinda translates `self` to `Crypt_Base` (assuming that `self` occurs in `Crypt_Base`) when it's building the documentation?

GrahamCampbell: meh, I don't really mind - I was just throwing the idea out there incase you liked it :)

Issue #843, opened 2015-10-09
Clarifying random_bytes() behavior in PHP 7


paragonie-scott:
Re: https://github.com/phpseclib/phpseclib/blob/0ba1a04c05de04fba9d09ae5b7b33b2399541dca/phpseclib/Crypt/Random.php#L60

`random_bytes()` will throw:

* `Exception` in the event of a catastrophic CSPRNG failure
* `TypeError` if `zend_parse_parameters` fails
* `Error` if `$length` is less than or equal to 0

paragonie-scott:
Also

https://github.com/phpseclib/phpseclib/blob/0ba1a04c05de04fba9d09ae5b7b33b2399541dca/phpseclib/Crypt/Random.php#L73

Probably should pass `MCRYPT_DEV_URANDOM` to `mcrypt_create_iv()` ;)

GrahamCampbell:
So we need to catch Throwable rather than Error?

GrahamCampbell:
> Probably should pass MCRYPT_DEV_URANDOM to mcrypt_create_iv() ;)

Could you send a PR please?

paragonie-scott:
Sure. Want me to address both?

GrahamCampbell:
Could you do it in two PRs please?

terrafrost:
I'm gonna close this since it's addressed by PR's now.

bantu:
@terrafrost Please do not close ticket that have not actually been addressed yet (i.e. PR actually merged).

PR #845, opened 2015-10-09
Fix usage of mcrypt_create_iv() on Windows.

paragonie-scott:
https://github.com/php/php-src/blob/ac750e091e1c72b2724bea60a3fa9c57d06dc227/ext/mcrypt/mcrypt.c#L1359-L1397

If you pass `MCRYPT_RAND` on Windows (which *used to be the default*), it won't use a CSPRNG. Passing `MCRYPT_DEV_URANDOM` is a good safety measure.

See #843 as well.

### DIFF
--- a/phpseclib/Crypt/Random.php
+++ b/phpseclib/Crypt/Random.php
@@ -70,7 +70,7 @@ class Random
             // method 1. prior to PHP 5.3 this would call rand() on windows hence the function_exists('class_alias') call.
             // ie. class_alias is a function that was introduced in PHP 5.3
             if (extension_loaded('mcrypt') && function_exists('class_alias')) {
-                return mcrypt_create_iv($length);
+                return mcrypt_create_iv($length, MCRYPT_DEV_URANDOM);
             }
             // method 2. openssl_random_pseudo_bytes was introduced in PHP 5.3.0 but prior to PHP 5.3.4 there was,
             // to quote <http://php.net/ChangeLog-5.php#5.3.4>, "possible blocking behavior". as of 5.3.4


GrahamCampbell:
I wonder. Does Laravel's mcrypt encrypter also suffer this issue?

paragonie-scott:
No. https://github.com/laravel/framework/blob/5.1/src/Illuminate/Encryption/McryptEncrypter.php#L189

GrahamCampbell:
Ah, good. :)

terrafrost:
Why do you say that `MCRYPT_RAND` used to be the default?

Here's what the PHP docs say of `mcrypt_create_iv`:

> Prior to 5.3.0, **MCRYPT_RAND** was the only one supported on Windows.

So if it was the only one supported then it stands to reason that it'd default. But when I look at PHP 5.2 branch on php-src on github.com I see this:

https://github.com/php/php-src/blob/PHP-5.2/ext/mcrypt/mcrypt.c#L1214

ie. there's `long source = RANDOM` and then the `if (source == RANDOM || source == URANDOM) {` check takes place. At this point `php_win32_get_random_bytes` is not called (although it is in later PHP versions) but rather `fd = open(source == RANDOM ? "/dev/random" : "/dev/urandom", O_RDONLY);` is. Since Windows doesn't have that that'd result in an error. Which is actually exactly what happens when I try it out on Windows with PHP 5.2.17:

> Warning: mcrypt_create_iv(): Cannot open source device

Let's now fast forward to PHP 5.3:

https://github.com/php/php-src/blob/PHP-5.3/ext/mcrypt/mcrypt.c#L1373

Now `php_win32_get_random_bytes` is called if it's Windows. `php_win32_get_random_bytes` then calls `CryptAcquireContext` and then `CryptGenRandom` per this:

https://github.com/php/php-src/blob/PHP-5.3/win32/winutil.c#L80

So it doesn't seem to me like `MCRYPT_RAND` ever was the default? I mean, it seems like it /should/ have been the default in PHP 5.2 but it doesn't appear to be.

terrafrost:
In Crypt/Random.php there's this:

```php
             // method 1. prior to PHP 5.3 this would call rand() on windows hence the function_exists('class_alias') call.
            // ie. class_alias is a function that was introduced in PHP 5.3
            if (extension_loaded('mcrypt') && function_exists('class_alias')) {
                return mcrypt_create_iv($length);
            }
```
So it doesn't do `mcrypt_create_iv` unless you're on PHP 5.3. But that said as for why I say in the comments that "prior to PHP 5.3 this would call rand()" isn't entirely clear to me given my earlier analysis and testing...

bantu:
What is the status here?

paragonie-scott:
It looks like you're using urandom on Linux correctly, and I don't think the constant matters on Windows. If you want to close it, feel free.

PR #844, opened 2015-10-09
Update random_bytes() to conform to PHP7.0.0-RC3

paragonie-scott:
Fixes #843

PR #846, opened 2015-10-10
Use paragonie/random-compat

cweagans:
`paragonie/random-compat` is a polyfill for `random_bytes()`. Leveraging it allows phpseclib to dump a lot of code and not lose any functionality.

Note that this PR also includes https://github.com/phpseclib/phpseclib/pull/844

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -51,7 +51,8 @@
         }
     ],
     "require": {
-        "php": ">=5.3.3"
+        "php": ">=5.3.3",
+        "paragonie/random_compat": "~1.0"
     },
     "require-dev": {
         "phing/phing": "~2.7",
--- a/phpseclib/Crypt/Random.php
+++ b/phpseclib/Crypt/Random.php
@@ -54,66 +54,21 @@ class Random
      */
     static function string($length)
     {
-        if (version_compare(PHP_VERSION, '7.0.0', '>=')) {
-            try {
-                return random_bytes($length);
-            } catch (\Error $e) {
-                // If a sufficient source of randomness is unavailable, random_bytes() will emit a warning.
-                // We don't actually need to do anything here. The string() method should just continue
-                // as normal. Note, however, that if we don't have a sufficient source of randomness for
-                // random_bytes(), most of the other calls here will fail too, so we'll end up using
-                // the PHP implementation.
-            }
+        // paragonie/random_compat defines random_bytes() for us if we're using
+        // PHP < 7.0.0. If it can't find any good sources of randomness,
+        // random_bytes() is defined as a function that just throws an exception.
+        try {
+            return \random_bytes($length);
+        } catch (\Throwable $e) {
+            // If a sufficient source of randomness is unavailable, random_bytes() will throw an
+            // object that implements the Throwable interface (Exception, TypeError, Error).
+            // We don't actually need to do anything here. The string() method should just continue
+            // as normal. Note, however, that if we don't have a sufficient source of randomness for
+            // random_bytes(), most of the other calls here will fail too, so we'll end up using
+            // the PHP implementation.
         }
 
-        if (strtoupper(substr(PHP_OS, 0, 3)) === 'WIN') {
-            // method 1. prior to PHP 5.3 this would call rand() on windows hence the function_exists('class_alias') call.
-            // ie. class_alias is a function that was introduced in PHP 5.3
-            if (extension_loaded('mcrypt') && function_exists('class_alias')) {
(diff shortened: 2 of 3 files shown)


phpseclib/Crypt/Random.php
```diff
@@ -54,66 +54,21 @@ class Random
...
-                // as normal. Note, however, that if we don't have a sufficient source of randomness for
-                // random_bytes(), most of the other calls here will fail too, so we'll end up using
-                // the PHP implementation.
-            }
+        // paragonie/random_compat defines random_bytes() for us if we're using
+        // PHP < 7.0.0. If it can't find any good sources of randomness,
+        // random_bytes() is defined as a function that just throws an exception.
+        try {
+            return \random_bytes($length);
+        } catch (\Throwable $e) {
```
GrahamCampbell: does throwable exist on php 5.6 though?

terrafrost:
https://raw.githubusercontent.com/paragonie/random_compat/master/lib/random_bytes_openssl.php

Per the comments of https://github.com/phpseclib/phpseclib/blob/5745c8a5beed2d98625f0bbc3a238a8a3387c49d/phpseclib/Crypt/Random.php#L75 there's "possible blocking behavior" in openssl_random_pseudo_bytes prior to PHP 5.3.4 that random_bytes_openssl doesn't take into account. Although the comments in phpseclib's Crypt/Random.php don't elaborate I believe the issue was due OpenSSL's `RAND_pseudo_bytes` trying to gather entropy from the screen and blocking until sufficient entropy was gathered. 

We can download PHP 5.3.2 for Windows here:

http://windows.php.net/downloads/qa/test/5.3.2-final/

phpseclib/Crypt/Random.php
cweagans: random-compat has an error polyfill: https://github.com/paragonie/random_compat/blob/master/lib/error_polyfill.php

cweagans:
openssl is prioritized last in random-compat for a reason - see last paragraph of https://github.com/paragonie/random_compat/blob/master/ERRATA.md.

If somebody is running a PHP app on a Windows server with a (very) old version of PHP installed and nothing else other than openssl, it will *still* use random_bytes() backed by the `com_dotnet` extension first, which is enabled by default on PHP > 5.3.15.

phpseclib/Crypt/Random.php
GrahamCampbell: Nice :)

GrahamCampbell: Hmm, well, assuming we're 100% certain all exceptions will defientely actually implement throwable on 5.6.

cweagans: The exceptions that the random-compat lib throws definitely implement Throwable (also in the error polyfill). In 7.0, *all* exceptions implement throwable.

GrahamCampbell: Yeh. I was just checking to be sure there's zero chance of an exception coming out of there that might not implement throwable.

cweagans: Fair enough. There's definitely a 0% chance.

cweagans:
Note also that if the PHP 5.4 requirement lands, the openssl version check becomes even less of an issue.

phpseclib/Crypt/Random.php
cweagans: So. Turns out I was wrong. https://github.com/paragonie/random_compat/pull/47

We'd need to catch both `\Throwable` and `\Exception` in this block.

GrahamCampbell: lol

cs278: See https://github.com/paragonie/random_compat/pull/56 you are better off catching `\Error` and `\Exception` that way you get the same execution path on PHP 5 and PHP 7.

bantu:
@terrafrost What is the status here?

composer.json
```diff
@@ -51,7 +51,8 @@
         }
     ],
     "require": {
-        "php": ">=5.3.3"
+        "php": ">=5.3.3",
+        "paragonie/random_compat": "~1.0"
```
GrahamCampbell: maybe `~1.1` would be a better idea here?

terrafrost:
> @terrafrost What is the status here?

I'm thinking about just letting this sit until a decision has been made on adding Composer as a pre-requisite, which kinda ties into #763. And since master is a good ways off, anyway, I'm in no particular rush to make a decision on this atm..

terrafrost:
This has been merged with #958

PR #850, opened 2015-10-15
Fixed a bug in SFTP::chmod not using realpath on the filepath

Bob4ever:


terrafrost:
I'll take a look this weekend. In the mean time it'd be cool if you could provide a unit test demonstrating the problem that this is fixing. If not that's okay too - maybe I can reproduce the problem that this is attempting to fix when I've had a chance to examine it in more depth.

Either way, thanks!

Bob4ever:
Here is a unit test to reproduce the bug:

``` php
class SftpTest extends \PHPUnit_Framework_TestCase
{
    public function testChmod()
    {    
        $sftp = new \phpseclib\Net\SFTP('HOST');
        $this->assertTrue($sftp->login('USER', 'PASSWORD'), 'Login failed');
        $this->assertTrue($sftp->chdir('/var/www/webs/test_project'), 'Chdir failed');
        $this->assertTrue($sftp->put('Asdf.txt', 'Hello World'), 'File write failed');
        $this->assertNotFalse($sftp->chmod(0775, 'Asdf.txt'), 'Chmod failed');
    }
}
```

bantu:
Confirming issue.

bantu:
Replacing this with #852 which contains a test case and is against 1.0. Thanks for your patch/report.

Issue #474, opened 2014-09-10
Using TOR or PROXY to login  SSH


Leanerz:
Hi,
I am looking for a way to add proxy / TOR to connect to SSH using phpseclib !

You can show me all the request connection used by the script to modify them and add proxy ?

Thank you in advance :)

PR #872, opened 2015-11-17
Proxy support

auro1:
If anyone wants to give a hand, just put a comment here and I'll grant you write access to the repo.
Stolen code from https://github.com/phpseclib/phpseclib/issues/474

DO __NOT__ MERGE THIS YET

### DIFF
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -1026,50 +1026,57 @@ class SSH2
 			/**
 				EDIT START
 			**/
-			$socks['port'] = $this->proxyPort; 
-			$socks['address'] = $this->proxyHost; 
+			if (!empty($this->proxyHost))
+			{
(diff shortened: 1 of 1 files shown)


terrafrost:
It'd be easier to review the code if all the commits were squashed into one, altho idk if you're ready for it to be reviewed yet.

Anyway, I think, in theory, #746 kinda eliminates the need for this. Setup the SOCKS5 / HTTP(S) proxy outside of the SSH class and then pass the socket resource to the Net_SSH2 constructor.

faridmovsumov:
@auro1 I have the same problem. Thanks for trying solve this problem. Does it work?

BreakHear7:
Does it work with SSH?

terrafrost:
#1460 demos SOCKS5 proxy support. https://github.com/phpseclib/phpseclib/issues/1339#issuecomment-462224179 demonstrates HTTP proxy support.

PR #875, opened 2015-11-19
README: Use the svg Travis image

bantu:
Replaces #874


bantu:
Updated 2.0 and master on merge.

PR #876, opened 2015-11-19
Fix PHP invalid index warning

splitice:
Fix PHP Notice: Undefined offset: -1 in getLastError

### DIFF
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -3774,7 +3774,11 @@ class SSH2
      */
     function getLastError()
     {
-        return $this->errors[count($this->errors) - 1];
+        $count = count($this->errors);
+
+        if ($count > 0) {
+            return $this->errors[$count - 1];
+        }
     }
 
     /**


GrahamCampbell:
Please follow PSR-2.

GrahamCampbell:
Also, what version of php>

splitice:
Every version of PHP since PHP 4 has had a warning for invalid index offset.

phpseclib/Net/SSH2.php
```diff
@@ -3774,7 +3774,11 @@ function getErrors()
      */
     function getLastError()
     {
-        return $this->errors[count($this->errors) - 1];
+        $count = count($this->errors);
+        if($count == 0) {
```
GrahamCampbell: missing space after if

phpseclib/Net/SSH2.php
```diff
@@ -3774,7 +3774,11 @@ function getErrors()
...
+            return null;
```
GrahamCampbell: just `return;` is fine

GrahamCampbell: or maybe, just write:

```php
$count = count($this->errors);

if($count > 0) {
    return $this->errors[$count - 1];
}
```

splitice:
changed

terrafrost:
Looks good! Can you squash the commits into one? I suppose I can do that too. If you don't I'll probably do so in a few days. I'm traveling atm so my availability is a bit sporadic.

splitice:
Should be good now

bantu:
Patch looks good to me. This is a bug fix though and will have to go into 1.0 and 2.0 too. I'll take care of that.

bantu:
Replaced by #883

PR #877, opened 2015-11-19
Enhanced disconnect on login failure handling

dropdevcoding:
added disconnects to failure returns in _login_helper because otherwise MASK::CONNECTED will stay set (since it automatically calls _connect), although credentials are invalid or all supported authentication methods have been tried already.

I discovered this issue while using thephpleague/flysystem-sftp since they use isConnected() to check if a login is necessary. The first time this lib throws an exception while at the second attempt no exception will be thrown anymore and this is caused by Net/SFTP because it does not disconnect properly.

If you don't see this fix is a fix to be made inside your lib because it is not generic enough, please let me know so I possibly may implement a workaround calling disconnect for that specific case.

### DIFF
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -2051,15 +2051,15 @@ class SSH2
                         $this->bitmap |= self::MASK_LOGIN;
                         return true;
                     }
-                    return false;
+                    return $this->_disconnect(NET_SSH2_DISCONNECT_NO_MORE_AUTH_METHODS_AVAILABLE);
                 }
-                return false;
+                return $this->_disconnect(NET_SSH2_DISCONNECT_BY_APPLICATION);
             case NET_SSH2_MSG_USERAUTH_SUCCESS:
                 $this->bitmap |= self::MASK_LOGIN;
                 return true;
         }
 
-        return false;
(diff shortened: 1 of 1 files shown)


bantu:
@peterhildenbrand Please consider specifying a name and an e-mail address in the commit.

peterhildenbrand:
I am sorry, this is the first time I use this machine to commit to github. Configuration was not set.

terrafrost:
Also, for the delayed response - I was in DC for a PHP conference week before last and then last week I was in NYC on vacation.

peterhildenbrand:
I understand, so from a general point of view the implementation works as expected. isLoggedIn might not be the right term in that case, but I would appreciate if there was a function I could ask for "am I currently in a state, in which I may operate on the foreign server?". Having this function offered would help a lot to handle that error more accurately/cleanly instead of handling it all-over-the-board "okay, if there's an error, disconnect and try again".

I will absolutely understand if you don't see the need for that, but I don't know other projects which are using this library, so it might be helpful for other cases, too.

bantu:
I think it makes sense to have the functionality described in the previous
comment.

terrafrost:
The newly added `isAuthenticated()` method should satisfy the problems this PR was attempting to address.

dropdevcoding:
Thank you very much!

PR #881, opened 2015-11-22
small tweaks

terrafrost:
If you're using gmp mode for BigInteger var_dump / print_r don't produce very usable output. Also, make a small CS change to ASN1.php.

### DIFF
--- a/phpseclib/Math/BigInteger.php
+++ b/phpseclib/Math/BigInteger.php
@@ -820,7 +820,28 @@ class Math_BigInteger
      */
     function __debugInfo()
     {
-        return array('value' => '0x' . $this->toHex(true));
+        $opts = array();
+        switch (MATH_BIGINTEGER_MODE) {
+            case MATH_BIGINTEGER_MODE_GMP:
+                $engine = 'gmp';
+                break;
+            case MATH_BIGINTEGER_MODE_BCMATH:
+                $engine = 'bcmath';
+                break;
+            case MATH_BIGINTEGER_MODE_INTERNAL:
+                $engine = 'internal';
+                $opts[] = PHP_INT_SIZE == 8 ? '64-bit' : '32-bit';
+        }
+        if (MATH_BIGINTEGER_MODE != MATH_BIGINTEGER_MODE_GMP && defined('MATH_BIGINTEGER_OPENSSL_ENABLED')) {
+            $opts[] = 'OpenSSL';
+        }
+        if (!empty($opts)) {
+            $engine.= ' (' . implode($opts, ', ') . ')';
+        }
+        return array(
+            'value' => '0x' . $this->toHex(true),
+            'engine' => $engine
+        );
     }
 
     /**
--- a/phpseclib/Math/BigInteger.php
+++ b/phpseclib/Math/BigInteger.php
@@ -812,6 +812,18 @@ class Math_BigInteger
     }
 
     /**
+     *  __debugInfo() magic method
+     *
+     * Will be called, automatically, when print_r() or var_dump() are called
+     *
+     * @access public
+     */
+    function __debugInfo()
+    {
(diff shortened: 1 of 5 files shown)


phpseclib/Math/BigInteger.php
```diff
@@ -812,6 +812,18 @@ function __wakeup()
     }
 
     /**
+     *  __debugInfo() magic method
+     *
+     * Will be called, automatically, when print_r() or var_dump() are called
+     *
+     * @access public
+     */
+    function __debugInfo()
```
bantu: tests?

phpseclib/Math/BigInteger.php
```diff
@@ -812,6 +812,18 @@ function __wakeup()
...
+    {
+        return array('value' => '0x' . $this->toHex(true));
```
bantu: How about we also include information such as driver?

terrafrost:
Well the unit test is failing and idk why.

__debugInfo, as a magic method, only works in PHP 5.6, per http://php.net/manual/en/language.oop5.magic.php#object.debuginfo

I'm using the @requires annotation as mentioned at https://phpunit.de/manual/current/en/incomplete-and-skipped-tests.html#incomplete-and-skipped-tests.requires.tables.api to make it so that the test will only run on PHP 5.6 and later but it doesn't appear to be working.

And the fact that it's seg faulting on PHP 5.6...  idk what that's all about. It's not seg faulting my copy of PHP 5.6..

tests/Unit/Math/BigInteger/TestCase.php
```diff
@@ -358,4 +358,15 @@ public function testDiffieHellmanKeyAgreement()
             'Failed asserting that Alice and Bob share the same BigInteger.'
         );
     }
+
+    /*
```
Joey3000: Can you try a double star here, i.e. `/**` instead of `/*`? That's how the example in https://phpunit.de/manual/current/en/incomplete-and-skipped-tests.html#incomplete-and-skipped-tests.requires.tables.api does it.

bantu: +1

Joey3000:
> And the fact that it's seg faulting on PHP 5.6... idk what that's all about. It's not seg faulting my copy of PHP 5.6..

Just a few ideas on that:

* runkit:
  * A new version 1.0.4 of runkit has been released, according to http://zenovich.blogspot.com/2015/10/runkit-1.0.4.html#rel104 (see also https://github.com/zenovich/runkit/releases). It seems to fix a bunch of segfault issues.
  * Also, it now supports PHP 5.6, so that a PHP version check in https://github.com/phpseclib/phpseclib/blob/master/travis/install-php-extensions.sh#L24-L29 would no longer be needed.
  * I'm not familiar with Travis CI, but https://github.com/phpseclib/phpseclib/blob/master/travis/install-php-extensions.sh#L24-L29 uses `git clone` of the runkit repository, which probably means that it uses runkit it its current default branch (i.e. `master`) snapshot. Maybe, for stability reasons, it would be better to use a defined version instead, i.e., to download one of the source code files attached to https://github.com/zenovich/runkit/releases/tag/1.0.4 using `wget`? Those can't be modified post-release. Alternatively, to make it faster and even less error-prone, the used runkit version could be included into phpseclib, using e.g. following directory structure (see, e.g., https://github.com/piwik/piwik/tree/master/tests):
    * `tests/lib/runkit` (contains runkit)
    * `tests/travis` (contains what's currently in the `travis` directory)
    * `tests/PHPUnit` (contains what's currently in the `tests` directory, as well as `phpunit.xml.dist`, which is currently in the root directory)

* https://stackoverflow.com/questions/6343674/phpunit-segmentation-fault and various other found searching for "php unit test segmentation fault" and "travis ci unit test segmentation fault".

PR #884, opened 2015-11-25
Fix indentation

DavidPrevot:
Automatically fixed by “phpcbf --extensions=php --standard=build/code-sniffer-ruleset.xml tests”.


terrafrost:
Good catch - looks good!

GrahamCampbell:
:+1:

bantu:
Let's see what we have to do with this in 2.0. I'll look into it.

bantu:
Ahahahahahaha lol.

bantu:
This patch is wrong and not acceptable. Only the line containing `'pvno' => array(` is improperly indented ($n mod 4 != 0). `phpcbf` incorrectly adds a space to lines with similar indentation as the affected line. This can be easily seen by a) looking at the affected block pre-patch or b) counting spaces in post-patch (my editor shows me that there is an extra space).

Issue #882, opened 2015-11-23
Recommendations by Paragon Initiative Enterprises, LLC


Joey3000:
Please see the "Phpseclib" part of the following blog post: https://paragonie.com/blog/2015/11/choosing-right-cryptography-library-for-your-php-project-guide. As well as the subsequent "Update".

E.g.:
* AES:

> phpseclib's AES implementation does not authenticate ciphertexts, so if you use it, you will be vulnerable to chosen-ciphertext attacks. (Their interactive documentation also dangerously defaults to ECB mode, which is the worst block cipher mode.)

* RSA:

> RSA Encryption: RSA-OAEP, taking care to $rsa->setMGFHash('sha256') first
RSA Signatures: RSA-PSS, once again explicitly using MGF1+SHA256

Whereas according to http://phpseclib.sourceforge.net/rsa/examples.html#encrypt,enc1, "the default hash algorithm is sha1" with phpseclib.

Disclaimer: I have nothing to do with them.

terrafrost:
First of all, I apologize for the tardyness of this response - week before last I was in DC for a PHP conference and then past week I was in NYC for Thanksgiving.

Anyway, authenticated ciphertext's would be good but it's not, atm, super high on my priority list. Right now I want to rewrite the X509 stuff and then make a lot of improvements to the SSH stuff as well. And then DSA, etc.

As for phpseclib's documentation defaulting to ECB mode...  that's something I can change. That said if you don't pass any parameters to the constructor the symmetric ciphers default the CBC mode.

As for changing the default hash of OAEP / PSS to sha256....  the default per PKCS1 v2.1 is sha1. Quoting it:

https://tools.ietf.org/html/rfc3447#section-7.1

```
      RSAES-OAEP-params ::= SEQUENCE {
          hashAlgorithm     [0] HashAlgorithm    DEFAULT sha1,
          maskGenAlgorithm  [1] MaskGenAlgorithm DEFAULT mgf1SHA1,
          pSourceAlgorithm  [2] PSourceAlgorithm DEFAULT pSpecifiedEmpty
      }
```

[...]

PR #902, opened 2015-12-25
RSA: changes to how encryption / signatures work

terrafrost:
- setEncryptionMode() / setSignatureMode() methods have been removed.
  instead you'll do `$rsa->verify($message, $signature, RSA::PADDING_PKCS1)` or whatever. the defaults remain the same as before. ie. `$rsa->verify($message, $signature)` still does PSS signatures (altho it now does them with sha256 for the hash and MGF). i thought about making it so that `$rsa->verify($message, $signature)` would try both `RSA::PADDING_PSS` and `RSA_PADDING_PKCS` in the same way that RSA.php brute forces key formats but decided against it. the reason being that a lot of the success of `verify()` and `decrypt()` depend on the hash algorithms being used and i don't want to brute force those as well.

- added a new signature verification mode - RSA::PADDING_RELAXED_PKCS1.
  in rare cases it might succeed where RSA_PADDING_PKCS1 doesn't. see the comments in RSA.php and the unit test for more info.

- added a new encryption mode - RSA::PADDING_PKCS15_COMPAT.
  replaces `define('CRYPT_RSA_PKCS15_COMPAT', true)`.

- the default hash algorithm is now sha256.
  i had considered making the MGF1 hash algorithm still stay as sha1 since that's [kinda what Java does](http://stackoverflow.com/questions/32161720/breaking-down-rsa-ecb-oaepwithsha-256andmgf1padding) with `RSA/ECB/OAEPWITHSHA-256ANDMGF1PADDING` but i ultimately decided against it since that's [not what OpenSSL does for PSS signatures](http://crypto.stackexchange.com/questions/31430/setting-the-mgf-hash-to-a-different-value-other-than-the-regular-hash-with-opens). if you want to be compatable with `RSA/ECB/OAEPWITHSHA-256ANDMGF1PADDING` you'll need to manually change the MGF hash instead of having it work out of the box.

- return false instead of throwing exceptions on errors.
  http://tools.ietf.org/html/rfc3447#page-32 says to "output 'invalid signature' and stop" on various errors, hence the throwing of exceptions. however, it also says to "output 'invalid signature'" or   "valid signature" depending on the output of EMSA-PSS-VERIFY. so, pretty much, anytime only "invalid signature" is being thrown we're now instead returning false. if a different message other than "invalid signature" is to be output we'll continue to use exceptions.

- PKCS1 v2.2 remains unsupported.
  the difference between v2.1 and v2.2 is that v2.2 adds support for sha512/224 and sha512/256 as hashes. the problem is that PHP's built-in hash algorithm does not support them. a bug ticket exists for this:
  https://bugs.php.net/71177
  phpseclib's Hash class, in the 1.0 and 2.0 branches, has pure-PHP implementations of sha256 and sha512, which it uses if the hash extension (or mhash) is unavailable. PHP 5.3 included the hash extension by default, which kinda rendered the pure-PHP implementations moot and thus they were removed in the master branch, but maybe we could go back to this approach for sha512/224 and sha512/256. but that can be done in another PR too if there's sufficient interest in it.

### DIFF
--- a/phpseclib/Crypt/RSA.php
+++ b/phpseclib/Crypt/RSA.php
@@ -45,8 +45,6 @@
 
 namespace phpseclib\Crypt;
 
-use phpseclib\Crypt\Hash;
-use phpseclib\Crypt\Random;
 use phpseclib\Math\BigInteger;
 use phpseclib\File\ASN1;
 
@@ -1683,6 +1681,7 @@ class RSA
      *
      * @access private
      * @param string $m
+     * @param bool $pkcs15_compat optional
      * @throws \OutOfBoundsException if strlen($m) > $this->k - 11
      * @return string
      */
@@ -1870,7 +1869,7 @@ class RSA
      *
      * @access private
      * @param string $m
-     * @return string
+     * @return bool|string
      */
     function _rsassa_pss_sign($m)
     {
@@ -1940,9 +1939,6 @@ class RSA
     function _emsa_pkcs1_v1_5_encode($m, $emLen)
     {
         $h = $this->hash->hash($m);
-        if ($h === false) {
-            return false;
-        }
 
         // see http://tools.ietf.org/html/rfc3447#page-43
         switch ($this->hashName) {
@@ -1986,7 +1982,7 @@ class RSA
      * @access private
      * @param string $m
      * @throws \LengthException if the RSA modulus is too short
-     * @return string
+     * @return bool|string
      */
     function _rsassa_pkcs1_v1_5_sign($m)
     {
@@ -1994,9 +1990,10 @@ class RSA
 
         // If the encoding operation outputs "intended encoded message length too short," output "RSA modulus
         // too short" and stop.
-        $em = $this->_emsa_pkcs1_v1_5_encode($m, $this->k);
-        if ($em === false) {
-            return false;
+        try {
+            $em = $this->_emsa_pkcs1_v1_5_encode($m, $this->k);
+        } catch (\LengthException $e) {
+            throw new \LengthException('RSA modulus too short');
         }
 
         // RSA signature
@@ -2017,8 +2014,9 @@ class RSA
      *
      * @access private
      * @param string $m
+     * @param string $s
      * @throws \LengthException if the RSA modulus is too short
-     * @return bool|string
+     * @return bool
      */
     function _rsassa_pkcs1_v1_5_verify($m, $s)
     {
@@ -2069,7 +2067,8 @@ class RSA
      *
      * @access private
      * @param string $m
-     * @return bool|string
+     * @param string $s
+     * @return bool
      */
     function _rsassa_pkcs1_v1_5_relaxed_verify($m, $s)
     {
@@ -2166,8 +2165,8 @@ class RSA
      * @see self::decrypt()
      * @access public
      * @param string $plaintext
-     * @param int $padding
-     * @return string
+     * @param int $padding optional
+     * @return bool|string
      * @throws \LengthException if the RSA modulus is too short
      */
     function encrypt($plaintext, $padding = self::PADDING_OAEP)
@@ -2177,7 +2176,11 @@ class RSA
                 $plaintext = str_split($plaintext, $this->k);
                 $ciphertext = '';
                 foreach ($plaintext as $m) {
-                    $ciphertext.= $this->_raw_encrypt($m);
+                    $temp = $this->_raw_encrypt($m);
+                    if ($temp === false) {
+                        return false;
+                    }
+                    $ciphertext.= $temp;
                 }
                 return $ciphertext;
             case self::PADDING_PKCS15_COMPAT:
@@ -2190,7 +2193,11 @@ class RSA
                 $plaintext = str_split($plaintext, $length);
                 $ciphertext = '';
                 foreach ($plaintext as $m) {
-                    $ciphertext.= $this->_rsaes_pkcs1_v1_5_encrypt($m, $padding == self::PADDING_PKCS15_COMPAT);
+                    $temp = $this->_rsaes_pkcs1_v1_5_encrypt($m, $padding == self::PADDING_PKCS15_COMPAT);
+                    if ($temp === false) {
+                        return false;
+                    }
+                    $ciphertext.= $temp;
                 }
                 return $ciphertext;
             //case self::PADDING_OAEP:
@@ -2203,7 +2210,11 @@ class RSA
                 $plaintext = str_split($plaintext, $length);
                 $ciphertext = '';
                 foreach ($plaintext as $m) {
-                    $ciphertext.= $this->_rsaes_oaep_encrypt($m);
+                    $temp = $this->_rsaes_oaep_encrypt($m);
+                    if ($temp === false) {
+                        return false;
+                    }
+                    $ciphertext.= $temp;
                 }
                 return $ciphertext;
         }
@@ -2215,7 +2226,7 @@ class RSA
      * @see self::encrypt()
      * @access public
      * @param string $plaintext
-     * @param int|bool $padding
+     * @param int $padding optional
      * @return string
      */
     function decrypt($ciphertext, $padding = self::PADDING_OAEP)
@@ -2258,7 +2269,7 @@ class RSA
      * @see self::verify()
      * @access public
      * @param string $message
-     * @param int $padding
+     * @param int $padding optional
      * @return string
      */
     function sign($message, $padding = self::PADDING_PSS)
@@ -2284,7 +2295,7 @@ class RSA
      * @access public
      * @param string $message
      * @param string $signature
-     * @param int|bool $padding
+     * @param int $padding optional
      * @return bool
      */
     function verify($message, $signature, $padding = self::PADDING_PSS)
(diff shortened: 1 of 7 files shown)


phpseclib/Crypt/RSA.php
```diff
@@ -1917,7 +1897,6 @@ function _rsassa_pss_sign($m)
      * @access private
      * @param string $m
      * @param string $s
-     * @throws \RuntimeException on invalid signature
      * @return string
```
Joey3000: Should this be `@return bool|string` as well? And also at https://github.com/phpseclib/phpseclib/pull/902/files#diff-b4b2acaebb5e3d125481ec5da11baff5R1989 below?

phpseclib/Crypt/RSA.php
```diff
@@ -2154,9 +2215,10 @@ function encrypt($plaintext)
      * @see self::encrypt()
      * @access public
      * @param string $plaintext
+     * @param int|bool $padding
```
Joey3000: Should this be `@param int $padding`, like in the case of `encrypt()` and `sign()`?

phpseclib/Crypt/RSA.php
```diff
@@ -2154,9 +2215,10 @@ function encrypt($plaintext)
...
      * @return string
```
Joey3000: Should this be `@return bool|string` as well?

phpseclib/System/SSH/Agent/Identity.php
```diff
@@ -115,29 +114,17 @@ function getPublicKey($format = null)
...
-    {
-    }
-
-    /**
      * Create a signature
      *
      * See "2.6.2 Protocol 2 private key signature request"
      *
      * @param string $message
+     * @param int|bool $padding
```
Joey3000: Should this be `@param int $padding`, like in RSA::sign()?

phpseclib/System/SSH/Agent/Identity.php
```diff
@@ -115,29 +114,17 @@ function getPublicKey($format = null)
...
      * @return string
      * @throws \RuntimeException on connection errors
      * @access public
      */
-    function sign($message)
+    function sign($message, $padding = false)
```
Joey3000: Should this be `function sign($message, $padding = RSA::PADDING_NONE)`, to have the same type as with RSA::sign()? (Would require `use phpseclib\Crypt\RSA;` at the top of the file.)

phpseclib/Crypt/RSA.php
```diff
@@ -1705,7 +1686,7 @@ function _raw_encrypt($m)
      * @throws \OutOfBoundsException if strlen($m) > $this->k - 11
```
Joey3000: `@param bool $pkcs15_compat` seems to be missing

Joey3000:
> the default hash algorithm is now sha256

That takes care of the "RSA" part of https://github.com/phpseclib/phpseclib/issues/882. Thanks!

phpseclib/Crypt/RSA.php
```diff
@@ -48,6 +48,7 @@
 use phpseclib\Crypt\Hash;
 use phpseclib\Crypt\Random;
```
Joey3000: Following is not needed and could be removed, because they are in the same `namespace` ( therefore, `Hash` and `Random` can be used without specific `use` declarations):
```
use phpseclib\Crypt\Hash;
use phpseclib\Crypt\Random;
```

phpseclib/Crypt/RSA.php
```diff
@@ -2072,55 +2055,133 @@ function _rsassa_pkcs1_v1_5_verify($m, $s)
...
+     * that if you're going to validate these types of signatures you "should indicate
+     * whether the underlying BER encoding is a DER encoding and hence whether the signature
+     * is valid with respect to the specification given in [PKCS1 v2.0+]". so if you do
+     * $rsa->getLastPadding() and get RSA::PADDING_RELAXED_PKCS1 back instead of
+     * RSA::PADDING_PKCS1... that means BER encoding was used.
      *
-     * @access public
-     * @param int $mode
+     * @access private
+     * @param string $m
```
Joey3000: `@param string $s` seems to be missing

phpseclib/Crypt/RSA.php
```diff
@@ -2072,55 +2055,133 @@ function _rsassa_pkcs1_v1_5_verify($m, $s)
...
+     * @return bool|string
```
Joey3000: Should this be `@return bool`? Because all returns, including `$this->_equals($em, $em2)`, are of boolean type.

phpseclib/Crypt/RSA.php
```diff
@@ -2036,16 +2017,15 @@ function _rsassa_pkcs1_v1_5_sign($m)
      *
      * @access private
      * @param string $m
```
Joey3000: `@param string $s` seems to be missing

phpseclib/Crypt/RSA.php
```diff
@@ -2013,9 +1992,11 @@ function _rsassa_pkcs1_v1_5_sign($m)
     {
         // EMSA-PKCS1-v1_5 encoding
 
+        // If the encoding operation outputs "intended encoded message length too short," output "RSA modulus
+        // too short" and stop.
```
Joey3000: This comment does not seem to correspond to the code:
* There is no catching of the "intended encoded message length too short" exception
* There is no output "RSA modulus too short" being made

phpseclib/Crypt/RSA.php
```diff
@@ -2053,17 +2033,20 @@ function _rsassa_pkcs1_v1_5_verify($m, $s)
...
 
         // EMSA-PKCS1-v1_5 encoding
 
-        $em2 = $this->_emsa_pkcs1_v1_5_encode($m, $this->k);
-        if ($em2 === false) {
+        // If the encoding operation outputs "intended encoded message length too short," output "RSA modulus
+        // too short" and stop.
+        try {
+            $em2 = $this->_emsa_pkcs1_v1_5_encode($m, $this->k);
+        } catch (\LengthException $e) {
```
Joey3000: This does not account for `_emsa_pkcs1_v1_5_encode()` possibly returning `false` (in https://github.com/phpseclib/phpseclib/blob/806249c7e91b60e81ae3052cf02a83b57ac74c7d/phpseclib/Crypt/RSA.php#L1944). Is that intentional?

phpseclib/Crypt/RSA.php
```diff
@@ -2220,18 +2284,21 @@ function sign($message)
      * @access public
      * @param string $message
      * @param string $signature
+     * @param int|bool $padding
```
Joey3000: Should this be `@param int $padding`?

phpseclib/Crypt/RSA.php
```diff
@@ -2072,55 +2055,133 @@ function _rsassa_pkcs1_v1_5_verify($m, $s)
...
      *
-     * Both self::ENCRYPTION_OAEP and self::ENCRYPTION_PKCS1 both place limits on how long $plaintext can be.
+     * Both self::PADDING_OAEP and self::PADDING_PKCS1 both place limits on how long $plaintext can be.
      * If $plaintext exceeds those limits it will be broken up so that it does and the resultant ciphertext's will
      * be concatenated together.
      *
      * @see self::decrypt()
      * @access public
      * @param string $plaintext
+     * @param int $padding
```
Joey3000: Should this be `@param int $padding optional`, as seems to be the convention for optional parameters?

phpseclib/Crypt/RSA.php
```diff
@@ -2196,18 +2258,20 @@ function decrypt($ciphertext)
      * @see self::verify()
      * @access public
      * @param string $message
+     * @param int $padding
```
Joey3000: Should this be `@param int $padding optional`, as seems to be the convention for optional parameters?

phpseclib/Crypt/RSA.php
terrafrost: For all practical purposes `_emsa_pkcs1_v1_5_encode()` can't return false. http://php.net/manual/en/function.hash.php doesn't return false in the master branch (where Hash.php was rewritten to be a wrapper for PHP's builtin hash extension) and even when phpseclib had it's own implementations of select hash algorithms none of them appear to return false either:

https://github.com/phpseclib/phpseclib/blob/1.0/phpseclib/Crypt/Hash.php

Here's what the RFC says:

```php
   1. Apply the hash function to the message M to produce a hash value
      H:

         H = Hash(M).

      If the hash function outputs "message too long," output "message
      too long" and stop.
```
According to https://en.wikipedia.org/wiki/Template:Comparison_of_SHA_functions you'd have to have a message 2048 petabytes long for that error to be output and if you're dealing with data of that length PHP is probably not the right language for you lol.

In light of this I've just removed the whole `return false` thing.

terrafrost:
Good observations - I've updated the code - thanks!

Joey3000:
This review round is complete. Feel free to have a look. Only the comments appearing in this PR since your last comment apply. (The older ones, if still visible, have been marked as completed (green checkbox).)

Alternatively, the new comments can be seen here: https://github.com/terrafrost/phpseclib/commit/d5a359b41e87024ff9f49a06a1b2a703a36d38df.

terrafrost:
The suggested updates have been made - thanks!

Joey3000:
I have no further comments. Thanks for the updates!

PR #903, opened 2015-12-27
BigInteger: make select methods static, add type hinting, etc

### LINKED ISSUE #568: make random() and randomPrime() static methods
Maybe compare() and equals() ought to be static methods to? idk.

I think there's justification for keeping methods like add() non-static because that's analogous to $a+= $x but there's no analog to that for compare() or equals() (or random() or randomPrime())

terrafrost:
This PR does the following:

- replace `$x->copy()` with `clone $x` and rm the `copy()` method

- makes a bunch of private methods static and makes `random()` and `randomPrime()` static.
  in so doing this PR resolves #568

- adds type hinting to public methods

### DIFF
--- a/phpseclib/Math/BigInteger.php
+++ b/phpseclib/Math/BigInteger.php
@@ -527,7 +527,7 @@ class BigInteger
                 return $this->precision > 0 ? str_repeat(chr(0), ($this->precision + 1) >> 3) : '';
             }
 
-            $temp = $comparison < 0 ? $this->add(new static(1)) : $this->copy();
+            $temp = $comparison < 0 ? $this->add(new static(1)) : $this;
             $bytes = $temp->toBytes();
 
             if (empty($bytes)) { // eg. if the number we're trying to convert is -1
@@ -580,13 +580,11 @@ class BigInteger
         if (!count($this->value)) {
             return $this->precision > 0 ? str_repeat(chr(0), ($this->precision + 1) >> 3) : '';
         }
-        $result = $this->_int2bytes($this->value[count($this->value) - 1]);
+        $result = self::_int2bytes($this->value[count($this->value) - 1]);
 
-        $temp = $this->copy();
-
-        for ($i = count($temp->value) - 2; $i >= 0; --$i) {
-            $temp->_base256_lshift($result, self::$base);
-            $result = $result | str_pad($temp->_int2bytes($temp->value[$i]), strlen($result), chr(0), STR_PAD_LEFT);
+        for ($i = count($this->value) - 2; $i >= 0; --$i) {
+            self::_base256_lshift($result, self::$base);
+            $result = $result | str_pad(self::_int2bytes($this->value[$i]), strlen($result), chr(0), STR_PAD_LEFT);
         }
 
         return $this->precision > 0 ?
@@ -691,7 +689,7 @@ class BigInteger
             return '0';
         }
 
-        $temp = $this->copy();
+        $temp = clone $this;
         $temp->is_negative = false;
 
         $divisor = new static();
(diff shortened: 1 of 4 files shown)


phpseclib/Math/BigInteger.php
```diff
@@ -752,18 +750,17 @@ function __toString()
...
     function __clone()
     {
-        return $this->copy();
+        $temp = new static();
+        $temp->value = $this->value;
+        $temp->is_negative = $this->is_negative;
+        $temp->precision = $this->precision;
+        $temp->bitmask = $this->bitmask;
+        return $temp;
     }
```
Joey3000: Since this is PHP >= 5 code (because type hinting was added below), could it be that the whole `__clone()` method can be removed? (It looks like it's only needed for modifying **internal objects** after cloning, not **internal variables**. And `value`, `is_negative`, `precision` and `bitmask` are not objects - their values should be copied automatically.) See, e.g., https://stackoverflow.com/questions/2144506/what-is-object-cloning-in-php.

terrafrost: No - it's needed. The following demonstrates why:

```php
<?php
include('Math/BigInteger.php');

$a = new Math_BigInteger(5);
$b = $a;
$b->setPrecision(4);

echo $a->precision;
```
That'll output 4 when it ought to output -1.

clone may do a shallow copy but it none-the-less does a copy of the primitives. PHP4 cloned every time you assigned an object to a new variable - PHP5 doesn't. So to make code that worked on both the `copy()` method was added since `__clone()` couldn't be called directly on PHP5. But that all changed when phpseclib introduced namespacing. That upped the version requirement to 5.3 and eliminated the need for a `copy()` method to emulate `__clone()`.

http://www.php.net/manual/en/migration5.oop.php elaborates:

> In previous versions of PHP, objects were handled like primitive types (for instance integers and strings). The drawback of this method was that semantically the whole object was copied when a variable was assigned, or passed as a parameter to a method. In the new approach, objects are referenced by handle, and not by value (one can think of a handle as an object's identifier).

Altho that said, that example does make me think a `getPrecision` method is needed lol

Joey3000: > Altho that said, that example does make me think a `getPrecision` method is needed lol

Definitely! :)

Thanks for looking into this!

I think I didn't explain myself very well. I did not mean that one should stop using the `clone` keyword and go back to using the assignment operator ("="). **Instead, I meant that when one uses the `clone` keyword, one does not need to define a `__clone()` method.** Unless the cloned object contains other objects.

I tried following code:
```
class MyClass
{
    public $a;
    public $b;
}

$o = new MyClass;

$o->a = 'a';
$o->b = 'b';

$o2= clone $o;
$o2->a = 'c';

var_dump($o);
var_dump($o2);
```

Which produces following:
```
object(MyClass)#1 (2) {
  ["a"]=>
  string(1) "a"
  ["b"]=>
  string(1) "b"
}
object(MyClass)#2 (2) {
  ["a"]=>
  string(1) "c"
  ["b"]=>
  string(1) "b"
}
```

As can be seen, `MyClass` does not define a `__clone()` method. Nevertheless, the properties were truly copied - they are not references like in the case of an assignment. Your quote from the manual talks about **objects inside the cloned object**. So, if the $a and $b in my example were objects, then yes, I would need to define the `__clone()` method like you do in your code, and copy them manually. Because otherwise they would be references to the original object.

So, in your code, you treat `value`, `is_negative`, `precision` and `bitmask` as if they were **objects** and manually copy them. Which doesn't break anything, but is unnecessary - it is already done automatically before `__clone()` is called. The `__clone()` method here just burns cycles. :)

So, essentially, the `__clone()` method can be removed and the cloning in this case would work just as well.

terrafrost:
Updated. Thanks for the feedback!

Issue #910, opened 2016-01-03
How to setIV?


nimasdj:
Hi,

As we will not now if OpenSSL/MCrypt/Internal will be used, how we will know the iv size of method then call OpenSSL/MCrypt function to generate an IV and assign it to setIV? does phpseclib provide some methods that we can use to generate an IV?

nimasdj:
as about question above I think I found the solution:

```php
switch ($cipher->getEngine()) {

        case 3: // openssl
               $iv_size = openssl_cipher_iv_length('AES-256-CBC');
               $iv = openssl_random_pseudo_bytes($iv_size);
               break;
        case 2: // mcrypt
               $iv_size = mcrypt_get_iv_size(MCRYPT_RIJNDAEL_256, MCRYPT_MODE_CBC);
               $iv = mcrypt_create_iv($iv_size, MCRYPT_RAND);
               break;
        case 1: //internal
               $iv = just something random.
}
```
but as about internal which length do you suggest?

terrafrost:
I would just do `\phpseclib\Crypt\Random::string($iv_size)` (if you're using the 2.0 or master branches).

As for the length of the IV...  do `$cipher->getBlockLength()`.

nimasdj:
@terrafrost there is no getBlockLength() in AES class in v2. There is only one setter method. It seems you forgot to add it for AES.
I added that method in Rijndael as below and it works fine:
```php
/* start my custome */
    function getBlockLength()
    {
        return $this->block_size;
    }
/* end my custome */
```
But I am afraid whenever I update it to a new version my custom goes too. It would be good that you add this method. I am going to do a PR with this method.

PR #911, opened 2016-01-04
getIVLength() and createIV() added.

nimasdj:
getBlockLength() method added:
https://github.com/nimasdj/phpseclib/blob/master/phpseclib/Crypt/Rijndael.php#L244
See https://github.com/phpseclib/phpseclib/issues/910

### DIFF
--- a/phpseclib/Crypt/Base.php
+++ b/phpseclib/Crypt/Base.php
@@ -557,6 +557,28 @@ abstract class Base
     }
 
     /**
+     * Returns the IV size in bytes
+     *
+     * @access public
+     * @return int
+     */
+    function getIVLength()
+    {
+        return $this->getBlockLength() / 8;
+    }
+
+    /**
+     * Returns random IV bytes
+     *
+     * @access public
+     * @return string
+     */
+    function createIV()
+    {
+        return Random::string($this->getIVLength());
+    }
+
+    /**
      * Sets the key.
      *
      * The min/max length(s) of the key depends on the cipher which is used.
--- a/phpseclib/Crypt/Rijndael.php
+++ b/phpseclib/Crypt/Rijndael.php
@@ -236,17 +236,6 @@ class Rijndael extends Base
     }
 
     /**
-     * Gets the block length
-     *
-     * @access public
-     * @return int $this->block_size
-     */
-    function getBlockLength()
-    {
-        return $this->block_size;
-    }
-
-    /**
      * Test for engine validity
      *
      * This is mainly just a wrapper to set things up for \phpseclib\Crypt\Base::isValidEngine()
--- a/phpseclib/Crypt/Rijndael.php
+++ b/phpseclib/Crypt/Rijndael.php
@@ -236,6 +236,17 @@ class Rijndael extends Base
     }
 
     /**
+     * Gets the block length
+     *
+     * @access public
+     * @return int $this->block_size
+     */
+    function getBlockLength()
+    {
+        return $this->block_size;
+    }
+
+    /**
      * Test for engine validity
      *
      * This is mainly just a wrapper to set things up for \phpseclib\Crypt\Base::isValidEngine()


terrafrost:
It's already defined in Crypt/Base.php:

https://github.com/phpseclib/phpseclib/blob/master/phpseclib/Crypt/Base.php#L554

Does it not work? I'll try to test it out tomorrow the day after or something. I need to go to bed atm lol

nimasdj:
@terrafrost 
No, not working:
```
 Call to undefined method phpseclib\Crypt\AES::getBlockLength()
```

nimasdj:
@terrafrost Would you like I add two methods for base64 decode/encode IVKey if user needs to transmit it?

nimasdj:
Or maybe pack/unpack it to have them as hexits instead of base64?

terrafrost:
First, I apologize for not being able to review this sooner.

Anyway, I see `getBlockLength()` in the 2.0 branch:

https://github.com/phpseclib/phpseclib/blob/2.0/phpseclib/Crypt/Base.php

It's not in the 2.0.0 tag but it is in the 2.0 branch. #827 added it to 1.0 and then 1.0 was merged into 2.0 and 2.0 merged into master.

A quick test demonstrates this:

```php
<?php
include('autoload.php'); 

$loader->addPsr4('phpseclib\\', __DIR__.'/phpseclib2.0'); 
$loader->register(); 

use \phpseclib\Crypt\Rijndael;

$cipher = new Rijndael();
echo $cipher->getBlockLength();
```

> It seems you forgot to release a new version!

Incorrect.

I don't necessarily have a problem releasing a new version soonish but for the 1.0 and 2.0 branches but I'm not going to do so every time I make a commit or add a new function.

There are a number of reasons for this.

1. You're best off using the 2.0 branch with Composer. You can get the latest git version by doing `composer require phpseclib/phpseclib:2.0.x-dev`.

2. I provide most of the support for this library. I develop most of the features for this library. I am the sole maintainer of the documentation. And I'm the only person who does releases and releases take time.

  In theory, a good formal release ought to be properly tested. PHP and phpBB and a lot of bigger projects do release candidates before they do an actual release. For the 1.0 branch I'd also need to update the PEAR channels. For both branches I'd need to research what changes have been made between the new version and the old version so they could be added to a changelog. I'd also need to update the links on the website, etc.

  A good release takes time and every time I do a release that's time I'm not spending doing development, support, or documentation or anything else.

  If someone else did releases and wanted to release more frequently I wouldn't have a problem with that. But no one else does do releases. And I'm not going to give you the ability the releases either. The people who are listed as developers and have commit access are people who've earned my trust. You've interacted with me for a week and in that short amount of time have made unfounded accusations. ie. is the project even being maintained anymore, etc.

  An easy way for you to earn my trust and respect would be to create a repo under your account and redo the documentation. I might not use it but then again I might. Sometimes that's just the risk you gotta take. But right now you haven't done nearly enough to merit the kind of access that would be necessary to do releases.

Anyway, moving on to your changes...

I'm not a big can of `getIVLength`. It's the block length. Having `getBlockLength` and `getIVLength` is just confusing. I mean, I get it - one is returning bits and one is returning bytes. But you wouldn't at all know that from the method names. If you want to be able to toggle between bytes and bits without diving by 8...  I think you'd be better off passing a new parameter into getBlockLength.

I'm also not a big fan of `createIV`. If you want to create an IV do `Random::string($cipher->getBlockLength() >> 3)` or something. That said, I don't feel as strongly about `createIV` as I do about `getIVLength`. If a compelling enough case can be created for its inclusion I might be up for it but right now I'm not.

> Would you like I add two methods for base64 decode/encode IVKey if user needs to transmit it?

If someone wants to base64_encode something can't they just do that themselves? The fact that I'm not a big fan of the `createIV` function not withstanding if we add it for that then do we also add options to return in base64 or hex for `encrypt` and `decrypt` as well? I'm thinking not. Building in PKCS1 padding makes sense because that's a little more involved than just passing the output through an existing function. PBKDF2 (eg. `setPassword`) makes sense for keys for the same reason. But base64_encode'ing? If you want base64_encode'd data you don't need built-in support to do that.

nimasdj:
That's all are fine but it would be better to release a tag to make getBlockLength() available, otherwise these all solutions you said about IV, is impossible.

nimasdj:
and I don't think normal users would even know what >> operator is, let alone applying this operator on 3!

terrafrost:
> and I don't think normal users would even know what >> operator is, let alone applying this operator on 3!

You can do `floor($bits / 8)` as well.

I also think that if it's mentioned in the documentation it should't be a huge issue.

terrafrost:
v1.0.1 and v2.0.1 have been tagged and released.

Issue #756, opened 2015-07-27
1.0.0 / 2.0.0 release this weekend


terrafrost:
I'd like to make releases of 1.0.0 and 2.0.0 this weekend. Maybe we should do a feature freeze until after the weekend?

I don't plan on making a PEAR channel for 2.0. 2.0 will just be installable via composer (or by download).

bantu:
I'd suggest by starting with closing all the open support questions that are stuck (missing information).

bantu:
Do we want to look into PHP7 support before 2.0? I assume we can also do that against master, maybe PHP7 will be in RC soon.

terrafrost:
> I'd suggest by starting with closing all the open support questions that are stuck (missing information).

I'll take a look.

> Do we want to look into PHP7 support before 2.0? I assume we can also do that against master, maybe PHP7 will be in RC soon.

Yah - if that can be done before the weekend that'd be great.

terrafrost:
1.0.0 has been released. I'll be tagging 1.0.0 and 2.0.0 later today. Need to update the README.md and the CHANGELOG.md file before doing that tho.

bantu:
Something went wrong during merge. https://github.com/phpseclib/phpseclib/blob/2.0.0/composer.json#L62 does not belong there.

bantu:
Not a reason to make another retag though.

bantu:
https://github.com/phpseclib/phpseclib/commit/844134df95e42b93ed5506a2d11ca5accda99412 added a phpseclib/phpseclib1.0.0.zip file to the repository history. @terrafrost Please just stick to pull requests.

terrafrost:
That file was deleted in a later commit.

Anyway, I don't have any problems with doing PR's for releases, assuming they're merged in a timely fashion. I mean, if I update the website to announce a new release, the README.md should be updated shortly there after (a few days delay is acceptable imho).

Whomever merges it would need to do `git push origin --tags` (or whatever the name of the remote is) and not just use github.com's built-in merge tool.

bantu:
> That file was deleted in a later commit.

As you probably know, that is not how repositories work. Without rewriting its history, which you should never do for blessed branches anyway, the file is locked permanently and forever in the repository, contributing to the repositories size.

> I mean, if I update the website to announce a new release, the README.md should be updated shortly there after (a few days delay is acceptable imho).

[...]

PR #912, opened 2016-01-04
Add 2.0.0 changelog

### LINKED ISSUE #842: 2.0 Changelog
Is there a changelog available for 2.0 since it has now been merged into master? Also, the website documentation seems to reference 1.0 still.

Thanks

mnapoli:
Fixes #842

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 2.0.0 - 2015-08-04
 
 - Classes were renamed and namespaced ([#243](https://github.com/phpseclib/phpseclib/issues/243))
-- No autoloader is provided anymore (use Composer)
+- The use of an autoloader is now required (e.g. Composer)
 
 ## 1.0.0 - 2015-08-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 2.0.0 - 2015-08-04
+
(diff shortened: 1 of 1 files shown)


terrafrost:
I think what might be better is to put this stuff in the tag descriptions themselves:

https://github.com/phpseclib/phpseclib/releases

The justification being as follows:

https://github.com/phpseclib/phpseclib/pull/624
https://github.com/phpseclib/phpseclib/issues/756#issuecomment-130324051

Of course, no one has really weighed in on that proposal.

And it could just be that I suck at release management too meh. Like maybe what works for other projects doesn't work for me because it's an area I'm deficient in :(

That said...

```
- No autoloader is provided anymore (use Composer)
```

An autoloader was never provided. In the 1.0 branch `include`'s are done on required files at the time that they're required. In 2.0 an autoloader is mostly required. Technically, you could just load all the classes you knew were going to be using in advance, but that's a bit impractical.

What if, instead, that line read like this?:

```
- the use of an autoloader is now required
```

It may not be technically correct but for 99% of use cases it is.

mnapoli:
Good call I've updated the file.

terrafrost:
I squashed your two commits into one and then cherry-picked your commit into the 2.0 branch, which I then merged into the master branch:

https://github.com/phpseclib/phpseclib/blob/2.0/CHANGELOG.md
https://github.com/phpseclib/phpseclib/blob/master/CHANGELOG.md

mnapoli:
:+1:

PR #920, opened 2016-01-09
setKey() validates length.

nimasdj:
setKey() validates length based on Output size and Block size.

### DIFF
--- a/phpseclib/Crypt/Hash.php
+++ b/phpseclib/Crypt/Hash.php
@@ -62,6 +62,15 @@ class Hash
     var $length;
 
     /**
+     * Bit-length of hash BlockSize (Internal HMAC)
+     *
+     * @see self::setHash()
+     * @var int
+     * @access private
+     */
+    var $BlockSize;
+    
+    /**
      * Hash Algorithm
      *
      * @see self::setHash()
@@ -100,6 +109,15 @@ class Hash
      */
     function setKey($key = false)
     {
+        if ($key < $this->length) 
+        {
+            throw new Exception ("Key is too short, it must be at least {$this->length}.");
+        } 
+        elseif ($key > $this->getBlockSize())
+        {
+               $key = substr($key, 0, $this->getBlockSize());
+        }
+        
         $this->key = $key;
     }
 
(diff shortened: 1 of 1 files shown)


terrafrost:
Give me a few days to look at this...  this weekend is just shaping up to be quite shitty. Last weekend the server my companies website is hosted on died and had to be rebuilt from scratch. Through out the week I've been finding issues relating to that and this weekend my CEO is wanting me to launch a big new feature I've been working on Sunday night. And I'm not at all ready for this feature to go live. Maybe I would have been had it not been for the server crash but...  well there was the server crash so I have a lot of shit to do.

And then, on top of that, I need to find a new apartment asap, 'cause my current apartment just denied my gf who's supposed to be moving in with me from being on the lease. And I want to take a look at #914 as well this weekend.

So give me a few days and I'll respond appropriately.

Thanks.

nimasdj:
Ok, if you like this, let me know do you want me to change Exception to BadConfigurationException or to create a new Exception class for that line within setKey method.

terrafrost:
So you're doing two things.

1. You're throwing an exception if the key isn't long enough.
2. You're truncating the key if it is too long.

I don't believe either of these is appropriate.

Quoting [RFC2104](http://tools.ietf.org/html/rfc2104#section-2):

```
   The authentication key K can be of any length up to B, the
   block length of the hash function.  Applications that use keys longer
   than B bytes will first hash the key using H and then use the
   resultant L byte string as the actual key to HMAC. In any case the
   minimal recommended length for K is L bytes (as the hash output
   length).
```

Using a key shorter than the block length may not be recommended but it is supported and may, for purposes of interoperability, sometimes be necessary.

Some test code:

```php
<?php
echo hash_hmac('md5', 'zzz', 'z');
echo "\n";
echo hash_hmac('md5', 'zzz', 'z' . str_repeat("\0", 63));
echo "\n";
echo hash_hmac('md5', 'zzz', 'z' . str_repeat("\0", 64));
echo "\n";
```

nimasdj:
If I understood correctly, condition 1 should be removed and condition 2 should hash the key with the same method to use it as key instead of truncating?

terrafrost:
> If I understood correctly, condition 1 should be removed and condition 2 should hash the key with the same method to use it as key instead of truncating?

Yup! Altho `hash_hmac` already does that

Issue #907, opened 2016-01-03
CBC with openssl


nimasdj:
sorry, I got confused if I use CBC with openssl for AES and if I don't use setIV() will I use warning because of null IV param of openssl_encrypt? Or if I don't call setIV(), phpseclib generates one IV to avoid warning?

terrafrost:
Yes. By default, phpseclib uses an IV of null characters if one isn't explicitly provided to it. Behavior like this isn't exactly without precedent either. Both OpenSSL and mcrypt null pad keys if they're not long enough.

nimasdj:
If I use a random number for IV, where should I store it to get it for decryption? Should I append it with a separator at the end of encoded string? if yes, is it secure as everyone can see that number at the end of that encoded string?

terrafrost:
That's fine. [Quoting wikipedia](http://en.wikipedia.org/wiki/Block_cipher_modes_of_operation#Initialization_vector_.28IV.29),

> An initialization vector has different security requirements than a key, so **the IV usually does not need to be secret**. However, in most cases, **it is important that an initialization vector is never reused under the same key**. For CBC and CFB, reusing an IV leaks some information about the first block of plaintext, and about any common prefix shared by the two messages.

https://en.wikipedia.org/wiki/Block_cipher_mode_of_operation#Initialization_vector_.28IV.29

Joey3000:
> By default, phpseclib uses an IV of null characters if one isn't explicitly provided to it. Behavior like this isn't exactly without precedent either. Both OpenSSL and mcrypt null pad keys if they're not long enough.

The internets seem to be full of warnings to avoid using a constant IV with the CBC mode, while many application devs nevertheless use incorrect library defaults. E.g.:

* https://webcache.googleusercontent.com/search?q=cache:0V0mORvULp0J:https://www.cs.ucsb.edu/~chris/research/doc/ccs13_cryptolint.pdf%2Biv+%3D+0+Security+Cryptographic+Misuse+of+Libraries+-+courses&gbv=1&hl=es&&ct=clnk:

  > Rule 2: Do not use a non-random IV for CBC encryp-
tion. CryptoLint identified 1,932 applications that make
use of constant initialization vectors in CBC mode encryp-
tion.

[...]

PR #926, opened 2016-01-17
Crypt/Base: throw an exception if an IV is required but not defined

terrafrost:
See https://github.com/phpseclib/phpseclib/issues/907#issuecomment-168563911.

Note that IV's and keys are still null padded if they're not long enough. Perhaps it ought not be doing this, either, but, none-the-less, this behavior is consistent with OpenSSL and mcrypt. I'm not opposed to changing that behavior, either, but that'll be another commit for another day.

That said, w.r.t. throwing exceptions if IV's aren't long enough...  the necessity for that is going to be somewhat mode dependent. In CTR mode I'm not sure that kind of behavior is optimal. ie. in CBC mode the IV depends on the encryption result of the last block. But in CTR mode it doesn't. In CTR mode you just increment the IV by one in each block. Ideally, you'd generate the IV by doing something like `random_string(32)` but doing something like `generate_random_number(0, 2 ** 32 - 1)` would be perfectly valid too and w.r.t. numbers there's a non-zero possibility that the number that was chosen would only be 31 bytes long instead of 32. Something to ponder I suppose..

### DIFF
--- a/phpseclib/Crypt/Base.php
+++ b/phpseclib/Crypt/Base.php
@@ -150,7 +150,7 @@ abstract class Base
      * @var string
      * @access private
      */
-    var $iv;
+    var $iv = false;
 
     /**
      * A "sliding" Initialization Vector
@@ -1871,12 +1871,17 @@ abstract class Base
      * after disableContinuousBuffer() or on cipher $engine (re)init
      * ie after setKey() or setIV()
      *
-     * @access public
+     * @access private
      * @internal Could, but not must, extend by the child Crypt_* class
+     * @throws \UnexpectedValueException when an IV is required but not defined
      */
     function _clearBuffers()
     {
         $this->enbuffer = $this->debuffer = array('ciphertext' => '', 'xor' => '', 'pos' => 0, 'enmcrypt_init' => true);
+
+        if ($this->iv === false && !in_array($this->mode, array(self::MODE_STREAM, self::MODE_ECB))) {
+            throw new \UnexpectedValueException('No IV has been defined');
+        }
 
         // mcrypt's handling of invalid's $iv:
         // $this->encryptIV = $this->decryptIV = strlen($this->iv) == $this->block_size ? $this->iv : str_repeat("\0", $this->block_size);
--- a/phpseclib/Crypt/Base.php
+++ b/phpseclib/Crypt/Base.php
@@ -500,10 +500,9 @@ abstract class Base
     }
 
     /**
-     * Sets the initialization vector. (optional)
+     * Sets the initialization vector.
      *
-     * SetIV is not required when self::MODE_ECB (or ie for AES: \phpseclib\Crypt\AES::MODE_ECB) is being used.  If not explicitly set, it'll be assumed
-     * to be all zero's.
+     * setIV() is not required when self::MODE_ECB (or ie for AES: \phpseclib\Crypt\AES::MODE_ECB) is being used.
      *
      * @access public
      * @param string $iv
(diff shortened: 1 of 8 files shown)


Joey3000:
Some feedback:

* In https://github.com/terrafrost/phpseclib/blob/d5a868ed4f07624bf9df76fd877ace54da85c093/phpseclib/Crypt/Base.php#L505-L506 it still says (referring to IV):

  > If not explicitly set, it'll be assumed to be all zero's.

  Also, the same function header still says that the `setIV()` call is optional:

  > Sets the initialization vector. **(optional)**

* The above comment line, as well as https://github.com/terrafrost/phpseclib/blob/d5a868ed4f07624bf9df76fd877ace54da85c093/phpseclib/Crypt/Base.php#L514 don't consider `self::MODE_STREAM`. Is this intended? (Similarly to `self::MODE_ECB`, the IV is not used with it in https://github.com/terrafrost/phpseclib/blob/d5a868ed4f07624bf9df76fd877ace54da85c093/phpseclib/Crypt/Base.php#L705 and other places.)

* Maybe the `mcrypt_encrypt()` implementation behaviour (code) in PHP >= 5.6.0 could be used as a guide with the following, when it's implemented:

  > That said, w.r.t. throwing exceptions if IV's aren't long enough... the necessity for that is going to be somewhat mode dependent. In CTR mode I'm not sure that kind of behavior is optimal. ie. in CBC mode the IV depends on the encryption result of the last block. But in CTR mode it doesn't. In CTR mode you just increment the IV by one in each block. Ideally, you'd generate the IV by doing something like random_string(32) but doing something like generate_random_number(0, 2 ** 32 - 1) would be perfectly valid too and w.r.t. numbers there's a non-zero possibility that the number that was chosen would only be 31 bytes long instead of 32. Something to ponder I suppose..

  From its [manual](https://secure.php.net/manual/en/function.mcrypt-encrypt.php) (emphasis mine):
  > iv
  > Used for the initialization in CBC, CFB, OFB modes, and in some algorithms in STREAM mode. If **the provided IV size is not supported by the chaining mode** or no IV was provided, but the chaining mode requires one, the function will emit a warning and return FALSE.

terrafrost:
>  On the other hand, maybe the non-consideration of `self::MODE_STREAM` is correct here (which makes the provision of IV for it optional, but possible), and the IV needs to be provided in https://github.com/terrafrost/phpseclib/blob/d5a868ed4f07624bf9df76fd877ace54da85c093/phpseclib/Crypt/Base.php#L705, etc.? E.g., from `mcrypt_encrypt()` [manual](https://secure.php.net/manual/en/function.mcrypt-encrypt.php) (emphasis mine):
 > > iv
 > > Used for the initialization in CBC, CFB, OFB modes, **and in some algorithms in STREAM mode**. If the provided IV size is not supported by the chaining mode or no IV was provided, but the chaining mode requires one, the function will emit a warning and return FALSE.

I think your initial thought was the correct one. phpseclib currently only implements one stream cipher - RC4. And in RC4 the only way you can have an "IV" is if you simulate it by adding it to the key as was done in WEP (ie. it's not really an IV).

Maybe there are other stream ciphers that use something that could be called an IV but RC4 isn't one of them.

>  From its [manual](https://secure.php.net/manual/en/function.mcrypt-encrypt.php) (emphasis mine):
>  > iv
>  > Used for the initialization in CBC, CFB, OFB modes, and in some algorithms in STREAM mode. If **the provided IV size is not supported by the chaining mode** or no IV was provided, but the chaining mode requires one, the function will emit a warning and return FALSE.

Nice observation!

Anyway, the changes you recommended have been made - thanks!

Joey3000:
Thanks as well!

PR #927, opened 2016-01-17
TravisCI: dont use openssl with composer for PHP 5.3.3

terrafrost:
See https://travis-ci.org/phpseclib/phpseclib/jobs/102928259 and https://docs.travis-ci.com/user/languages/php#PHP-installation


terrafrost:
I don't know why this isn't working but I suppose it's not essential that it does. Travis CI is already testing against PHP 5.3 so testing against PHP 5.3.3, in particular, seems a little unnecessary, so I've just removed it from .travis.yml.

PR #928, opened 2016-01-18
Cleanup of `use` operator usage

Joey3000:
This does the following:

1. Removes superfluous usage of the `use` operator. Specifically, that operator is not needed to use classes in the same `namespace` (see https://secure.php.net/manual/en/language.namespaces.rules.php). E.g.:

  ```php
  namespace some\namespace;

  use some\namespace\SomeClass; // this is not needed

  class MyClass
  {
      $a = new SomeClass;
  }
  ```

  It would only be needed, if "aliasing" was used (see https://secure.php.net/manual/en/language.namespaces.importing.php), e.g.:

  ```php
  namespace some\namespace;

  use some\other\namespace\SomeClass;
  use some\namespace\SomeClass as MySomeClass; // aliased due to name clash (e.g. introduced at a later time, aliased to not have to adapt the code if aliasing the above class instead)

  class MyClass
  {
      $a = new SomeClass;    // this is some\other\namespace\SomeClass
      $b = new MySomeClass;  // this is some\namespace\SomeClass
  }
  ```

2. While at it, also sorted the `use` operators alphabetically, for a better overview.

### DIFF
--- a/phpseclib/Crypt/AES.php
+++ b/phpseclib/Crypt/AES.php
@@ -49,8 +49,6 @@
 
 namespace phpseclib\Crypt;
 
-use phpseclib\Crypt\Rijndael;
-
 /**
  * Pure-PHP implementation of AES.
  *
--- a/phpseclib/Crypt/Base.php
+++ b/phpseclib/Crypt/Base.php
@@ -36,8 +36,6 @@
 
 namespace phpseclib\Crypt;
 
-use phpseclib\Crypt\Hash;
-
 /**
  * Base Class for all \phpseclib\Crypt\* cipher classes
  *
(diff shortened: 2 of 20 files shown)


terrafrost:
Looks good! Do you want to do a commit like this for the 2.0 branch too? You could merge that PR into this branch using `git merge -X ours use_cleanup-2.0` as well. Or I can do the merging myself too lol. But I think they'd definitely be good for the 2.0 branch as well :)

Joey3000:
> Do you want to do a commit like this for the 2.0 branch too?

Done: https://github.com/phpseclib/phpseclib/pull/931

PR #929, opened 2016-01-18
Update license year range to 2016

pra85:


terrafrost:
Merged: https://github.com/phpseclib/phpseclib/blob/master/LICENSE

Thanks!

PR #934, opened 2016-01-25
Random: fix for issues with serialize()

### LINKED ISSUE #932: PHP fatal error when generating a random string
I have a user, for whom this section in crypt_random_string() / Random::String() (depending on your version) in Crypt/Random.php is causing a PHP fatal error:

```
$v = $seed = $_SESSION['seed'] = pack('H*', sha1(
                serialize($_SERVER) .
                serialize($_POST) .
                serialize($_GET) .
                serialize($_COOKIE) .
                serialize($GLOBALS) .
                serialize($_SESSION) .
                serialize($_OLD_SESSION)
            ));
```
The fatal error is "Fatal error: Call to private method Vc_Manager::__sleep()", and it's coming from the attempt to serialize $GLOBALS.

The object Vc_Manager appears to be from the popular WordPress plugin "Visual Bakery Composer". Looking at https://secure.php.net/manual/en/language.oop5.magic.php#object.sleep, it appears that that's a bug in that plugin - that __sleep() ought to be a public, not a private method. However... it's a bit unfortunate to be triggering PHP fatal errors in other plugins simply in an attempt to get some random data. serialize() can cause other code to be run, which doesn't seem an optimal thing to be doing at this point in the code.

I'd suggest that you have a method which first parses all the things in the to-be-serialized arrays, and removes all the objects, prior to serialisation.

e.g. instead of calling serialize($GLOBALS), call serialize($this->remove_objects($GLOBALS)), with something like this (untested) method:

```
function remove_objects($from) {
  foreach ($from as $k => $v) {
    if (is_object($v)) {
      unset($from[$k]);
    } elseif (is_array($v)) {
      $from[$k] = $this->remove_objects($v);
    }
  }
  return $from;
}
```

In case it's necessary to say, the above snippet is public domain, full permission given for any/all use in anything.

terrafrost:
Fixes #932.

(latest branch: https://github.com/terrafrost/phpseclib/tree/random-serialize-3)

### DIFF
--- a/phpseclib/Crypt/Random.php
+++ b/phpseclib/Crypt/Random.php
@@ -148,13 +148,13 @@ if (!function_exists('crypt_random_string')) {
             session_start();
 
             $v = $seed = $_SESSION['seed'] = pack('H*', sha1(
-                serialize($_SERVER) .
-                serialize($_POST) .
-                serialize($_GET) .
-                serialize($_COOKIE) .
-                serialize($GLOBALS) .
-                serialize($_SESSION) .
-                serialize($_OLD_SESSION)
+                phpseclib_safe_serialize($_SERVER) .
+                phpseclib_safe_serialize($_POST) .
+                phpseclib_safe_serialize($_GET) .
+                phpseclib_safe_serialize($_COOKIE) .
+                phpseclib_safe_serialize($GLOBALS) .
+                phpseclib_safe_serialize($_SESSION) .
+                phpseclib_safe_serialize($_OLD_SESSION)
             ));
             if (!isset($_SESSION['count'])) {
                 $_SESSION['count'] = 0;
@@ -257,6 +257,44 @@ if (!function_exists('crypt_random_string')) {
             $result.= $r;
         }
         return substr($result, 0, $length);
+    }
+}
+
+if (!function_exists('phpseclib_safe_serialize')) {
+    /**
+     * Safely serialize variables
+     *
+     * If a class has a private __sleep() method it'll give a fatal error on PHP 5.2 and earlier.
+     * PHP 5.3 will emit a warning.
+     *
+     * @param mixed $arr
+     * @param array $refs optional
+     * @access public
+     */
+    function phpseclib_safe_serialize($arr, $refs = array())
+    {
+        if (is_object($arr)) {
+            return '';
+        }
+        if (!is_array($arr)) {
+            return serialize($arr);
+        }
+        $safearr = array();
+        foreach (array_keys($arr) as $key) {
+            if (is_object($arr[$key])) {
+                continue;
+            }
+            // without this recursive arrays (eg. $a = []; $a[] = &$a;) would result in an
+            // "Allowed memory size of ... bytes exhausted" Fatal error
+            foreach ($refs as $ref) {
+                if ($ref === $arr[$key]) {
+                    continue 2;
+                }
+            }
+            $refs[] = &$arr[$key];
+            $safearr[$key] = is_array($arr[$key]) ? phpseclib_safe_serialize($arr[$key], $refs) : $arr[$key];
+        }
+        return serialize($safearr);
     }
 }
 


phpseclib/Crypt/Random.php
```diff
@@ -147,15 +147,17 @@ function crypt_random_string($length)
...
-            $v = $seed = $_SESSION['seed'] = pack('H*', sha1(
-                serialize($_SERVER) .
-                serialize($_POST) .
-                serialize($_GET) .
-                serialize($_COOKIE) .
-                serialize($GLOBALS) .
-                serialize($_SESSION) .
-                serialize($_OLD_SESSION)
-            ));
+            if (function_exists('json_encode')) {
```
Joey3000: Is this check (still) necessary? It seems that `json_encode()` is not used (any longer).

phpseclib/Crypt/Random.php
```diff
@@ -260,6 +262,32 @@ function crypt_random_string($length)
...
+        }
+        if (!is_array($arr)) {
+            return serialize($arr);
+        }
+        $safearr = array();
+        foreach (array_keys($arr) as $key) {
+            if (is_object($arr[$key])) {
+                continue;
+            }
+            $safearr[$key] = is_array($arr[$key]) ? phpseclib_safe_serialize($arr[$key]) : $arr[$key];
```
Joey3000: `$GLOBALS` seems to be recursive. So that it seems that some sort recursion prevention would be needed. E.g., I tried this function with `echo phpseclib_safe_serialize($GLOBALS);` on http://sandbox.onlinephpfunctions.com/ and:
* For PHP 4 - 5.2 it doesn't return anything
* For PHP 5.3 - 7.0 it shows:

  ```
  <b>Fatal error</b>:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes) in <b>[...][...]</b> on line <b>14</b><br />
  ```

This might be dependent on php.ini configuration.

A simple solution of changing line 282 to `if (is_object($arr[$key]) || ($arr[$key] === $GLOBALS)) {` seems to work on higher PHP versions, but not on lower ones. So that maybe a recursion depth limitation could be used as a universal solution? (Like, e.g., in https://secure.php.net/manual/en/function.var-dump.php#76072 or https://secure.php.net/manual/en/function.print-r.php#78851)

terrafrost:
Good catches! I've updated the code - thanks!

phpseclib/Crypt/Random.php
```diff
@@ -260,6 +260,44 @@ function crypt_random_string($length)
...
+        }
+        $safearr = array();
+        foreach (array_keys($arr) as $key) {
+            if (is_object($arr[$key])) {
+                continue;
+            }
+            // without this recursive arrays (eg. $a = []; $a[] = &$a;) would result in an
+            // "Allowed memory size of ... bytes exhausted" Fatal error
+            foreach ($refs as $ref) {
+                if ($ref === $arr[$key]) {
```
Joey3000: This comparison does not seem to work with older PHP versions. When tested using:

```
[copy-paste of the phpseclib_safe_serialize() definition]
echo phpseclib_safe_serialize($GLOBALS);
```

on http://sandbox.onlinephpfunctions.com/, it outputs following error:

```
<br />
<b>Fatal error</b>:  Nesting level too deep - recursive dependency? in <b>[...][...]</b> on line <b>19</b><br />
```

on the following PHP versions:
5.4.4
5.3.14
5.2.17
5.1.6
5.0.5
4.4.9

Maybe a recursion depth limitation could be used instead? (Like, e.g., in https://secure.php.net/manual/en/function.var-dump.php#76072 or https://secure.php.net/manual/en/function.print-r.php#78851 .)

Note: The behaviour changes in PHP 5.3.15 and 5.4.5, where the error no longer occurs (including later versions). Both were released on 19 Jul 2012, according to https://secure.php.net/releases/.

terrafrost: I'm just not a big fan of recursion depth limits.

Like what if I arbitrarily had a recursion depth limit 5 for File/ASN1.php? Some things go deeper than that so what am I to do in that situation? At that point File/X509.php might not even be able to parse the output and even if it could stuff might be missing and people would wonder why.

Having to set an arbitrary recursion depth limit is just asking for problems imho. It's just a non-intuitive setting that no one is ever going to think of to set. And even if they did they wouldn't know what'd be an appropriate value to set it to. It'd just be a method you'd need to call with a random parameter and it'd just be one extra step that people could mess up on.

I suppose in this particular case it's not as important. It reduces the initial entropy but they'll still be entropy.

It looks like the issue on older verions of PHP is the `===`:

```php
<?php
$a = array();
$a[] = &$a;

echo $a === $a[0] ? 'match'  : 'no match';
```

That causes the error.

One thing that does seem to work...  `debug_zval_dump`. According to http://php.net/debug-zval-dump it's been available as early as PHP 4.2.

It works with recursive stuff:

```php
<?php
$a = array();
$a[] = &$a;

debug_zval_dump($a);
```

It also works with private __sleep()'s:

```php
<?php
class test {
  var $var = 'zz';

  private function __sleep() {
    return array('var');
  }
}

$a = new test();

debug_zval_dump($a);
```

The one disadvantage to it is that it does dump to stdout. Output buffering (eg. `ob_start()` / `ob_flush()`) would need to be used to capture the output to a string. And that poses problems for people who might already be utilizing output buffering.

Another possibility that occured to me...

```php
<?php
$a = array();
$a[] = &$a;

$a = (object) $a;
```
`var_dump` seems to suggest that that works but `$a->0` isn't a valid variable name. I can't even hit it with reflection (not that that'd work on PHP 4):


```php
<?php
$a = array();
$a[] = &$a;

$a = (object) $a;

$reflectionClass = new ReflectionObject($a);
echo $a === $reflectionClass->getProperty('0')->getValue() ? 'matches' : 'does not match';
```

So I guess the recursion depth is the best approach meh. And if we go that route the next question would, I guess, be this: what would be a good recursion depth? If we did 1 that'd make for much simpler serialization. eg.

```php
    function phpseclib_safe_serialize($arr)
    {
        if (is_object($arr)) {
            return '';
        }
        if (!is_array($arr)) {
            return serialize($arr);
        }
        foreach (array_keys($arr) as $key) {
            if (is_object($arr[$key]) || is_array($arr[$key])) {
                unset($arr[$key]);
            }
        }
        return serialize($safearr);
    }
```

terrafrost:
I'm just not a big fan of recursion depth limits.

Like what if I arbitrarily had a recursion depth limit 5 for File/ASN1.php? Some things go deeper than that so what am I to do in that situation? At that point File/X509.php might not even be able to parse the output and even if it could stuff might be missing and people would wonder why.

Having to set an arbitrary recursion depth limit is just asking for problems imho. It's just a non-intuitive setting that no one is ever going to think of to set. And even if they did they wouldn't know what'd be an appropriate value to set it to. It'd just be a method you'd need to call with a random parameter and it'd just be one extra step that people could mess up on.

I suppose in this particular case it's not as important. It reduces the initial entropy but they'll still be entropy.

It looks like the issue on older verions of PHP is the `===`:

```php
<?php
$a = array();
$a[] = &$a;

echo $a === $a[0] ? 'match'  : 'no match';
```

That causes the error.

One thing that does seem to work...  `debug_zval_dump`. According to http://php.net/debug-zval-dump it's been available as early as PHP 4.2.

It works with recursive stuff:

```php
<?php
$a = array();
$a[] = &$a;

debug_zval_dump($a);
```

It also works with private __sleep()'s:

```php
<?php
class test {
  var $var = 'zz';

  private function __sleep() {
    return array('var');
  }
}

$a = new test();

debug_zval_dump($a);
```

The one disadvantage to it is that it does dump to stdout. Output buffering (eg. `ob_start()` / `ob_flush()`) would need to be used to capture the output to a string. And that poses problems for people who might already be utilizing output buffering.

Another possibility that occured to me...

```php
<?php
$a = array();
$a[] = &$a;

$a = (object) $a;
```
`var_dump` seems to suggest that that works but `$a->0` isn't a valid variable name. I can't even hit it with reflection (not that that'd work on PHP 4):


```php
<?php
$a = array();
$a[] = &$a;

$a = (object) $a;

$reflectionClass = new ReflectionObject($a);
echo $a === $reflectionClass->getProperty('0')->getValue() ? 'matches' : 'does not match';
```

So I guess the recursion depth is the best approach meh. And if we go that route the next question would, I guess, be this: what would be a good recursion depth? If we did 1 that'd make for much simpler serialization. eg.

```php
    function phpseclib_safe_serialize($arr)
    {
        if (is_object($arr)) {
            return '';
        }
        if (!is_array($arr)) {
            return serialize($arr);
        }
        foreach (array_keys($arr) as $key) {
            if (is_object($arr[$key]) || is_array($arr[$key])) {
                unset($arr[$key]);
            }
        }
        return serialize($safearr);
    }
```

terrafrost:
Didn't mean to close this - I hit the wrong button lol. It's been re-opened.

phpseclib/Crypt/Random.php
Joey3000: Here is an alternative proposal: https://github.com/terrafrost/phpseclib/pull/4 (can be merged into this PR branch - I'm just playing with GitHub :) ). What do you think of it? It seems to work on both, older and newer PHP versions (tested on http://sandbox.onlinephpfunctions.com/).

phpseclib/Crypt/Random.php
```diff
@@ -148,13 +148,13 @@ function crypt_random_string($length)
...
-                serialize($GLOBALS) .
-                serialize($_SESSION) .
-                serialize($_OLD_SESSION)
+                phpseclib_safe_serialize($_SERVER) .
+                phpseclib_safe_serialize($_POST) .
+                phpseclib_safe_serialize($_GET) .
+                phpseclib_safe_serialize($_COOKIE) .
+                phpseclib_safe_serialize($GLOBALS) .
+                phpseclib_safe_serialize($_SESSION) .
+                phpseclib_safe_serialize($_OLD_SESSION)
```
Joey3000: The following:
```php
echo phpseclib_safe_serialize($_SESSION);
echo phpseclib_safe_serialize($_OLD_SESSION);
```

gives the following notices on PHP versions 4.4.9 to 7.0.3 on http://sandbox.onlinephpfunctions.com/ (possibly due to PHP.INI configuration?):
```
<br />
<b>Notice</b>:  Undefined variable: _SESSION in <b>[...][...]</b> on line <b>32</b><br />
N;---;<br />
<b>Notice</b>:  Undefined variable: _OLD_SESSION in <b>[...][...]</b> on line <b>33</b><br />
```

Maybe an existence check using `isset()` is needed before using the variables with `phpseclib_safe_serialize()`?

terrafrost: Crypt/Random.php does `session_start()` before trying to call `$_SESSION`.  Normally that'd create `$_SESSION` but not on http://sandbox.onlinephpfunctions.com/ and that's because of the php.ini thing you mentioned. eg.
```php
<?php
session_start();
serialize($_SESSION);
```

That yields a `Warning: session_start() has been disabled for security reasons` error:

http://sandbox.onlinephpfunctions.com/code/15254dd8e75ef36649a2afae66470662d8782c64

So they probably have that function mentioned in `disable_functions`:

http://php.net/manual/en/ini.core.php#ini.disable-functions

I guess we could still do `isset()` but idk...  issues arising from disabled functions are not issues that we should be overly concerned with imho. You disable functions at your own peril.

phpseclib/Crypt/Random.php
terrafrost:
```php
$GLOBALS['a'] = &$GLOBALS;

phpseclib_safe_serialize($GLOBALS, 'GLOBALS');
```

That causes the same "Fatal error: Allowed memory size" error. I mean, on one hand, people ought not be doing that, but then again, people ought not be making `__sleep` private either.

In light of this, what about this?:

https://github.com/terrafrost/phpseclib/commit/086bd845fe236213bb9f6860dd6c2c879974fb77

Joey3000: > In light of this, what about this?:

> terrafrost@086bd84

It looks good. Here is an incremental improvement proposal: https://github.com/terrafrost/phpseclib/pull/5 :)

phpseclib/Crypt/Random.php
Joey3000: > I guess we could still do isset() but idk... issues arising from disabled functions are not issues that we should be overly concerned with imho. You disable functions at your own peril.

True, but disabling `session_start()` seems to be a valid configuration option. (Unlike, e.g., a patch to the PHP interpreter itself.)

To the other Superglobals:
* `$_OLD_SESSION`: Does this one exist? It's not mentioned in https://secure.php.net/manual/en/language.variables.superglobals.php
* `$GLOBALS`: This one seems to actually always exist.

* Others: It seems that they may be non-existent, too. From https://secure.php.net/manual/en/language.variables.superglobals.php:
  > Note: Variable availability
  > By default, all of the superglobals are available but there are directives that affect this availability. For further information, refer to the documentation for [variables_order](https://php.net/manual/en/ini.core.php#ini.variables-order). 

Those are just some thoughts anyway.

phpseclib/Crypt/Random.php
terrafrost: Found another one that fails with both your updated version and with mine:

```php
$z = [];
$z[] = [&$z];
```

That causes a "Nesting level too deep - recursive dependency?" Fatal error even on PHP 5.5 when you do `===`. eg. `$z === $z[0]`.

Here's another version:

https://github.com/terrafrost/phpseclib/commit/102d53bd275f5c08e7042eaac5fb74f73eded62a

It's based on http://stackoverflow.com/a/9293146/569976

terrafrost:
> True, but disabling `session_start()` seems to be a valid configuration option. (Unlike, e.g., a patch to the PHP interpreter itself.)

It's just as valid a configuration option as disabling `fsockopen` is (which'll break SSH1 / SSH2 / SFTP), as disabling `base64_decode` is (which'll break a lot of RSA.php file formats and X509.php), etc. There are lots of functions which, if disabled, will break phpseclib, and checking for all of them isn't practical.

> `$_OLD_SESSION`: Does this one exist? It's not mentioned in https://secure.php.net/manual/en/language.variables.superglobals.php

It's defined, earlier, with this line:

```php
            $_OLD_SESSION = isset($_SESSION) ? $_SESSION : false;
```

The idea being to save it if it's already been defined by a previous `session_start()`. `crypt_random_string()` doesn't do a previous `session_start()` but it's possible the calling program could.

> Others: It seems that they may be non-existent, too. From https://secure.php.net/manual/en/language.variables.superglobals.php:

That's a good point. I can add an isset check for everything else. If `$_SESSION` isn't available the whole

phpseclib/Crypt/Random.php
Joey3000: > Found another one that fails with both your updated version and with mine:

Good find!

> Here's another version:

> terrafrost@102d53b

> It's based on http://stackoverflow.com/a/9293146/569976

I added a few minor comments to it. But, more importantly, it also causes a "Fatal error:  Allowed memory size of ... bytes exhausted" if tested with the following on http://sandbox.onlinephpfunctions.com/ on the higher PHP versions, and doesn't return anything on the lower ones (note: the two `$z` lines need to be commented out for the lower versions):

```php
$a['key1'] = 'value1';
$a[] = &$a;

$b[] = &$a;
$b[] = &$b;

$c[] = &$b;

$z = [];
$z[] = [&$z];

echo phpseclib_safe_serialize($GLOBALS);
```

Here is a try to address all of that, and to also make the implementation smaller by bringing it closer to the above Stackoverflow example:
https://github.com/phpseclib/phpseclib/commit/047f2617f164f8f95d1c576994008719af68fecb

P.S.: This time just a commit as well (to stop the PR spam) - please let me know if you'd prefer a PR.

terrafrost: I think the key part of your commit was the `!isset($arr[$key]['__phpseclib_marker'])`. I don't think removing the `$unset` stuff was necessary and, thinking about it, I don't think `$key !== '__phpseclib_marker' && ` is necessary either.

Anyway, I've merged your PR into mine and then made the following additional commit:

https://github.com/terrafrost/phpseclib/commit/27f659f3d9d76754892e7a2dee0562187331e6d1

Joey3000: Thanks!

> I don't think `$key !== '__phpseclib_marker' &&` is necessary

* Without it, each '__phpseclib_marker' key is also serialized. E.g., the test variables in https://github.com/phpseclib/phpseclib/pull/934/files#r52842344 above result in the following output (on PHP 7.0.3):

  ```
  a:9:{s:4:"_GET";s:36:"a:1:{s:18:"__phpseclib_marker";b:1;}";s:5:"_POST";s:36:"a:1:
  {s:18:"__phpseclib_marker";b:1;}";s:7:"_COOKIE";s:36:"a:1:
  {s:18:"__phpseclib_marker";b:1;}";s:6:"_FILES";s:36:"a:1:
  {s:18:"__phpseclib_marker";b:1;}";s:1:"a";s:60:"a:2:
  {s:4:"key1";s:6:"value1";s:18:"__phpseclib_marker";b:1;}";s:1:"b";s:108:"a:2:{i:0;s:60:"a:2:
  {s:4:"key1";s:6:"value1";s:18:"__phpseclib_marker";b:1;}";s:18:"__phpseclib_marker";b:1;}
  ";s:1:"c";s:157:"a:2:{i:0;s:108:"a:2:{i:0;s:60:"a:2:
  {s:4:"key1";s:6:"value1";s:18:"__phpseclib_marker";b:1;}";s:18:"__phpseclib_marker";b:1;}
  ";s:18:"__phpseclib_marker";b:1;}";s:1:"z";s:84:"a:2:{i:0;s:36:"a:1:
  {s:18:"__phpseclib_marker";b:1;}";s:18:"__phpseclib_marker";b:1;}";s:18:"__phpseclib_marker";b:1;}
  ```

  As opposed to the following otherwise:

  ```
  a:8:{s:4:"_GET";s:6:"a:0:{}";s:5:"_POST";s:6:"a:0:{}";s:7:"_COOKIE";s:6:"a:0:{}";s:6:"_FILES";s:6:"a:0:
  {}";s:1:"a";s:30:"a:1:{s:4:"key1";s:6:"value1";}";s:1:"b";s:48:"a:1:{i:0;s:30:"a:1:
  {s:4:"key1";s:6:"value1";}";}";s:1:"c";s:66:"a:1:{i:0;s:48:"a:1:{i:0;s:30:"a:1:
  {s:4:"key1";s:6:"value1";}";}";}";s:1:"z";s:23:"a:1:{i:0;s:6:"a:0:{}";}";}
  ```

  So, roughly the double RAM consumption on that non-representative test. (And a similar proportion on PHP 4.4.9.) I don't know how much impact that has in a typical PHP environment though, with respect to the number of existing global variables containing arrays (don't need to be circular).

* Also, the comment on the following line would need to be removed then as well: https://github.com/terrafrost/phpseclib/pull/6/files#diff-b08ee4e00815fdb38838fb5016bd2ee4R285.

Joey3000:
> That's a good point. I can add an isset check for everything else. If $_SESSION isn't available then there's not even any pretense of crypt_random_string() being secure

Thanks for the update and clarification!

phpseclib/Crypt/Random.php
terrafrost: Good point. I've updated the code and rebased the branch:

https://github.com/terrafrost/phpseclib/tree/random-serialize-3

terrafrost: Actually, I found an issue with the last commit. I made another to fix it:

https://github.com/terrafrost/phpseclib/commit/d1ef2210af4c0452d2013fb3aae6c62389b944e8

Basically, if `$arr[$key]` isn't an array the `isset($arr[$key]['__phpseclib_marker'])` call will fail.

bantu:
@terrafrost Don't you think it is time to just require proper randomness to be available from the OS (and otherwise just fail fatally)? The amount of time wasted on this PR is ridiculous.

phpseclib/Crypt/Random.php
Joey3000: Yeah, I found that, too: It gave the following error on an object:
> Fatal error:  Uncaught Error: Cannot use object of type test as array in [...][...]:16

There seem to be a few issues with your update, though - I added comments to it.

I have an alternative approach: https://github.com/phpseclib/phpseclib/commit/062aa41ba002684101fe5dd526c161c9afa6085d. The change is:
* Instead of adding a `if (!is_object($arr))` before the recursion, it moves `if (isset($arr['__phpseclib_marker']))` out to the beginning of the recursion. The latter needs to be done **after** the former, instead of the same line, because it needs to already be sure that it's not an object at this point, because (from https://secure.php.net/manual/en/language.operators.precedence.php):

  > PHP does not (in the general case) specify in which order an expression is evaluated and code that assumes a specific order of evaluation should be avoided, because the behavior can change between versions of PHP or depending on the surrounding code.

Here is the ultimate test, which get passed on all versions (`private` and `$z` need to be commented out on lower versions):
```php
class test {
    private function __sleep() {
        return array();
    }
}

$aobj = new test();

$scalar_value = 'whatever';

$a['key1'] = 'value1';
$a[] = &$a;

$b[] = &$a;
$b[] = &$b;

$c[] = &$b;

$z = [];
$z[] = [&$z];

echo phpseclib_safe_serialize($GLOBALS);
```

What do you think?

Joey3000:
@bantu You mean we won't get paid for this engineering exercise? :D

terrafrost:
> The amount of time wasted on this PR is ridiculous.

I actually kinda agree. I've half been tempted just to say "so you have code that has a private `__sleep` method? That's not a problem phpseclib can be reasonably expected to fix - that's a bug in your code!". But I think with my latest commit this should be resolved. Of course, I've had that thought that for the past few commits lol meh.

phpseclib/Crypt/Random.php
Joey3000: And just to be sure, although there weren't any errors reported when `if (isset($arr['__phpseclib_marker']))` was done on a scalar: https://github.com/phpseclib/phpseclib/commit/94b669d7ed4251519cbfcd15c3b1da6d790dcf83

terrafrost: I merged your changes into 1.0 / 2.0 / master:

https://github.com/phpseclib/phpseclib/commits/master

Thanks!

PR #936, opened 2016-01-26
IPv6 support for X509v3 SubjectAltNames

metaclassing:
Switched out pack/unpack and ip2long/long2ip with inet_pton/ntop for ipv6 encoding support in x509v3 subject alternative names. The functions should be in PHP5.1 (or 5.3 for Windows) or later.

I am attaching my before and after certificates generated with this patch for comparison.
[ipv6.before.crt.txt](https://github.com/phpseclib/phpseclib/files/105873/ipv6.before.crt.txt)
[ipv6.after.crt.txt](https://github.com/phpseclib/phpseclib/files/105872/ipv6.after.crt.txt)

Please let me know if this is acceptable, or if there are changes / alternative approaches you would prefer. Thanks.

### DIFF
--- a/phpseclib/File/X509.php
+++ b/phpseclib/File/X509.php
@@ -2198,9 +2198,7 @@ class X509
      */
     function _decodeIP($ip)
     {
-        $ip = base64_decode($ip);
-        list(, $ip) = unpack('N', $ip);
-        return long2ip($ip);
+        return inet_ntop(base64_decode($ip));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


bantu:
Unit tests for the existing IPv4 behaviour and the newly added IPv6 functionality would be nice. Or is that already covered somewhere?

metaclassing:
I added a quick unit test to decode both IPv4 and IPv6 elements in subject alt names.

Other than the x509v3 subjectaltname I don't know of another location where IP address information must be encoded as binary "in network byte order", other locations like subject encode them as text.

terrafrost:
I apologize for the delay! This has been merged - thanks!!

PR #942, opened 2016-02-07
Crypt: become a lot less tolerant of bad parameters

terrafrost:
phpseclib 1.0 and 2.0 are fairly forgiving when it comes to the parameters it takes for the various symmetric ciphers. If your IV is too short it'll be null padded and if it's too long it'll be truncated. Same thing for the key and for the mode as well.

With this PR the master branch (aka 3.0) will not be forgiving at all. If the key or IV that are provided are not of a valid length a LengthException will be thrown. The ciphers also no longer have a default mode at all - the mode will need to be explicitly defined and your mode of choice isn't valid an Exception will be thrown there as well.

The idea behind this being that the symmetric ciphers phpseclib provides are cryptographic primitives. If you need to use them for interoperability purposes you're free to do so (phpseclib does), however, if you only need to be interoperable with yourself you'd be better off using something like the Cryptographic Message Syntax or PGP or libsodium or something. It is my intention to add more general purpose crypto stuff to phpseclib (CMS / PGP / etc), however, that's a ways away.

Superficially setKeyLength() seems like it could be removed but PKCS.php demonstrates a use case where it's removal would be detrimental. Also, setPassword() could benefit from it.

Superficially one might also think that setBlockLength() could be removed - that the block length could be auto-detected from the setIV - but ECB mode doesn't use an IV so that wouldn't work. So if you do want to use a block length of, say, 256 bits, with Rijndael, doing setIV(crypt_random_string(32)) isn't what you'd need to do but rather setBlockLength(256); setIV(crypt_random_string(32)) (or crypt_random_string(getBlockLength()).

I had to add usesIV() since RC4 doesn't use an IV. At first I was thinking that since that's the only stream cipher phpseclib currently implements that no stream ciphers used IV's but I guess that's not true. If it had been maybe making StreamCipher and BlockCipher classes that could be used or implemented could have been worthwhile but I guess for this particular use-case it's not of any great benefit. Maybe it can be done at a future date however.


phpseclib/Crypt/AES.php
```diff
@@ -67,31 +67,32 @@ class AES extends Rijndael
      *
      * @see \phpseclib\Crypt\Rijndael::setBlockLength()
      * @access public
-     * @param int $length
```
GrahamCampbell: i assume removing this param annotation was unintentional?

phpseclib/Crypt/AES.php
```diff
@@ -67,31 +67,32 @@ class AES extends Rijndael
...
-            case 160:
-                $length = 192;
+            case 128:
+            case 192:
+            case 256:
                 break;
-            case 224:
-                $length = 256;
+            default:
+                throw new \LengthException('Key of size ' . strlen($key) . ' not supported by this algorithm. Only keys of sizes 16, 24 or 32 supported');
```
Joey3000: * The `$key` variable doesn't exist in this function. Should it be `$length`?
* Should the following be rather "Key **length** of"? (For consistency with the function name.)

  > Key of **size**

* Should the following be rather "Only key **lengths of 128, 192 or 256** supported."? (For consistency with the function name, and in bits.)

  > Only keys of **sizes 16, 24 or 32** supported

Note: This could be a copy-paste error from https://github.com/phpseclib/phpseclib/pull/942/files#diff-05acee93a75378713d649bc1b9601be2R119.

phpseclib/Crypt/Twofish.php
```diff
@@ -380,21 +396,43 @@ class Twofish extends Base
...
-            case $length <= 192:
-                $this->key_length = 24;
+        switch ($length) {
+            case 128:
+            case 192:
+            case 256:
                 break;
             default:
-                $this->key_length = 32;
+                throw new \LengthException('Key of size ' . strlen($key) . ' not supported by this algorithm. Only keys of sizes 16, 24 or 32 supported');
```
Joey3000: See https://github.com/phpseclib/phpseclib/pull/942/files#r52751087 (same applies here).

terrafrost:
Good catches GrahamCampbell and Joey3000! The suggested changes have been implemented - thanks!

phpseclib/Crypt/AES.php
```diff
@@ -68,30 +68,32 @@ class AES extends Rijndael
...
-            case 160:
-                $length = 192;
+            case 128:
+            case 192:
+            case 256:
                 break;
-            case 224:
-                $length = 256;
+            default:
+                throw new \LengthException('Key of size ' . $length . ' not supported by this algorithm. Only keys of sizes 16, 24 or 32 supported');
```
Joey3000: Should the following be "**128, 192 or 256**"?

  > 16, 24 or 32

phpseclib/Crypt/Twofish.php
```diff
@@ -380,21 +396,43 @@ class Twofish extends Base
...
-            case $length <= 192:
-                $this->key_length = 24;
+        switch ($length) {
+            case 128:
+            case 192:
+            case 256:
                 break;
             default:
-                $this->key_length = 32;
+                throw new \LengthException('Key of size ' . $length . ' not supported by this algorithm. Only keys of sizes 16, 24 or 32 supported');
```
Joey3000: See https://github.com/phpseclib/phpseclib/pull/942/files#r52847391 (same applies here).

terrafrost:
The suggested changes have been made and the branch rebased!

PR #946, opened 2016-02-11
Remove the dependency on phpinfo()

raggi:
The code now looks for a supported openssl version with at least the heartbleed patch level.

There may be contention as to whether this solves the original XAMPP issues. I don't have a good way of testing this. The problems for which this was introduced existed back in 2013. Unless a user system is horribly insecure, header<->library mismatches should no longer result in behavior broken as the original kludge was trying to avoid.

The core goal here is to remove the reliance on phpinfo. If other methods are preferred, I'm all ears.

### DIFF
--- a/phpseclib/Crypt/RSA.php
+++ b/phpseclib/Crypt/RSA.php
@@ -416,37 +416,17 @@ class RSA
                     break;
                 case extension_loaded('openssl') && file_exists(self::$configFile):
                     // some versions of XAMPP have mismatched versions of OpenSSL which causes it not to work
-                    ob_start();
-                    @phpinfo();
-                    $content = ob_get_contents();
-                    ob_end_clean();
-
-                    preg_match_all('#OpenSSL (Header|Library) Version(.*)#im', $content, $matches);
-
-                    $versions = array();
-                    if (!empty($matches[1])) {
-                        for ($i = 0; $i < count($matches[1]); $i++) {
-                            $fullVersion = trim(str_replace('=>', '', strip_tags($matches[2][$i])));
-
-                            // Remove letter part in OpenSSL version
-                            if (!preg_match('/(\d+\.\d+\.\d+)/i', $fullVersion, $m)) {
-                                $versions[$matches[1][$i]] = $fullVersion;
-                            } else {
-                                $versions[$matches[1][$i]] = $m[0];
-                            }
-                        }
-                    }
-
-                    // it doesn't appear that OpenSSL versions were reported upon until PHP 5.3+
-                    switch (true) {
-                        case !isset($versions['Header']):
-                        case !isset($versions['Library']):
-                        case $versions['Header'] == $versions['Library']:
-                            define('CRYPT_RSA_MODE', self::MODE_OPENSSL);
-                            break;
-                        default:
-                            define('CRYPT_RSA_MODE', self::MODE_INTERNAL);
-                            define('MATH_BIGINTEGER_OPENSSL_DISABLE', true);
+                    // This used to rely on phpinfo, but that is disabled in
+                    // common secure environments. Instead, assume that if the
+                    // library version is high enough we'll be fine (which we
+                    // will).  The first 1.0.1 heartbleed fixed level is picked
+                    // here, as users shouldn't be running lower anyway. 1.0.0
+                    // and below are no longer supported.
+                    if (OPENSSL_VERSION_NUMBER >= 0x1000107f) {
+                        define('CRYPT_RSA_MODE', self::MODE_OPENSSL);
+                    } else {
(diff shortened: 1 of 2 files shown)


terrafrost:
I'm not really a big fan of this. Heartbleed affects the SSL protocol and phpseclib isn't using OpenSSL to implement SSL. It's using OpenSSL to speed up modular exponentiation and it's using it to generate prime numbers faster. ie. the heartbleed vulnerability isn't really relevant for what phpseclib is using OpenSSL for.

That said, I can empathize with the concern over `phpinfo()` being disabled. I mean, on one hand, I question just how much phpseclib ought to accommodate PHP installs that have disabled arbitrary functions. `fsockopen`, being disabled, will, for instance, prevent phpseclib from working. So will `fread` and `fputs`. But in the case of `phpinfo`...  phpseclib doesn't really need it that much. It does some quick compatibility checking but if the function is disabled phpseclib can run just fine.

I would say that doing `function_exists('phpinfo')` would be better but then again...   calling `phpinfo()` when it's disabled doesn't result in a fatal error - it results in a warning. A warning that is suppressed. Thus `ob_get_contents()` won't return anything and then OpenSSL will default to being enabled because of `case !isset($versions['Header'])` and `case !isset($versions['Library'])`.

I'd be up for removing the `phpinfo()` checks from the master branch (which is what this PR is targeting) but not because of `phpinfo()`. Rather, I'd be up for removing it because of the output buffering, which I'm now thinking I dislike. But if we do that then I think the version check should be removed as well. The constants, however - `MATH_BIGINTEGER_OPENSSL_DISABLE`, `MATH_BIGINTEGER_OPENSSL_ENABLED` and `CRYPT_RSA_MODE` should probably be kept.

raggi:
I would honestly expect you could nuke the whole set of checks. I doubt they apply to safe-to-run systems anyway, and the original bug was a bug in _other software_.

If you are able to articulate clearly exactly how you might like the code to be rearranged, I'll be happy to do so.

raggi:
This just hit another customer. Has an alternative solution been proposed?

terrafrost:
What about just always using openssl on the master branch (which your PR is targeting), anyway? No version check, no phpinfo check, nothing. I'm thinking of upping the PHP requirements of the master branch to 5.6, anyway, so that, in and of itself, might take care of old OpenSSL versions.

That said, people shouldn't be using the master branch as the API is non-stable anyway.

As for the 1.0 / 2.0 branches...  if I actually knew what the error was I might be able to able to come up with alternative ideas for fixing it. But as is it's not an issue I'm able to reproduce and you've given me precious little information on the error.

Quoting my last post, "_calling `phpinfo()` when it's disabled_ (via `the `disable_functions` PHP INI directive)  _doesn't result in a fatal error - it results in a warning. A warning that is suppressed._" So it is unclear to me how anyone would be seeing an error. If there are conditions under which `@` doesn't actually suppress errors that'd be news to me.

I can test to see if `function_exists('phpinfo')` returns true or not for disabled functions and include that if it does but I just don't see the point given the above.

Is the issue that the `phpinfo` call fixes old? Yes. But then again, so is PHP4.4 and the 1.0 branch still supports it (assuming PHP_Compat is used). The fact that PHP4-style constructors exist cause E_DEPRECATED warnings for the 1.0 branch on PHP 7 but changing it would be a BC break. Removing old functionality just because it's intended for out-of-date versions of PHP is BC-breaking behavior.

raggi:
The most recent case was caused by a valid difference between Library and Header in openssl versions in a way that is non-breaking. These values being different does not in and of itself indicate any problem.

terrafrost:
I've removed the `phpinfo()` check from the master branch:

https://github.com/phpseclib/phpseclib/commit/7a628c62b0071d03bde6be82c2f77acdadb20df2

For the 1.0 and 2.0 branches...  not using OpenSSL doesn't make phpseclib unusable - it just makes certain operations slower. At least if the gmp extension isn't present. But I can make loosen the condition that's currently in place. Instead of doing `$versions['Header'] == $versions['Library']` I could do something like `$versions['Header'] == $versions['Library'] && version_compare($versions['Header'], '1.0.1a') < 0)` or some such (ie. the version in the header is less than 1.0.1a or something). But to do that I'd need to know what you're getting as the header and library versions.

raggi:
e.g. http://a-know-phptest.appspot.com/

OpenSSL Library Version	OpenSSL 1.0.2g 1 Mar 2016
OpenSSL Header Version	OpenSSL 1.0.1s 1 Mar 2016

terrafrost:
For the 1.0 and 2.0 branches I've made it so that if both the header version and library version are both above 1.0.0 it'll let OpenSSL be used. When it was implemented the specific issue was with 1.0.0e (library) vs 0.9.8r (header).

PR #947, opened 2016-02-11
Correct Sodium check

paragonie-scott:
PECL Libsodium 1.0.x uses `\Sodium\function_name` instead of `\Sodium::method_name`.

https://paragonie.com/book/pecl-libsodium/read/03-utilities-helpers.md#version

### DIFF
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -1171,7 +1171,7 @@ class SSH2
             'diffie-hellman-group-exchange-sha1', // RFC 4419
             'diffie-hellman-group-exchange-sha256', // RFC 4419
         );
-        if (!class_exists('\Sodium')) {
+        if (!function_exists('\\Sodium\\library_version_major')) {
             $kex_algorithms = array_diff(
                 $kex_algorithms,
                 array('curve25519-sha256@libssh.org')


terrafrost:
Seems like doing both might be worthwhile? `class_exists` for people using <1.0.0 and `function_exists` for people using >=1.0.0? idk - I'm okay doing just >=1.0.0 as well!

It's just hard to have a super informed opinion since I have no idea what the deployment of either are like. Like if <1.0.0 had an install base of half a billion and >=1.0.0 had an install base of ten then keeping <1.0.0 support, for the time being, would probably be prudent, but idk what the install base is like!

paragonie-scott:
> It's just hard to have a super informed opinion since I have no idea what the deployment of either are like.

I would strongly recommend not running anything less than 1.0.1 especially on PHP 7: https://github.com/jedisct1/libsodium-php/issues/58

I've audited a couple applications that use libsodium and wrote Halite (a libsodium wrapper). My impression is that the install base is mostly going to be >= 1.0.1.

terrafrost:
Works for me! I'll try to merge tomorrow.

Thanks!

terrafrost:
I cherry picked this into the 2.0 branch and then merged the 2.0 branch into master.

Thanks!

PR #955, opened 2016-03-18
changing valid condition order, _slidingWindow biginteger

abhishektaneja:
the loop throws a warning undefined offer 6 at Math/BigInteger.php:1827

        // calculate the appropriate window size.
        // $window_size == 3 if $window_ranges is between 25 and 81, for example.
        for ($i = 0, $window_size = 1; $e_length > $window_ranges[$i] && $i < count($window_ranges); ++$window_size, ++$i) {
        }

It should be 

        // calculate the appropriate window size.
        // $window_size == 3 if $window_ranges is between 25 and 81, for example.
        for ($i = 0, $window_size = 1; $i < count($window_ranges) && $e_length > $window_ranges[$i] ; ++$window_size, ++$i) {
        }

### DIFF
--- a/phpseclib/Math/BigInteger.php
+++ b/phpseclib/Math/BigInteger.php
@@ -1783,7 +1783,7 @@ class BigInteger
 
         // calculate the appropriate window size.
         // $window_size == 3 if $window_ranges is between 25 and 81, for example.
-        for ($i = 0, $window_size = 1; $e_length > $window_ranges[$i] && $i < count($window_ranges); ++$window_size, ++$i) {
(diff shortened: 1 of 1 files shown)


bantu:
The patch looks good to me. I did not look at the context, though.

bantu:
Do you have an example for reproductiono or even better a test case?

bantu:
Do you have an example for reproduction (or even better a test case)?

terrafrost:
This has been merged:

https://github.com/phpseclib/phpseclib/commit/cb2ccea219ac050ad92c7aaa296fe3110e55d634

I created a unit test demonstrating the issue as well:

https://github.com/phpseclib/phpseclib/commit/56acb6723cd6f8cddc27e87f41f74afed1137f46

PR #956, opened 2016-03-24
implement sha512/224 and sha512/256

terrafrost:
This PR implements sha512/224 and sha512/256 and makes RSA.php fully PKCS1 v2.2 compliant.

A feature request exists to bring these hash algorithms to PHP but has yet to be acted upon: https://bugs.php.net/71177

I did not add unit tests for the HMAC'd version of those because I could not find any official test vectors for those.

### DIFF
--- a/phpseclib/Crypt/Hash.php
+++ b/phpseclib/Crypt/Hash.php
@@ -33,6 +33,7 @@
 
 namespace phpseclib\Crypt;
 
+use phpseclib\Math\BigInteger;
 use phpseclib\Exception\UnsupportedAlgorithmException;
 
 /**
@@ -80,6 +81,39 @@ class Hash
     var $key = false;
 
     /**
+     * Initial Hash
+     *
+     * Used only for sha512/*
+     *
+     * @see self::_sha512()
+     * @var array
+     * @access private
+     */
+    var $initial = false;
+
+    /**
+     * Outer XOR (Internal HMAC)
+     *
+     * Used only for sha512/*
+     *
+     * @see self::hash()
+     * @var string
+     * @access private
+     */
+    var $opad;
+
+    /**
+     * Inner XOR (Internal HMAC)
+     *
+     * Used only for sha512/*
+     *
+     * @see self::hash()
+     * @var string
+     * @access private
+     */
+    var $ipad;
+
+    /**
      * Default Constructor.
      *
      * @param string $hash
@@ -88,6 +122,9 @@ class Hash
     function __construct($hash = 'sha256')
     {
         $this->setHash($hash);
+
+        $this->ipad = str_repeat(chr(0x36), 128);
+        $this->opad = str_repeat(chr(0x5C), 128);
     }
 
     /**
@@ -126,10 +163,13 @@ class Hash
     {
         $this->hashParam = $hash = strtolower($hash);
         switch ($hash) {
+            case 'md2-96':
             case 'md5-96':
             case 'sha1-96':
             case 'sha256-96':
             case 'sha512-96':
+            case 'sha512/224-96':
+            case 'sha512/256-96':
                 $hash = substr($hash, 0, -3);
                 $this->length = 12; // 96 / 8 = 12
                 break;
@@ -140,7 +180,11 @@ class Hash
             case 'sha1':
                 $this->length = 20;
                 break;
+            case 'sha512/224':
+                $this->length = 28;
+                break;
             case 'sha256':
+            case 'sha512/256':
                 $this->length = 32;
                 break;
             case 'sha384':
@@ -171,6 +215,23 @@ class Hash
                 );
         }
 
+        if ($hash == 'sha512/224' || $hash == 'sha512/256') {
+            // from http://csrc.nist.gov/publications/fips/fips180-4/fips-180-4.pdf#page=24
+            $this->initial = $hash == 'sha512/256' ?
+                array(
+                    '22312194FC2BF72C', '9F555FA3C84C64C2', '2393B86B6F53B151', '963877195940EABD',
+                    '96283EE2A88EFFE3', 'BE5E1E2553863992', '2B0199FC2C85B8AA', '0EB72DDC81C52CA2'
+                ) :
+                array(
+                    '8C3D37C819544DA2', '73E1996689DCD4D6', '1DFAB7AE32FF9C82', '679DD514582F9FCF',
+                    '0F6D2B697BD44DA8', '77E36F7304C48942', '3F9D85A86A1D36C8', '1112E6AD91D692A1'
+                );
+            for ($i = 0; $i < 8; $i++) {
+                $this->initial[$i] = new BigInteger($this->initial[$i], 16);
+                $this->initial[$i]->setPrecision(64);
+            }
+        }
+
         $this->hash = $hash;
     }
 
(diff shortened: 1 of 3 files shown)


phpseclib/Crypt/Hash.php
```diff
@@ -140,7 +180,11 @@ function setHash($hash)
             case 'sha1':
                 $this->length = 20;
                 break;
+            case 'sha512/224':
```
Joey3000: 'sha224' has also been added (see https://github.com/phpseclib/phpseclib/pull/956/files#diff-b4b2acaebb5e3d125481ec5da11baff5R1972), but its length is not set here. Is this on purpose?

tests/Unit/Crypt/HashTest.php
```diff
@@ -95,6 +95,33 @@ public static function hashData()
                 'The quick brown fox jumps over the lazy dog.',
                 '87a7ff096082e3ffeb86db10feb91c5af36c2c71bc426fe310ce662e0338223e217def0eab0b02b80eecf875657802bc5965e48f5c0a05467756f0d3f396faba'
             ),
+            array(
+                'whirlpool',
+                'The quick brown fox jumps over the lazy dog.',
+                '87a7ff096082e3ffeb86db10feb91c5af36c2c71bc426fe310ce662e0338223e217def0eab0b02b80eecf875657802bc5965e48f5c0a05467756f0d3f396faba'
+            ),
+            // from http://csrc.nist.gov/groups/ST/toolkit/documents/Examples/SHA512_224.pdf
```
Joey3000: 'sha224' has also been added (see https://github.com/phpseclib/phpseclib/pull/956/files#diff-b4b2acaebb5e3d125481ec5da11baff5R1972), but it doesn't have test cases defined here. Is this on purpose?
Please note that its examples are similarly available in http://csrc.nist.gov/groups/ST/toolkit/documents/Examples/SHA224.pdf.

terrafrost: I didn't think of it lol. But yah - they should be added. I'll do that.

Thanks!

terrafrost: The unit tests have been added.

tests/Unit/Crypt/HashTest.php
```diff
@@ -183,6 +221,19 @@ public static function hmacData()
...
+                str_repeat("\x0b", 20),
+                'Hi There',
+                '896fb1128abbdf196832107cd49df33f47b4b1169912ba4f53684b22',
+            ),
+            array(
+                'sha224',
+                'Jefe',
+                'what do ya want for nothing?',
+                'a30e01098bc6dbbf45690f3a7e9e6d0f8bbea2a39e6148008fd05e44',
+            ),
```
Joey3000: Thanks for adding this! Just a few minor points more, in case it has been missed:

* I just noticed that the test vectors for HMAC-SHA-**384** are also missing here (they are also in https://tools.ietf.org/rfc/rfc4231.txt).
* There are more test cases for HMAC-SHA-256 and HMAC-SHA-512 (namely four each) than what has been added here - see, e.g., https://github.com/terrafrost/phpseclib/blob/089a4cb101c025611d066e13bd8c3e305c451558/tests/Unit/Crypt/HashTest.php#L160-L188.

Please also note the comment in https://github.com/phpseclib/phpseclib/pull/956#discussion_r58241646.

Thanks!

terrafrost: > There are more test cases for HMAC-SHA-256 and HMAC-SHA-512 (namely four each) than what has been added here

Yah - I know. I just don't know that I care that much lol. I think adding a subset is probably sufficient. I mean, I can add the other two, as well, but idk...

terrafrost:
This has been merged.

Joey3000:
Thanks for all the updates! Just a few minor points:

1. This hasn't been addressed: https://github.com/phpseclib/phpseclib/pull/956#discussion_r58241646. To give it more details: For all 256-bit hashes, the length is set in https://github.com/terrafrost/phpseclib/blob/a757584630ce1f7c473a6964a9c9697f5c5bb6fd/phpseclib/Crypt/Hash.php#L186-L189. But 'sha224' is missing in https://github.com/terrafrost/phpseclib/blob/a757584630ce1f7c473a6964a9c9697f5c5bb6fd/phpseclib/Crypt/Hash.php#L183-L185. So, shouldn't the latter be, similarly:

```php
            case 'sha224':   // this line is currently missing
            case 'sha512/224':
                $this->length = 28;
                break;
```

  `2`. This isn't really related, but since the hash tests are touched, I thought it could be done here, too:

   `a`. The exact same Whirlpool test is there three times: https://github.com/terrafrost/phpseclib/blob/3aad5a70e57dd5ad4313af137c81fffbe375921c/tests/Unit/Crypt/HashTest.php#L88-L102. Should two of them be removed, or one removed and another modified to not have the final point in the test string? On the other hand, there seem to be official test vectors in http://www.larc.usp.br/~pbarreto/whirlpool.zip (linked to from the official http://www.larc.usp.br/~pbarreto/WhirlpoolPage.html; the tests are supposedly ISO/IEC 10118-3:2004 conform, but I can't confirm since the latter is behind a paywall).

  `b`. But, actually, Whirlpool doesn't seem to be supported (any longer?) by phpseclib - searching for the "whirlpool" string only brings up its tests. (Except for phpseclib\Crypt\Base.php where its PHP-provided implementation is used if available.) So, maybe its tests need to be removed altogether?

Thanks in advance!

terrafrost:
Nice finds! I've updated the code accordingly:

https://github.com/phpseclib/phpseclib/commit/e40d6cc1cc9c7f0a6411583a599a4c54aadec31b

> On the other hand, there seem to be official test vectors in http://www.larc.usp.br/~pbarreto/whirlpool.zip (linked to from the official http://www.larc.usp.br/~pbarreto/WhirlpoolPage.html; the tests are supposedly ISO/IEC 10118-3:2004 conform, but I can't confirm since the latter is behind a paywall).

I think it's fine as is tbh.

> But, actually, Whirlpool doesn't seem to be supported (any longer?) by phpseclib - searching for the "whirlpool" string only brings up its tests. (Except for phpseclib\Crypt\Base.php where its PHP-provided implementation is used if available.) So, maybe its tests need to be removed altogether?

Yah - the way way Crypt/Hash.php works is that if the algorithm isn't explicitly supported it'll fall back to see if PHP's hash function provides built-in support. See https://github.com/phpseclib/phpseclib/pull/749#issuecomment-122706354 for more info. I guess what we could do is make it so that it throws an exception if an unsupported algorithm is provided. This would be consistent with the changes I made to the symmetric ciphers to make them less tolerant of bad input.

ie. in 1.0 Hash.php silently mangles bad input. If you do `new Crypt_Hash('zzz')` it'll do sha1. If the provided option is invalid it'll silently fall back to a valid option. Which is what how the symmetric ciphers worked in 1.0 / 2.0. ie. silently null padding keys, IV's, etc. But the paradigm in the master branch is already different for symmetric ciphers so changing it for hashes, as well, would be consistent.

Joey3000:
Thanks for the updates!

> I guess what we could do is make it so that it throws an exception if an unsupported algorithm is provided. This would be consistent with the changes I made to the symmetric ciphers to make them less tolerant of bad input.

That makes sense to me.

Issue #761, opened 2015-07-31
unpack() is not cache-timing safe


paragonie-scott:
If `unpack()` is used on cryptographic secrets, it may leak information through cache-timing side-channels

Background:
* http://cr.yp.to/antiforgery/cachetiming-20050414.pdf
* https://crypto.stackexchange.com/a/5527/24405

Based on:
* https://github.com/phpseclib/phpseclib/search?utf8=%E2%9C%93&q=unpack
* https://github.com/magento/magento2/blob/develop/lib/internal/phpseclib/Crypt/Random.php#L71

See also:
* https://github.com/php/php-src/blob/c47c526290b924a1588f317abce94ca3eeddb980/ext/standard/pack.c#L881-L913
* https://github.com/php/php-src/blob/c47c526290b924a1588f317abce94ca3eeddb980/ext/standard/pack.c#L538

-------

### Safe examples in PHP

Cache-timing-safe bin2hex-encoding (taken from libsodium):
* https://github.com/defuse/php-encryption/blob/d2c33279e31d5f3d4b74a82139b6ce56d551965b/src/Crypto.php#L691-L747

Cache-timing-safe bin2int (from random_compat):
* https://github.com/paragonie/random_compat/blob/d6efc0a466f973cdf8235fc3a53dfc022354c32c/lib/random.php#L380-L393

terrafrost:
phpseclib re-implementing functions to protect against timing attacks is certainly not without precident. One example I can think of off hand:

https://github.com/phpseclib/phpseclib/blob/master/phpseclib/Crypt/RSA.php#L2201

Anyway, I'll try to take a look at this this weekend or, depending on how busy I am, early next week.

Thanks!

terrafrost:
On what basis are you alleging that unpack() is vulnerable to timing attacks? You posted a link to the C implementation of unpack and two "background" links as if that's a prima facie explanation but I would like more detail.

My thoughts follow.

So in order to be vulnerable to time attacks the processing time has to vary based on the input. In the case of `==` it makes sense if we consider the most intuitive implementation of `==`. eg. if the length's are equal go through each character one by one, until you find a non-matching character, and then return immediately. `RSA::_equals()` addresses this by creating an implementation of `==` that doesn't return prematurely.

In the case of `unpack('N', ...)` it's essentially doing this:

```
issigned = input[inputpos] & 0x80;
map = big_endian_long_map;
if (SIZEOF_ZEND_LONG > 4 && issigned) {
[...]

PR #957, opened 2016-03-24
Constant-time encoding

paragonie-scott:
See #761

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -56,6 +56,7 @@
     "require-dev": {
         "phing/phing": "~2.7",
         "phpunit/phpunit": "~4.0",
+        "paragonie/constant_time_encoding": "^0.5",
         "sami/sami": "~2.0",
         "squizlabs/php_codesniffer": "~2.0"
     },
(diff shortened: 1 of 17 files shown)


composer.json
```diff
@@ -56,6 +56,7 @@
     "require-dev": {
         "phing/phing": "~2.7",
         "phpunit/phpunit": "~4.0",
+        "paragonie/constant_time_encoding": "^0.5",
```
GrahamCampbell: :-1: until this is stable

GrahamCampbell:
Surely this would be better targeted at 2.x, rather than 3.x, since it could be a year before 3.0 gets released.

paragonie-scott:
Okay, I'll send a superseding PR against 2.x once I've gotten enough independent review to feel confident about tagging 1.0.0.

GrahamCampbell:
:+1:

terrafrost:
One thing we need to keep in mind with doing Composer stuff for the 2.0 branch is that 2.0 doesn't necessarily require composer. What it requires is a PSR-4 autoloader whereas composer does PSR-0, classmap and files autoloading as well. I guess that doesn't apply so much to paragonie/constant_time_encoding as it does to paragonie/random_compat but it is, none-the-less, worth keeping in mind.

With 3.0, in contrast, we're free to do mostly whatever we want as there's no BC we need to worry about.

paragonie-scott:
By the way, we tagged v1.0 and v2.0: https://github.com/paragonie/constant_time_encoding/releases

PR #958, opened 2016-03-24
Use random_compat

paragonie-scott:


composer.json
```diff
@@ -51,6 +51,7 @@
         }
     ],
     "require": {
+        "paragonie/random_compat": "^1.4",
```
GrahamCampbell: doing this conflicts with people who want 2.x

paragonie-scott: Good point. Shall I change it to `^1.4|2.0`?

GrahamCampbell: `^1.4|^2.0`

terrafrost:
What if we just do away with the whole `version_compare` thing and do this?:

```php
        try {
            return \random_bytes($length);
        } catch (Exception $e) {}

        // at this point we have no choice but to use a pure-PHP CSPRNG
        // ...
```

phpseclib/Crypt/Random.php
```diff
@@ -46,7 +46,7 @@ class Random
      */
     static function string($length)
     {
-        if (version_compare(PHP_VERSION, '7.0.0', '>=')) {
+        if (version_compare(PHP_VERSION, '7.0.0', '>=') || defined('RANDOM_COMPAT_READ_BUFFER')) {
             try {
                 return \random_bytes($length);
             } catch (\Throwable $e) {
```
GrahamCampbell: we need to catch Exception now too

bantu: @GrahamCampbell Exception implements Throwable?

paragonie-scott: Not in PHP 5. I couldn't make Exception implement Throwable with random_compat

bantu: Oh ok. Now I see. Add another empty catch block for Exception?

GrahamCampbell: yeh

paragonie-scott:
OK, how does that look now?

phpseclib/Crypt/Random.php
```diff
@@ -46,64 +46,17 @@ class Random
...
+        try {
+            return \random_bytes($length);
+        } catch (\Throwable $e) {
+            // If a sufficient source of randomness is unavailable, random_bytes() will throw an
+            // object that implements the Throwable interface (Exception, TypeError, Error).
+            // We don't actually need to do anything here. The string() method should just continue
+            // as normal. Note, however, that if we don't have a sufficient source of randomness for
+            // random_bytes(), most of the other calls here will fail too, so we'll end up using
+            // the PHP implementation.
+        } catch (\Exception $e) {
```
GrahamCampbell: It's conventional to catch Exception first, then Throwable.

paragonie-scott: Ok, easily done.

paragonie-scott:
I think the unit tests are failing because of `composer.lock`.

terrafrost:
Looks good! I'll figure out what's up with the unit tests later - I need to go do some errands lol.

Thanks!!

terrafrost:
The composer.lock was the issue and the PR has been merged.

Thanks!

PR #966, opened 2016-04-08
Fixed PHP notice thrown by undefined key_comment variable

d-naumann-yatego:
Variable key_comment should be predefined, otherwise there will be PHP notices if key_comment is undefined, cause its optional.

### DIFF
--- a/phpseclib/System/SSH/Agent.php
+++ b/phpseclib/System/SSH/Agent.php
@@ -169,6 +169,7 @@ class Agent
         $identities = array();
         $keyCount = current(unpack('N', fread($this->fsock, 4)));
         for ($i = 0; $i < $keyCount; $i++) {
+            $key_comment = '';
             $length = current(unpack('N', fread($this->fsock, 4)));
             $key_blob = fread($this->fsock, $length);
             $length = current(unpack('N', fread($this->fsock, 4)));


bantu:
@terrafrost This looks good. But more things are broken here, see https://github.com/d-naumann-yatego/phpseclib/commit/2039530f7c865b8370d4d30137e8b9045c737261.

terrafrost:
When I add a key without a comment ssh-agent auto-adds "rsa w/o comment" as the comment (as the following demonstrates [requires the master branch]):

```php
<?php
include 'vendor/autoload.php';

error_reporting(E_ALL);

$agent = new \phpseclib\System\SSH\Agent();

foreach ($agent->requestIdentities() as $identity) {
	echo $identity->getPublicKey('OpenSSH') . "\n\n";
}

```

The output I get back from ssh-agent is as follows:

```
Identity added: test.pem (rsa w/o comment)
```

So I don't know how to duplicate the reported issue but I'm willing to concede that there may be circumstances where keys wouldn't have comments, despite my inability to create such keys. And even if there weren't the fact that the comment is being conditionally added...  if the comment were required the `$length` check ought not be there. So I think this is a good observation.

Also, I don't know why one of the `$length = current(unpack('N', fread($this->fsock, 4)));` was deleted from https://github.com/d-naumann-yatego/phpseclib/commit/2039530f7c865b8370d4d30137e8b9045c737261 . That change breaks Agent.php. But the change from `\BadConfigurationException` to `BadConfigurationException` is a good one.

Anyway, I've adapted this change to 1.0 and merged 1.0 into 2.0 and master.

Thanks!

PR #969, opened 2016-04-11
Mark the conflict with Math_BigInteger

ftzdomino:
phpseclib contains a diverged fork of PEAR's Math_BigInteger in the global namespace, causing a conflict if both are installed.  This change request marks them as conflicting.  Ideally, the phpseclib variant would be contained in its own namespace.


ftzdomino:
This was based on an old version.  Please ignore.

Issue #973, opened 2016-04-23
Please remove dependencies on autoloaders


cubiclesoft:
phpseclib has gotten a bit too dependency-heavy as of late.  Currently, to instantiate the RSA class without an autoloader and/or Composer, a user has to do something along the lines of:

```php
	$rootpath = str_replace("\\", "/", dirname(__FILE__));

	require_once $rootpath . "/path/to/phpseclib/Crypt/RSA.php";
	require_once $rootpath . "/path/to/phpseclib/Crypt/Hash.php";

	$path = $rootpath . "/path/to/phpseclib/Crypt/RSA/";
	$dir = opendir($path);
	if ($dir)
	{
		while (($file = readdir($dir)) !== false)
		{
			if (substr($file, -4) === ".php")  require_once $path . $file;
		}

		closedir($dir);
	}

	require_once $rootpath . "/path/to/phpseclib/Math/BigInteger.php";

	$rsa = new phpseclib\Crypt\RSA();
	...
```

To have any hope of future compatibility, the opendir()/readdir()/closedir() bit is necessary.  The files in the RSA directory might as well be bundled with the RSA class since instantiating/using RSA() throws fits until every last file in the RSA directory is loaded.

A possible solution from within your classes could look like:

```php
	if (!class_exists("phpseclib\\Crypt\\RSA\\MSBLOB", false))  require_once str_replace("\\", "/", dirname(__FILE__)) . "/RSA/MSBLOB.php";
```

OR:

```php
	if (!class_exists("phpseclib\\Crypt\\RSA\\MSBLOB", false) && file_exists(str_replace("\\", "/", dirname(__FILE__)) . "/RSA/MSBLOB.php"))  require_once str_replace("\\", "/", dirname(__FILE__)) . "/RSA/MSBLOB.php";
```

You probably know the most common use-cases for each class in phpseclib and therefore can preload the appropriate classes as-needed with something like the above without resorting to an autoloader.  Inclusion might even be as early as __construct() (i.e. long before an autoloader would ever be triggered).  Your library also knows where all of its files are supposed to be in relation to each other.

Not everyone wants to use autoloaders, Composer, PSRs, etc.  Some of us actually prefer small, compact libraries and don't mind a few 'require_once's here and there.

terrafrost:
[...]

PR #986, opened 2016-05-10
Constant-time RFC-4648 encoding

paragonie-scott:
Supersedes #957 

(Will fix the autoloading in my next commit)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -51,7 +51,8 @@
         }
     ],
     "require": {
-        "php": ">=5.3.3"
+        "php": ">=5.3.3",
+        "paragonie/constant_time_encoding": "^1|^2"
     },
     "require-dev": {
         "phing/phing": "~2.7",
--- a/composer.lock
+++ b/composer.lock
@@ -4,9 +4,72 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "8599992bf6058a9da82372eb8bcae2c2",
-    "content-hash": "fde47c84178c55c06de858a2128e3d07",
-    "packages": [],
+    "hash": "d601a1cc278a49b0490ceb146c5b4512",
+    "content-hash": "dd5b2810894da3cb4757ef8ca5f630bf",
+    "packages": [
+        {
+            "name": "paragonie/constant_time_encoding",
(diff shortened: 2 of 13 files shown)


paragonie-scott:
OK, I'm not sure why 5.5 is failing but 5.5.9 is passing, but I'm going to pause any tweaks for now.

bantu:
@paragonie-scott The 5.5 environment runs the code sniffer. The problem is "There must be one blank line after the namespace "PSR2.Namespaces.NamespaceDeclaration.BlankLineAfter". See https://travis-ci.org/phpseclib/phpseclib/jobs/129270067#L577 and following.

bantu:
@paragonie-scott Might want to fix the author/commiter on cda2cdc. There is a typo in the e-mail address.

terrafrost:
Would you be up for doing this for the master branch instead?

In https://github.com/phpseclib/phpseclib/pull/957#issuecomment-203037624 I suggested that this would be fine in 2.0 if it was PSR-4 but I've since had a bit of a change of heart, in part due to #973. Among other things that made me realize was...  in a lot of the 2.0 examples I do this:

```php
include 'autoload.php';

$loader->addPsr4('phpseclib\\', __DIR__ . '/path/to/phpseclib2.0');
$loader->register();
```

(where autoload.php is https://github.com/composer/composer/blob/master/src/Composer/Autoload/ClassLoader.php)

source: http://phpseclib.sourceforge.net/2.0.html

Problem is...  if people are doing that (as is done in the examples) then they'd have to add a new `$loader->addPsrR('...');` line to every file that's using phpseclib (assuming they're not using a common file). What would have been better is if I had just included autoload.php in the 2.0 branch. The vendor/composer directory could exist (maybe as a submodule) and autoload.php could have whatever lines it needed to have. And that one wouldn't be a submodule (or maybe it would be idk).

I mean, probably 95% of 2.0 users are using Composer but for the (guestimated) 5% that aren't this could cause problems. Meh :( This makes me glad that in the master / 3.0 branch I'm just requiring composer outright. You can certainly get phpseclib to work without composer but if you do so with master / 3.0 you're doing so at your own peril.

paragonie-scott:
> There is a typo in the e-mail address.

Huh, I didn't even notice. Thanks.

> Would you be up for doing this for the master branch instead?

Sure, not a problem.

PR #988, opened 2016-05-12
Use paragonie/constant_time_encoding

paragonie-scott:
As per the comment in #986

### DIFF
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -49,6 +49,7 @@
 
 namespace phpseclib\Net;
 
+use ParagonIE\ConstantTime\Base64;
 use phpseclib\Crypt\Base;
 use phpseclib\Crypt\Blowfish;
 use phpseclib\Crypt\Hash;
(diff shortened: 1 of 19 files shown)


terrafrost:
Thanks! I'm about to get on a plane but I'll try to take a look and git it merged after I land or this weekend as time permits.

paragonie-scott:
Safe travels. :)

phpseclib/Net/SSH2.php
```diff
@@ -4096,7 +4097,7 @@ function getServerPublicHostKey()
                 throw new NoSupportedAlgorithmsException('Unsupported signature format');
         }
 
-        return $this->signature_format . ' ' . base64_encode($this->server_public_host_key);
+        return $this->signature_format . ' ' . Base64::encode($this->server_public_host_key);
```
GrahamCampbell: this is ineeded

GrahamCampbell: this is not needed

GrahamCampbell: this information is already public

terrafrost:
This has been merged. Net/SFTP.php didn't have the use declaration, which was causing Travis CI to fail so I added that and re-based that into your commit.

PR #995, opened 2016-06-02
fixed sending stream resulting in a zero byte file

mpipet:
When using SFTP `put()` with the STDIN file descriptor and the content is *piped* (and not redirected, see [this](https://bugs.php.net/bug.php?id=23424)), fstat (see [here](https://github.com/phpseclib/phpseclib/blob/master/phpseclib/Net/SFTP.php#L1876)) returns a size of 0. Consequently an empty file is created on the remote system, even if the local file descriptor provides a non-empty stream.

This PR provides a solution by reading the stream even if size is equal to zero, and letting fread detect the end of file.

Let me know if anything more can be done to get this PR merged

Thanks !


terrafrost:
I rebased this into one commit and then cherry-picked this into 2.0 and master:

https://github.com/phpseclib/phpseclib/commit/6c8affcf11a8cfb256c98e53933a6298b33a5d13

Thanks!

PR #996, opened 2016-06-05
remove a few calls to paragonie/constant_time_encoding

terrafrost:
- Hex::decode('aabb') can be replaced with "\xaa\xbb"
- Hex::decode(sha1('...')) can be replaced with sha1('...', true)

### DIFF
--- a/phpseclib/Crypt/Base.php
+++ b/phpseclib/Crypt/Base.php
@@ -35,9 +35,6 @@
  */
 
 namespace phpseclib\Crypt;
-use ParagonIE\ConstantTime\Hex;
-
-use ParagonIE\ConstantTime\Hex;
 
 /**
  * Base Class for all \phpseclib\Crypt\* cipher classes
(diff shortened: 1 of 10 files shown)


phpseclib/Crypt/Base.php
```diff
@@ -2581,10 +2578,10 @@ function _hashInlineCryptFunction($bytes)
                 $len = strlen($bytes);
                 for ($i = 0; $i < $len; $i+=20) {
                     $t = substr($bytes, $i, 20);
-                    $hash = Hex::decode($hash);
+                    $hash = sha1($hash, trie);
```
bantu: Typo

bantu:
I think most of these were pointless in the first place as public information was converted.

GrahamCampbell:
> I think most of these were pointless in the first place as public information was converted.

I agree.

terrafrost:
Yah - I do too. Meaning no disrespect to anyone, but I think it was just a mass replace that took place. Find all instances of `pack()` and replace with `Hex::decode()`, etc, even if there are better approaches in specific instances.

I figure I'll go back and remove the calls from the public stuff at some point.

And FWIW the `sha1(...)` to `sha1(..., true)` stuff could be done in the 2.0 branch, as well, but it's not a huge priority for me. It can't be done in the 1.0 branch for BC purposes (the second parameter of `md5` and `sha1` wasn't added until PHP 5.0.0).

PR #997, opened 2016-06-05
Change default signature algorithm from SHA1 to SHA256

DennisBirkholz:
All major browsers will stop to accept certificates that use SHA1 by 2017 (see https://en.wikipedia.org/wiki/SHA-1). So at least SHA256 should be the default now.

### DIFF
--- a/phpseclib/File/X509.php
+++ b/phpseclib/File/X509.php
@@ -3222,7 +3222,7 @@ class X509
      * @access public
      * @return mixed
      */
-    function sign($issuer, $subject, $signatureAlgorithm = 'sha1WithRSAEncryption')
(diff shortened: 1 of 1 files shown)


terrafrost:
This was something I was going to do when I redo X509 for the master branch. Basically, my planned X509 changes include moving the CSR / SPKAC / CRL stuff into their own files. ie. X509 won't look even remotely similar to what it does today by the time I'm done with it.

The default hash for RSA.php was changed some time ago but the rewrite of RSA.php is also complete, unlike the rewrite for X509.php.

But that said, doing it early, too, isn't really a problem, so I guess I'll go ahead and do the merge.

Issue #871, opened 2015-11-16
Can't read CRL using phpseclib


danowar2k:
Hello, (Hallo)

I'm trying to read a CRL file using phpseclib.
The CRL in question is:
http://ca.fernuni-hagen.de/certserver/certs/crl2009.crl

My code:
        $der_data = file_get_contents(CertificateValues::getLocalCrlLocation()); // works local and via url
        $der2pem = someClass::der2pem($der_data);
        $crl = new File_X509();
        $crl->loadCRL($der2pem);

Did I miss something? Do I have to have something configured differently?
My code seems to hang around inside the _decodeBER function. The file is 2MB big.

terrafrost:
First, I apologize for the tardyness of this response. I've been out of town this week per https://github.com/phpseclib/phpseclib/issues/835 .

Anyway, this is slow for a number of reasons.

1. It's doing lots of substr()'s. There are ~150,000 revoked certs in that CRL so that's over 300,000 substr() calls at minimum, thanks to __string_shift(), which, itself, does 2x substr() calls.

 The problem with that is that substr() just isn't the fastest thing in the universe. See https://github.com/phpseclib/phpseclib/pull/129 for an example of how eliminating one substr() call yielded a pretty significant speedup.

2. A PHP array with ~150,000 elements (and subelements) is, itself, fairly slow. Check out https://github.com/terrafrost/phpseclib/commit/b3dfb99536428809606814b4c8e98234413c95a5 and read the in-line comments.

 Maybe it wouldn't be so slow on PHP7 since PHP7 is supposed to be a lot faster with arrays:

 https://nikic.github.io/2014/12/22/PHPs-new-hashtable-implementation.html

 But who knows.

danowar2k:
Gonna close this.

PR #1000, opened 2016-06-19
CRL optimization

johnsterling:
I happened to come across issue #871 which referenced a previous PR of mine dealing with performance, and it piqued my curiosity. With this pull request I think I have addressed the primary contributors to low performance when loading large CRLs.

By sprinkling some timestamp output throughout the code paths I determined 2 main areas where a slowdown was happening. The vast majority of time was spent in `X509->loadCRL()`, where it runs `_mapInExtensions()` for each item in `$crl['tbsCertList']['revokedCertificates']`. I narrowed this down to `is_array()` running abnormally slowly. As it turns out, this is due to the use of references. As far as I can tell, the section of code in question passes around references to arrays within arrays using `_subArray()` as a convenience for validating and addressing nested associative entries.

The problem lies in how PHP references work. Normally, variable assignment by-value operates with copy-on-write, so if you pass a large array to a function by-value, its contents don't actually get copied in memory unless one of the copied variables is written to. However, if a variable is created as a reference and it subsequently gets passed by-value to a function, it will be immediately copied.

So, what was happening is that in `_subArray()` the `$root` variable is passed as a reference, and then subsequently re-assigned locally as a reference to deeper sub-arrays, and at every step it is passed to `is_array()` which caused the entire thing to be copied in memory. Specifically, the `$rclist` array reference was passed to `_subArray()` once for every entry in the 'revokedCertificates' array. In addition, there are several places outside of `_subArray()` where a reference it returned was passed to `is_array()`.

At first I attempted to remove the use of references altogether, but that proved difficult as it's clear the reason for using them in the first place is to allow modification of these sub-arrays in place. So short of a massive refactor around that, my solution was to split the `_subArray()` function into 2 parts: a validator to which the array in question is passed by-value, and a reference returner which doesn't do any `is_array()` calls. This requires a bit of extra code everywhere `_mapInExtensions()` is called (to do the validation call first) but the speedup is huge (see below for performance data).

The second major slowdown area is in `ASN1->_decode_ber()`. This is where the use of `_string_shift()` as hypothesized in the issue actually manifests itself. Similar to PR #129, the issue here is that `_string_shift()` calls `substr()` on *both* halves of the (potentially large) input data as it is parsed. This causes both the (relatively small) current data to be copied for parsing, as well as the (relatively large) rest of the data to be copied, effectively only for the purpose of tracking where the next data is (it seems like the goal was to emulate consuming a file handle).

My solution to this was, rather than copying the remaining data over and over, to simply maintain a position index of where the next data is in the input string. Then, the only data that is copied is the "current" data being parsed, bit by bit.

Below are some timed runs of a test script to illustrate the speedups from the above changes. I used a local copy of the ~2MB CRL file from the original issue. Here is the test script:

```php
<?php

require_once 'vendor/autoload.php';

$crl_data = file_get_contents('crl2009.crl');
$cer = new phpseclib\File\X509();
$crl = $cer->loadCRL($crl_data);
```

Before any changes:
```
$ time php crltest.php
^C

real	70m55.993s
user	69m46.406s
sys	1m7.734s
$ 
```
As you can see, I killed the script after over an hour. Based on rough estimates from some previous runs with embedded timestamp output, it looked like this would have taken on the order of 10 hours to actually complete.

After the updates around the usage of `_subArray()`:
```
$ time php crltest.php

real	1m32.754s
user	1m31.713s
sys	0m1.002s
$ 
```

And after additionally the refactoring of the usage of `_string_shift()`:
```
$ time php crltest.php

real	0m6.804s
user	0m6.539s
sys	0m0.250s
$ 
```

As far as I can tell I have not affected the actual logic or correctness of the code with these optimizations, though I admit I haven't taken the time to 100% grok the CRL parser. The isolated operations I changed should be equivalent to the originals, and I did run the test suite to make sure that didn't catch anything.

As an additional note, I had to up my maximum memory setting in php.ini to even run this use case, and the PHP process ended up using about 430MB of RAM.

### DIFF
--- a/phpseclib/File/ASN1.php
+++ b/phpseclib/File/ASN1.php
@@ -225,14 +225,15 @@ class ASN1
      *
      * @param string $encoded
      * @param int $start
+     * @param int $encoded_pos
      * @return array
      * @access private
      */
-    function _decode_ber($encoded, $start = 0)
+    function _decode_ber($encoded, $start = 0, $encoded_pos = 0)
     {
         $current = array('start' => $start);
 
-        $type = ord($this->_string_shift($encoded));
+        $type = ord($encoded[$encoded_pos++]);
         $start++;
 
         $constructed = ($type >> 5) & 1;
@@ -244,23 +245,24 @@ class ASN1
             do {
                 $loop = ord($encoded[0]) >> 7;
                 $tag <<= 7;
-                $tag |= ord($this->_string_shift($encoded)) & 0x7F;
+                $tag |= ord($encoded[$encoded_pos++]) & 0x7F;
                 $start++;
             } while ($loop);
         }
 
         // Length, as discussed in paragraph 8.1.3 of X.690-0207.pdf#page=13
-        $length = ord($this->_string_shift($encoded));
+        $length = ord($encoded[$encoded_pos++]);
         $start++;
         if ($length == 0x80) { // indefinite length
             // "[A sender shall] use the indefinite form (see 8.1.3.6) if the encoding is constructed and is not all
             //  immediately available." -- paragraph 8.1.3.2.c
-            $length = strlen($encoded);
+            $length = strlen($encoded) - $encoded_pos;
         } elseif ($length & 0x80) { // definite length, long form
             // technically, the long form of the length can be represented by up to 126 octets (bytes), but we'll only
             // support it up to four.
             $length&= 0x7F;
-            $temp = $this->_string_shift($encoded, $length);
+            $temp = substr($encoded, $encoded_pos, $length);
+            $encoded_pos += $length;
             // tags of indefinte length don't really have a header length; this length includes the tag
             $current+= array('headerlength' => $length + 2);
             $start+= $length;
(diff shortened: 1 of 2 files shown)


terrafrost:
Nice job! I knew redoing `_decode_ber()` would speed things up but wasn't super interested in doing it lol.

But that said I think it could also speed up the thing https://github.com/terrafrost/phpseclib/commit/3cf5c9266ca6defd4db86fe24a3c8166e7df5c4b was supposed to speed up.

So the issue with that was that I wrote a CMS parser for someone. They were using it to verify and create signatures for PDFs. The PDF's were embedded in the CMS and so you'd have really large OCTET STRINGs. Except that the OCTET STRING, in this case, was of an indefinite length with a ton of 1,000 character length OCTET STRINGs nested within it. So _string_shift() was being called probably 10,000+ times and _string_shift() (and more generally, strings in general, imho) just isn't that fast. So what I did there is have a regular expression to assume that if you have one OCTET STRING of a fixed length assume that the rest of the strings are of the same length.

Anyway, I'll try to review over the course of the coming week and merge next weekend. That'll give other people a chance to review if they're so inclined. That said, I suspect it'll prob just be merged. It's passing all the unit tests and my cursory review of it looks good.

Nice job and thanks!!!

terrafrost:
I cherry-picked this into the 1.0 branch and then merged into 2.0 and master:

https://github.com/phpseclib/phpseclib/commit/a1e16797cacfa8559dcbfa0263fcd123d3b5c53b
https://github.com/phpseclib/phpseclib/commit/88ce26f8ca405a7572ef584b63131e0ef3c4b170

Thanks!!

PR #1001, opened 2016-06-21
Fix zero salt length RSA signatures.

andrewbelcher:
When checking whether we have a salt length, it uses a non-strict TRUE comparison, which results in a zero length not being distinguished from a NULL. This results in you not being able to generate signatures with a zero salt length (which is useful for testing).


terrafrost:
I rebased your commit into one, cherry-picked it into the 1.0 branch and then merged up to master:

https://github.com/phpseclib/phpseclib/commit/dd754283bb36ba2f5731bebc5305703dae9c7811

Thanks!

PR #1005, opened 2016-06-26
SFTP: make symlink support relative target's

### LINKED ISSUE #1004: SFTP: Relative symlinks
Hi

With regular SFTP, I am able to create a relative symlink like this:

```
sftp> symlink normal_file some_folder/my_link
```

This would create a link called `my_link` inside `some_folder` that points to the file at `some_folder/normal_file`. In other words, the link-target argument (`normal_file`) specifies a path relative to the directory containing the link, and is completely independent of the current working directory. The link-target argument is also equal to the value that would be returned by a subsequent call to readlink().

It seems that the SFTP::symlink function unintentionally changes these semantics. Firstly, it forces the target to always be converted to an absolute path, making it impossible to create relative symlinks. Secondly, the conversion performed assumes that the link-target is specified relative to the current working directory, rather than the link directory. So if I pass the same arguments as I would for the raw SFTP command, the new symlink would likely be created with a broken target-path. This is particularly unhelpful when I simply want to copy a relative symlink from one machine to another, keeping the relative target unchanged.

I think this issue could be fixed by simply removing the line: `$target = $this->_realpath($target);`. This gives the expected behaviour for me. Would this be an acceptable change?

terrafrost:
Fixes #1004

### DIFF
--- a/phpseclib/Net/SFTP.php
+++ b/phpseclib/Net/SFTP.php
@@ -642,6 +642,21 @@ class Net_SFTP extends Net_SSH2
     }
 
     /**
+     * Returns canonicalized absolute pathname
+     *
+     * realpath() expands all symbolic links and resolves references to '/./', '/../' and extra '/' characters in the input
+     * path and returns the canonicalized absolute pathname.
+     *
+     * @param string $path
+     * @return mixed
+     * @access public
+     */
+    function realpath($path)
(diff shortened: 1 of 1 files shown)


terrafrost:
This has been merged into 1.0 / 2.0 / master.

PR #1018, opened 2016-08-02
Added root, pow, max, min, loopforeach in BigInteger

danog:
Added root, pow, max, min, loopforeach functions and tests in BigInteger and applied styleci fixes.

### DIFF
--- a/phpseclib/Math/BigInteger.php
+++ b/phpseclib/Math/BigInteger.php
@@ -3674,4 +3674,241 @@ class BigInteger
         // self::$base === 31
         return ($x - ($x % $y)) / $y;
     }
+
+    /**
+     * Calculates the nth root of a biginteger.
+     *
+     * Returns the nth root of a positive biginteger, where n defaults to 2
+     *
+     * Here's an example:
+     * <code>
+     * <?php
+     *    $a = new \phpseclib\Math\BigInteger('625');
+     *
+     *    $root = $a->root();
+     *
+     *    echo $root->toString(); // outputs 25
+     * ?>
+     * </code>
+     *
+     * @param \phpseclib\Math\BigInteger $n
+     * @access public
+     * @return \phpseclib\Math\BigInteger
+     *
+     * @internal This function is based off of {@link http://mathforum.org/library/drmath/view/52605.html this page} and {@link http://stackoverflow.com/questions/11242920/calculating-nth-root-with-bcmath-in-php this stackoverflow question}.
+     */
+    function root($n = null)
+    {
+        $one = new static(1);
+        $two = new static(2);
+        if ($n === null) {
+            $n = $two;
+        }
+        if ($n->compare($one) == -1) {
+            return new static(0);
+        } // we want positive exponents
+        if ($this->compare($one) == -1) {
+            return new static(0);
+        } // we want positive numbers
+        if ($this->compare($two) == -1) {
+            return $one;
+        } // n-th root of 1 or 2 is 1
+        $root = new static();
+        switch (MATH_BIGINTEGER_MODE) {
+            case self::MODE_BCMATH:
+                // g is our guess number
+                $g = 2;
+                // while (g^n < num) g=g*2
+                while (bccomp(bcpow($g, $n->value), $this->value) == -1) {
+                    $g = bcmul($g, '2');
+                }
+                // if (g^n==num) num is a power of 2, we're lucky, end of job
+                if (bccomp(bcpow($g, $n->value), $this->value) == 0) {
+                    $root->value = $g;
+                    break;
+                }
+
+                // if we're here num wasn't a power of 2 :(
+                $og = $g; // og means original guess and here is our upper bound
+                $g = bcdiv($g, '2'); // g is set to be our lower bound
+                $step = bcdiv(bcsub($og, $g), '2'); // step is the half of upper bound - lower bound
+                $g = bcadd($g, $step); // we start at lower bound + step , basically in the middle of our interval
+
+                // while step!=1
+
+                while (bccomp($step, '1') == 1) {
+                    $guess = bcpow($g, $n);
+                    $step = bcdiv($step, '2');
+                    $comp = bccomp($guess, $this->value); // compare our guess with real number
+                    if ($comp == -1) { // if guess is lower we add the new step
+                        $g = bcadd($g, $step);
+                    } elseif ($comp == 1) { // if guess is higher we sub the new step
+                        $g = bcsub($g, $step);
+                    } else { // if guess is exactly the num we're done, we return the value
+                        $root->value = $g;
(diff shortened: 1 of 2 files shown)


terrafrost:
Can you do the styleci fixes in another PR? I don't have a problem with them, per say, but they really ought to be in a separate commit and PR.

I will say that one thing the styleci fixes did I don't agree with. ie. They removed `@access private` from the phpdoc blocks. Those functions are implicitly public since no visibility keyword is explicitly used but they're not really intended for public use all the same, hence the `@access private` (which hides them from the default phpdoc generated documentation). They're not explicitly private because sometimes I've asked people to directly access otherwise private functions to assist me in debugging. If I had to ask them to go into the source and modify the visibility and then call those functions...  well that'd be more hoops people would have to jump through and the odds of them not jumping through those hoops would increase.

Also, Net/SCP.php does call "private" functions. ie. they're private in that they're not intended for third parties to access (unless they're told to by me) and when I say that the API is stable I'm referring to functions that have `@access public` - not `@access private` - so BC breaks with "private" functions isn't a concern of mine - just "public" ones. Altho in the case of SCP.php, I have been thinking I might delete that from the master branch, but still...

There's also stuff like this:

```php
        $block = ($shuffleip[$r        & 0xFF] & "\x80\x80\x80\x80\x80\x80\x80\x80") |
                 ($shuffleip[($r >>  8) & 0xFF] & "\x40\x40\x40\x40\x40\x40\x40\x40") |
                 ($shuffleip[($r >> 16) & 0xFF] & "\x20\x20\x20\x20\x20\x20\x20\x20") |
                 ($shuffleip[($r >> 24) & 0xFF] & "\x10\x10\x10\x10\x10\x10\x10\x10") |
```

It used to be like this:

        $block = ($shuffleip[ $r        & 0xFF] & "\x80\x80\x80\x80\x80\x80\x80\x80") |
                 ($shuffleip[($r >>  8) & 0xFF] & "\x40\x40\x40\x40\x40\x40\x40\x40") |
                 ($shuffleip[($r >> 16) & 0xFF] & "\x20\x20\x20\x20\x20\x20\x20\x20") |
                 ($shuffleip[($r >> 24) & 0xFF] & "\x10\x10\x10\x10\x10\x10\x10\x10") |

For one of those code blocks the square brackets are aligned. For the other (the one this PR produces) they're not.

I do like that the spaces in the middle are preserved but sometimes trailing (and preceding) spaces are intended.

Also, why are the `@package RSA` stuff being removed? I mean, I guess I'm not overly married to it but is that now deemed as being unnecessary or something?

And speaking of RSA, why are `use phpseclib\Crypt\AES` and `use phpseclib\Crypt\TripleDES` being removed?

Replacing all the `array()` instances with `[]` does raise the minimum PHP version required but for the master branch I was thinking about doing that anyway so I actually don't have a problem with that (square bracket syntax was introduced in 5.4 - phpseclib 2.0 is designed to work on 5.3; changing the minimum version reqs is a BC break, in-so-far as I'm concerned).

I'm actually ok with most of the changes, with the caveats that I've mentioned above. Well and with the caveat that I haven't reviewed everything yet. But the fact that I've spent this much time reviewing the CS changes without actually looking at what the title of your PR is about...  well that's just testament to the fact that the CS changes ought to be in a separate PR.

danog:
Ok, I'm gonna undo the styleci fixes and open a separate pr for them.

bantu:
I don't think having loopforeach makes sense.

Also, please squash the commits to remove the CS changes.

danog:
I've squashed the commits.  
I think that the loopforeach function is useful if you want to execute some commands n times without necessarily having to use the slow (compared to php's native ++ operator) add function.

terrafrost:
```php
function root($n = null)
{
    $one = new static(1);
    $two = new static(2);
```

In randomPrime those variables are declared static within the function to avoid having to reinitialize them each time:

https://github.com/phpseclib/phpseclib/blob/dd23d91d411d812b2765c5bc3b8bd43fe58be0a2/phpseclib/Math/BigInteger.php#L3125

But that said, it'd probably be better to make them static for the entire class instead..

Also, on PHP 5.6, internal mode and gmp mode give different answers:

```php
$root = new BigInteger(4);
$num = new BigInteger(300);

echo $num->root($root);
```

gmp gives 4 as the answer and internal mode gives 5. It should give the same result regardless of the mode being used.

I believe GMP uses Newton's Method to calculate n'th roots:

https://gmplib.org/manual/Nth-Root-Algorithm.html
https://en.wikipedia.org/wiki/Nth_root_algorithm

What I did when implementing a lot of the initial functions was to look at Java's implementation of BigInteger:

https://docs.oracle.com/javase/8/docs/api/java/math/BigInteger.html

Unfortunatley, that doesn't appear to have any method for finding the nth root. Neither does BigDecimal:

https://docs.oracle.com/javase/8/docs/api/java/math/BigDecimal.html

An n'th root algorithm is also not described in the Handbook of Applied Cryptography:

http://citeseer.ist.psu.edu/viewdoc/download?doi=10.1.1.99.2838&rep=rep1&type=pdf

Anyway, I have an idea on how to fix the issue with gmp / internal modes output not matching (that wouldn't require deleting `gmp_root` but I'll play around with that later.

terrafrost:
So I'm honestly a bit on the fence about adding a root() method given that, aside from gmp, I'm not aware of any biginteger implementations that use it.

That said, I made some changes to the commit:

https://github.com/terrafrost/phpseclib/tree/danog2

So gmp_root() seems to work by doing the floor() operation on the numbers. Like the 4th root of 3500 is 7.69160 but gmp returns 7. http://php.net/gmp-root basically says as much when it says "take the integer part of nth root"

Anyway, I rewrote this a bit. The bcmath and the pure-PHP versions were pretty much the same so I just removed the bcmath version and am having it use the pure-PHP version when GMP is present.

bitwise_leftShift() doesn't do this but bitwise_leftShit() is also a lot less complex with what it's doing.

I've also made min() and max() static. Like how random() and randomPrime() are in the master branch. But unlike those functions I've made it so that min() and max() can take an arbitrary number of parameters (eg. more than 2 but it could be less than two too lol)

As for loopforeach...  so you have two methods for incrementing / decrementing...  one increments using the regular arithmetic operations and the other increments using a BigInteger object. The limit could, in some cases, be improved upon by using floating point numbers. See `self::$maxDigit` in `__construct`. That said, I think it's fine, as is. There's room for improvement but it's not a show stopper either.

bantu:
@terrafrost I'd like to point out again that the loopforeach method does not make any sense to me. Executing more iterations of a loop than fit the int type is a niche application. It does not belong into the already 4K lines BigInteger class.

terrafrost:
@bantu - fair enough. I've removed it. If people want that functionality they can always extend the BigInteger class and add that functionality.

PR #1022, opened 2016-08-08
SSH-1.99 support and timeout conversion added

Thorbijoern:
I've added some lines of code which enshure that there is no connection/authentication failure when the host sends the SSH Version String with compatibility Version 1.99 and without Carriage Return (like "SSH-1.99-[...]\n").
SSH Version 2.x ("SSH-2.[...]\r\n") is still also supported.
I think the timeout shouln't be only some microseconds, so I added conversion of microseconds (microtime) into seconds. Maybe I'm wrong.

### DIFF
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -1026,7 +1026,7 @@ class SSH2
             }
             $elapsed = microtime(true) - $start;
 
-            $this->curTimeout-= $elapsed;
+            $this->curTimeout-= $elapsed / 1e6;
 
             if ($this->curTimeout <= 0) {
                 $this->is_timeout = true;
@@ -1062,7 +1062,7 @@ class SSH2
                         return false;
                     }
                     $elapsed = microtime(true) - $start;
-                    $this->curTimeout-= $elapsed;
+                    $this->curTimeout-= $elapsed / 1e6;
                 }
 
                 $temp = stream_get_line($this->fsock, 255, "\n");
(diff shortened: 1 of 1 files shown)


terrafrost:
I took an alternate approach for the 1.99 stuff:

https://github.com/phpseclib/phpseclib/commit/c20b661a4417cab29965581a5fabc4f42aba3f7d

For the microtime handling...  `microtime` returns an int whereas `microtime(true)` returns a float. So when a second has passed you're not subtracting 1e6 from `$this->curTimeout` - you're subtracting 1e-6. So by diving the elapsed time (in seconds with microseconds as a fractional amount) by 1e6 you're essentially making the timeout unlimited. Like normally it'd timeout if 30 seconds have passed but if you're going to divide that 30 by 1e6 then that 30 essentially becomes 0 and the default timeout ceases to be 30s but rather, almost a full year. Unless my analysis is incorrect?

Thorbijoern:
from [http://php.net/manual/en/function.microtime.php](http://php.net/manual/en/function.microtime.php):
>  By default, microtime() returns a string in the form "msec sec", where sec is the number of seconds since the Unix epoch (0:00:00 January 1,1970 GMT), and msec measures microseconds that have elapsed since sec and is also expressed in seconds.

> If get_as_float is set to TRUE, then microtime() returns a float, which represents the current time in seconds since the Unix epoch accurate to the nearest microsecond. 

`microtime` or `microtime(false)` returns a string, but `microtime(true)` returns a float in seconds with the microseconds as decimals i got that wrong, sorry.

Your solution of the 1.99 problem is good to.
On line 1097 in SSH2.php the version 1.99 is also acceptet.
Thanks

PR #1026, opened 2016-08-24
Handle empty fread return correctly when uploading with SFTP

mbardelmeijer:
The pull request #995 introduced an issue when zero-byte file uploading is required.
For instance, we upload a directory which contains a '.gitkeep' file. This file doesn't contain any data and so the filesize is 0 bytes.

The issue is that the `fread` method never returns false, only a empty string is returned for these types of files. This case should also be handled to exit out of the upload while loop.


terrafrost:
I'll review this this weekend. Thanks!

terrafrost:
I cherry picked this into the 1.0 branch and merged that into 2.0 and master:

https://github.com/phpseclib/phpseclib/commit/1eda62913aa4c3f0806b23e180fd7b022dbde1e8

Thanks!

PR #1036, opened 2016-09-20
Adding php 7.1 to travis

DeepDiver1975:


terrafrost:
A lot of the failures are because phpseclib uses mcrypt. I'm not going to remove those calls, however, or even disable mcrypt on PHP versions > 7.1. SSH2 cannot use PHP's bindings to OpenSSL but it can use PHP's bindings to mcrypt. This is because, among other things, mcrypt can propagate the state of the cipher from call to call whereas OpenSSL cannot. I mean, I'm sure the C version of OpenSSL can but the PHP bindings don't expose that functionality.

I can, however, use the error suppression operator (@) to make those errors go away. Which, right now, is what I'm thinking I'll do.

terrafrost:
I squashed your two commits and cherry-picked them into the 1.0 branch and got phpseclib working on PHP 7.1 without errors:

https://github.com/terrafrost/phpseclib/commit/640f10609fdb2aa825f8a42c6bee975879974e1c

In #1037 you mentioned "A non-numeric value encountered". That was fixed a while ago with this commit:

https://github.com/phpseclib/phpseclib/commit/dc7f7e4d859468ce2318fc1562738c4d887d9647

DeepDiver1975:
great - thx

terrafrost:
The fix has been merged.

Thanks!

Issue #1037, opened 2016-09-20
PHP 7.1 Support ....


DeepDiver1975:
Let's collect to dos for php 7.1

- 7.1 to travis shows some warnings - https://github.com/phpseclib/phpseclib/pull/1036
- 'A non-numeric value encountered' in https://github.com/phpseclib/phpseclib/blob/master/phpseclib/File/ASN1.php#L507

oanhnn:
Please support PHP 7.1
https://travis-ci.org/deployphp/deployer/jobs/162086438#L352

terrafrost:
See https://github.com/phpseclib/phpseclib/pull/1036#issuecomment-248585831

I want to do a PR, give people a chance to review it, and then merge it. I'm going to try to do that later today.

terrafrost:
phpseclib is now passing on Travis CI  with PHP 7.1:

https://travis-ci.org/phpseclib/phpseclib/builds/164359081

Thanks!

PR #1040, opened 2016-09-23
remove E_DEPRECATED notices from PHP 7.1

### LINKED ISSUE #1028: mcrypt is deprecated in PHP 7.1
I've been testing my projects with PHP 7.1 and came across this. It's just a deprecation notice, so easily silenced, but probably worth looking at sooner rather than later, as mcrypt will likely be removed from PHP in 7.2 at the earliest.

https://wiki.php.net/rfc/mcrypt-viking-funeral

terrafrost:
PHP 7.1 deprecated mcrypt per https://wiki.php.net/rfc/mcrypt-viking-funeral but mcrypt is still, none-the-less, useful for phpseclib because of deficiencies in the OpenSSL bindings PHP has.

Among other things, OpenSSL does not persist the state from function call to function call. So if you call encrypt('plaintext') twice, with the same plaintext, you'll get the same ciphertext each time. This is not how SSH operates or how phpseclib operates when you're in continuous mode. phpseclib is able to emulate state persistence with OpenSSL with block ciphers but it can't do this with stream ciphers like RC4.

Further, SSH / phpseclib support ciphers like arcfour256, which is RC4 with a 256-bit key. OpenSSL only supports RC4 with 40-bit, 64-bit and 128-bit keys (called rc4-40, rc4-64 and rc4 in OpenSSL parlance, respectively).

For these reasons I don't think simply refusing to use mcrypt for PHP >= 7.1 is ideal. But if it is used in PHP 7.1 E_DEPRECATED notices will appear. So, in light of this, I think just suppressing the errors with the `@` operator (as this PR does) is the best approach. 

This PR replaces #1036 and fixes #1028 / #1037

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ php:
   - 5.5
   - 5.6
   - 7.0
+  - 7.1
   - hhvm
 
 env:
@@ -27,7 +28,7 @@ install:
   - cd ..
   - eval `ssh-agent -s`
   - travis/setup-secure-shell.sh
-  - sh -c "if [ '$TRAVIS_PHP_VERSION' != 'hhvm' -a '$TRAVIS_PHP_VERSION' != '7.0' ]; then travis/install-php-extensions.sh; fi"
+  - sh -c "if [ '$TRAVIS_PHP_VERSION' != 'hhvm' -a '$TRAVIS_PHP_VERSION' != '7.0' -a '$TRAVIS_PHP_VERSION' != '7.1' ]; then travis/install-php-extensions.sh; fi"
   - travis/setup-composer.sh
 
 script:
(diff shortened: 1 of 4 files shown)


terrafrost:
I cannot figure out why this is failing.

https://travis-ci.org/terrafrost/phpseclib/builds/161539623
https://travis-ci.org/terrafrost/phpseclib/builds/162353763

Those two work just fine.

I do `git pull upstream 1.0` in the 1.0-php71 branch and get the following message:

```
 * branch            1.0        -> FETCH_HEAD
Already up-to-date.
```

The diff looks good too:

https://github.com/phpseclib/phpseclib/compare/1.0...terrafrost:1.0-php71

terrafrost:
https://github.com/phpseclib/phpseclib/pull/1040/commits/e81b8d90d3fa98d155c0e050f3d56b32affbcae7 seems to have fixed the unit test timeout errors. The timeout errors were inconsistent but even if there wasn't a timeout it was still taking a long time to run on PHP 7.1. Took 10+ minutes before the "Travis: PHP 7.1 tweaks" commit. Now it takes ~1 minute.

PR #1045, opened 2016-10-04
New SSH2 Batch hangling class

sasf54:
New SSH2_Batch class for executing commands on ssh in a batch fashion. Has some safety checks for not interrupt running commands.

### DIFF
--- a/Net/SSH2_Batch.php
+++ b/Net/SSH2_Batch.php
@@ -1,0 +1,143 @@
+<?php
+
+use phpseclib\Net\SSH2;
+
+/**
+ * Created: sasf54
+ * License: MIT
+ * Date: 2016.10.04
+ *
+ * SSH2_Bash: Will use an existing SSH2 connection, and will execute commands in a batch fashion.
+ * Will use single session unless SSH2::execute.
+ * Will wait for the commands to finish, unless a logout command is issued to close the ssh connection.
+ * You can send multiple commands at once, and this class will make sure, that it will stay connected until
+ * the commands are all executed / failed. Set timeout to an acceptable low value (0.1).
+ * This class will execute batched commands in order, and in whole. You only have the opportunity to check the output
+ * at the read() for errors.
+ *
+ * Example code:
+ * // initializing ssh2 connection
+ * $ssh2 = new SSH2($server_hostname, $server_port);
+ * if ( !$ssh2->login($username, $password)) {
+ *     echo var_export($ssh2, true);
+ *     return 'Was not able to login to the server';
+ * }
+ *
+ * $ssh2_bash = new SSH2_BASH($ssh2,false,0.1);
+ * // executing commands (as they are processed)
+ * $ssh2_bash->sshWrite("sleep 1");
+ * $ssh2_bash->sshWrite("echo \"finished\"");
+ * $ssh2_bash->sshWrite("exit");
+ * echo $ssh2_bash->read();
+ *
+ * $ssh2->->disconnect();
+ *
+ */
+class SSH2_Batch {
+    /**
+     * @var $ssh2 SSH2, connection to in the class
+     */
+    var $ssh;
+    /**
+     * @var bool $debug Do you want to enable debugging (will echo everything)
+     */
+    var $debug;
+
+    /**
+     * @var int $echo Does echo expected (0 for no, 1 is for yes)
+     */
+    var $echo = 1;
+
+    /**
+     * SiteManager constructor.
+     * @param $ssh2 SSH2 Existing ssh2 connection
+     * @param $debug boolean On debug, will echo everything (echoed input, and the output)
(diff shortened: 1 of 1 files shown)


terrafrost:
There are a number of coding standard issues. The class isn't namespaced, curly brackets aren't on new lines when they should be or, in some cases, aren't present at all (phpseclib basically uses [PSR2](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md) with a few exceptions as delineated at https://github.com/phpseclib/phpseclib/blob/master/build/code-sniffer-ruleset.xml) and the phpDoc comments aren't really compliant with that standard either.

That said, tbh, as useful as this class is, I think this is best left as a helper class that does not live in the phpseclib repo.

sasf54:
Hi,

Sorry for the not standard code formatting.
I was using / trying to use it on live environment for deploying a new sub
site on a remote box.
I was met PHPSecLib on sourceforge (where there's still the 1.x code).
I was trying to use as the documentation there. There are a lot of traps,
and the documentation not mentions explicitly :( :

- ssh session for separate commands
- When the execution ends (in time)
- where the execution ended (in output)
- High ping / waiting time, if you try to solve the above problems, by
reading per command.

(I've got some issues, with my local fully qualified test environment
running phpseclib, so I trashed it from my project :( (now using API)
host machine: amd64 with gentoo linux
guest machine(s): amd64 ubuntu 16.04 with PHP7.0
(php sec lib cannot make connection (I can make connection from ssh)
(it's working with the same settings on live site...))

Bests,
Ferenc

terrafrost:
1.0 is still actively maintained. It - and the 2.0 branch - had releases about two weeks ago and it's unit tested on Travis CI up to PHP 7.1.

phpseclib 2.0 works best with Composer altho I wouldn't go so far as to say that it requires it. But, none-the-less, that's why I'm not providing a downloadable zip file of that. If you really want one you can get it from github.com, open the tagged release, and download the zip file of that.

phpseclib dev-master (which'll become 3.0) will be Composer-only.

As for the documentation...  I would like to redo it but in lieu of volunteers actually stepping up I'll probably do it when 3.0 is nearing release (ie. not for a while). In the mean time...

> - ssh session for separate commands

http://phpseclib.sourceforge.net/ssh/intro.html

In that link it says "successive calls may not work as expected" and links to this:

http://phpseclib.sourceforge.net/ssh/examples.html#chdir

I'm not sure if that's what you're talking about?

> - When the execution ends (in time)

It's 10s by default but that's not documented anywhere. I can add that.

> - where the execution ended (in output)

That functionality does not exist within phpseclib but people using phpseclib can do `$start = microtime(true); $ssh->exec('do something'); $elapsed = microtime(true) - $start;`

> - High ping / waiting time, if you try to solve the above problems, by
reading per command.

I'm not sure what you mean?

> (php sec lib cannot make connection (I can make connection from ssh)

Without more information it's possible it's something like suexec preventing fsockopen from making socket connections or some such. But if you've already worked around it that works too!

sasf54:
Hi,


   - ssh session for separate commands

Yes, it's there, but I was skipped over it, until the problem occurred ...
Execution ends in time:
I meant, that if you are executing (write) a process, that usually runs for
a long time, and if you are closing the connection before it's ended, it
will be terminated by system, unless "nohup" is used at execution (>nohup
sleep 12).
in output: if you are writing up more than one command, (mysql import /
copy a lot of files), and than you are trying to read, (reading for double
to be sure), there's no guarantee, that all of your commands are executed:
 * You will have to analyze the output to be sure, for all commands are
finished.

High ping time:
The normal case, on executing more commands:
send the command by write, and do read for output (console is ready).
If you do this on a remote host, and you have 20 or more commands, it will
result in 40*pingtime+execution time...

connection problem:
I'm using Virtualbox (vagrant) for emulating the live system, it might be
some artifact with:
AMD64 + Virtualbox + virtualbox amd64 acceleration + ubuntu 16.04 amd64 +
php7 + ssh daemon
(however logging in from the remote host to itself works fine... (and maybe
needed to accept the ssh key from the server...))

About my ssh2 batch handling class:
I understand, it's dangerous, and not advised to use:
 - if you have more commands put it in a batch file,
    * Put checks in it to make sure it doesn't failed / fix it...
    * Minimizing the traffic
 - It's not so trivial to process output, as by command by command
 - If a command fails, all other input are sent regarding of result
   * worst example:
       > vagrant ssh    /*  FAILS, because we are in root directory */
       > rm -rf *           /*  Will delete the host environment, not the
guest's home / login-ed directory*/

(Why I created:
I was making a sub site, complex deploy, with new DB user, DB, DB import,
copying a lot of files, setting permissions, creating config files.
I found out, the hard way, that php mysqli tries to be too smart, and
returns before actually finishing the user, and db creation...)

I see the amount of work, and bug-chasing in the lib...


Bests,
Ferenc

PR #1050, opened 2016-10-19
Composer: fix autoloading for 1.0 branch

### LINKED ISSUE #1048: Autoloading in 1.* does not work
I have updated phpseclib from 0.* to 1.* (1.0.4 actually) in my Yii project and I had issues with autoloading.

After debugging I found that this is not valid PSR-4 entry because files under `phpseclib` are not namespaced.
```
    "autoload": {
        "files": [
            "phpseclib/bootstrap.php"
        ],
        "psr-4": {
            "phpseclib\\": "phpseclib/"
        }
    }
```

After adding in my root's `composer.json`:

```
    "autoload": {
        "psr-4": {
            "": "protected/vendor/phpseclib/phpseclib/phpseclib/"
        }
    }
```

files are loaded correctly.

PS. There is another issue with Yii, when loading via `Yii::import()` - files are not named like classes (`AES.php` -> `Crypt_AES`), but this is whole different story...

terrafrost:
Fixes #1048

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -66,11 +66,16 @@
     },
     "include-path": ["phpseclib/"],
     "autoload": {
+        "psr-0": {
+            "Crypt": "phpseclib/",
+            "File": "phpseclib/",
+            "Math": "phpseclib/",
+            "Net": "phpseclib/",
+            "System": "phpseclib/"
+        },
         "files": [
+            "phpseclib/Crypt/Random.php",
             "phpseclib/bootstrap.php"
-        ],
-        "psr-4": {
-            "phpseclib\\": "phpseclib/"
-        }
+        ]
     }
 }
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -66,11 +66,16 @@
...
     "autoload": {
+        "psr-0": {
+            "Crypt": "phpseclib/",
+            "File": "phpseclib/",
+            "Math": "phpseclib/",
+            "Net": "phpseclib/",
+            "System": "phpseclib/"
+        },
         "files": [
+            "phpseclib/Crypt/Random.php",
```
bantu: This should be below bootstrap.

bantu:
Missing composer.lock changes as requested in the comments?

bantu:
This is not a too big issue, though.

terrafrost:
> Missing composer.lock changes as requested in the comments?

Done.

PR #1056, opened 2016-11-02
Incorrect namespace hint

lbajsarowicz:
When trying to create SCP instance, IDE reports incorrect object type given. Fixed that by changing namespace in SCP class to correct one.


terrafrost:
I cherry-picked this into the 2.0 branch and then merged that into master.

Thanks!

PR #1061, opened 2016-11-17
RSA: refactor PKCS1/PKCS8 to facilitate re-use

terrafrost:
- PKCS1/PKCS8 now use File/ASN1 to decode stuff. this will facilitate re-use of these components by future DSA / ECDSA clases
- PKCS8 supports a ton more encryption algorithms than before
- a lot of the ASN1 definitions are done using arrays / define's. This makes these changes require PHP7. There's a high probability this will change with a later PR.
- the timeout for BigInteger::randomPrime and RSA::createKey has been removed. the real bottle neck is isPrime and timeout() doesn't work so well there (unless it were also added to modPow, etc).
  it doesn't work with BigInteger::MODE_GMP nor does it work with RSA::ENGINE_OPENSSL
- several RSA properties, which previously were set via `define()` calls, are now set via static setters. this means they can be changed dynamically without the use of runkit.
  Math/BigInteger, otoh, still uses define() for it's constants since mixing a gmp-based bigint object with a internal-based bigint object won't work whereas having two RSA objects
  with different engine's isn't a huge issue (just as it's not a huge issue with the symmetric ciphers)
- the various methods to retrieve the sizes of stuff were normalized. so most methods will assume size being set or retrieved is in bits unless method has "InBytes" at the end of it now. an "InBytes" method should only exist as a method if an (implicit) "in bits" version does as well
- other misc master-branch specific bugs have been fixed

One thing I considered was that having `$rsa->load()` and `RSA::createKey()` seems inconsistent. That it ought to be `$rsa->load()` and `RSA::create()` or `$rsa->loadKey()` / `$rsa::createKey()`. But, after thinking about it, I think what it's currently doing is best.

`$rsa->load()` is intended somewhat for parity with stuff like `$x509->load`, `$csr->load`, etc. But also, load() will auto detect what's being loaded. In the case of DSA, for example, you have keys and you have parameters. So would I do `$dsa->loadKey` and `$dsa->loadParameter` or would I have `$dsa->load` auto-recognize what's being loaded? I've opted to do the latter.

As for `RSA::create` vs `RSA::createKey`...  for DSA you can create parameters in addition to keys.  Thus there is the need for `DSA::createParameters()` and `DSA::createKey()`. I suppose I could do `DSA::create()` for `DSA::create()` and `DSA::Parameters::create()` for parameters but idk. For the time being I think the method I've chosen works. Maybe I'll change my mind later.

In theory the keys could be loaded in the constructor to make this a bit of a non-issue but X509 won't be doing that. In X509 the constructor will initialize a skeletal X509 array that can be modified. To "finalize" the skeletal structure you'd need to sign it but none-the-less the constructor wouldn't be used for stuff like that so for parity purposes it makes sense for RSA not to load keys in the constructor. And for RSA it doesn't make sense to create a skeletal structure in the constructor. You wouldn't ever set individual values for RSA anyway whereas for X509 you might want take a base X509, change the domain, sign it and get an X509 cert.


terrafrost:
> There's a high probability this will change with a later PR.

https://github.com/terrafrost/phpseclib/tree/asn1-revamp-pt1
https://github.com/terrafrost/phpseclib/commit/817c0a460a527566f02b41712f2f9ea149df824e

That stuff is in a different branch, tho, and will be submitted as a PR after this one is merged.

bantu:
Only made it halfway through and at the end also realised that some code I commented on was actually moved.

PR #1075, opened 2016-12-17
SSH2: make it so disabling PTY closes exec() channel if it's open

### LINKED ISSUE #1009: Unable to open channel - Consecutive Execs
I am writing a provisioning script to setup my server for me. Now everything seems to be working except it seems that the connection is being shut down early? 

```
ErrorException (E_USER_NOTICE)
Unable to open channel

```

 I am first connecting via SSH , then running lots of commands (50 or more) by running 

`run('do some linux stuff here');
`

Below is more code 

```
    public function ssh($ip)
    {
        $key = new RSA();
        $key->setPassword(env('SSH_KEY_PASSWORD'));
        $key->loadKey(file_get_contents('/home/vagrant/.ssh/id_rsa'));

        $ssh = new SSH2($ip);

        $ssh->enableQuietMode();

        if (!$ssh->login('root', $key)) {
            exit('Login Failed');
        }

        $this->session = $ssh;
    }
}

 public function run($command, $read = false)
    {
        $results = $this->session->exec($command.";");

        if(!empty($error = $this->session->getStdError())) {
            \Log::error($error);
        }

        if(!empty($results)) {
            \Log::info($results);
        }

        if($read) {
            return $this->session->read();
        }
    }


```

Is there a problem with how im doing it , is there a better way ?

terrafrost:
Fixes #1009

### DIFF
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -2601,6 +2601,11 @@ class Net_SSH2
             return false;
         }
 
+        if ($this->in_request_pty_exec) {
+            user_error('If you want to run multiple exec()\'s you will need to disable (and re-enable if appropriate) a PTY for each one.');
+            return false;
+        }
+
         // RFC4254 defines the (client) window size as "bytes the other party can send before it must wait for the window to
         // be adjusted".  0x7FFFFFFF is, at 2GB, the max size.  technically, it should probably be decremented, but,
         // honestly, if you're transferring more than 2GB, you probably shouldn't be using phpseclib, anyway.
@@ -3408,6 +3413,10 @@ class Net_SSH2
      */
     function disablePTY()
     {
+        if ($this->in_request_pty_exec) {
+            $this->_close_channel(NET_SSH2_CHANNEL_EXEC);
+            $this->in_request_pty_exec = false;
+        }
         $this->request_pty = false;
     }
 


CaptainConner:
I have downloaded the latest patch which should close channels after finishing an exec, however it appears to not be working as you can see here: https://gyazo.com/430818f39ac0bfab69b44904b6b294da

terrafrost:
I have no idea what that video is showing. It looks like, in the video, that you have 10+ connections to some server with phpseclib? Well they're not all with the same instance of phpseclib, I can tell you that much, because phpseclib only supports five channels max. Quoting SSH2.php (2.0 branch):

```php
    const CHANNEL_EXEC          = 0; // PuTTy uses 0x100
    const CHANNEL_SHELL         = 1;
    const CHANNEL_SUBSYSTEM     = 2;
    const CHANNEL_AGENT_FORWARD = 3;
```

SFTP.php adds support for one more channel:

```php
    const CHANNEL = 0x100;
```

This is something I plan on changing, in due time, in the master branch, but I'm working on other changes in the master branch currently.

What'd be much more helpful than a video are the SSH logs and some sample code that reproduces the problem. SSH logs can be obtained by doing `define('NET_SSH2_LOGGING', 2)` at the top and then `$ssh->getLog();` later on in the script.

That said, your post does make me realize that there's not a unit test for this change so I added one:

https://github.com/phpseclib/phpseclib/commit/9bccc01b973c91ec8c1f2b88aa9df13f0b7001d7

PR #1084, opened 2017-01-18
PHP5 updates for master branch

terrafrost:
- classes with static only methods are now abstract
- short array syntax is now used
- class variables / methods now have their visibility declared (public / private / protected)
- the "hi-optimized code" used in SymmetricKey.php (formerly Crypt/Base.php) is no longer cached. it's now it's own distinct engine type that can be tested distinct from the other internal mode. in the process the "lo-optimized code" has been deleted. previously, Crypt/Base.php would cache the "hi-optimized code" globally and once ten "hi-optimized" versions had been cached "lo-optimized" versions would be cached so as not to make PHP run out of memory. i view this caching as unnecessary. this also has the advantage that it makes the [proposed deprecation of `create_function`](https://wiki.php.net/rfc/deprecations_php_7_2) a non-issue since it's now using `eval()` to create [anonymous functions](http://php.net/manual/en/functions.anonymous.php). plus, [`eval` can't be disabled via php.ini](https://bugs.php.net/bug.php?id=62397), whereas `create_function` can be.

In a later PR I'll introduce variadic functions, which will replace `func_get_args()`, etc. I suppose there's no real distinct advantage to using variadic functions but the master branch already requires PHP 5.6 due to it's use of arrays as constants, so might as well more fully embrace PHP 5.6.

I have other big BC breaking changes planned that'll rely heavily on newer PHP features, but one step at a time!


phpseclib/Common/Functions/Strings.php
```diff
@@ -33,10 +33,27 @@ class Strings
...
+     * String Pop
+     *
+     * Inspired by array_pop
+     *
+     * @param string $string
+     * @param int $index
+     * @access public
+     * @return string
+     */
+    public static function pop(&$string, $index = 1)
```
bantu: This is never called? Tests?

phpseclib/Common/Functions/Objects.php
```diff
@@ -0,0 +1,75 @@
...
+
+namespace phpseclib\Common\Functions;
+
+/**
+ * Common Object Functions
+ *
+ * @package Functions\Objects
+ * @author  Jim Wigginton <terrafrost@php.net>
+ */
+abstract class Objects
```
bantu: What is the purpose of this class? This looks very fishy. Using Reflection outside of tests usually indicates that there is something wrong with the architecture. And it looks like this provides the ability to access properties and call methods that are declared non-accessible.

bantu: This of course a result of not properly designing an API and should not be fixed via hacking around using Reflection.

phpseclib/Common/Functions/Objects.php
```diff
@@ -0,0 +1,75 @@
...
+{
+    /**
+     * Accesses a private variable from an object
+     *
+     * @param Object $obj
+     * @param string $var
+     * @return mixed
+     * @access private
```
bantu: Does not match. Remove?

phpseclib/Common/Functions/Strings.php
terrafrost: It's called. eg.

https://github.com/terrafrost/phpseclib/blob/a25c8411087b1bf3531c0aab919470252b41147a/phpseclib/Crypt/Common/SymmetricKey.php#L878

phpseclib/Common/Functions/Objects.php
terrafrost: RSA.php has been significantly redesigned for the master branch. At this point, probably any change I make to RSA.php in the 1.0 / 2.0 branches will result in a merge conflict in the master branch. That is how significant the changes to RSA.php are.

SSH2.php is due to get some significant changes too but it has not had them. Right now, SSH2.php, as it currently exists in master, is pretty much 2.0's SSH2.php, which, in turn, is pretty much 1.0's SSH2.php, just namespaced. ie. a change made 1.0 will probably /not/ result in a merge conflict in the master branch's SSH2.php.

And this visibility changing stuff is used primarily by SSH2.php.

There are several alternatives I could be doing.

1. I could change the visibility only after every /big/ change I have planned has been made.
2. I could change the visibility to the files as big changes are being made to them. If I did that then RSA.php would have the visibility set but SSH2.php would not.
3. I could just delete the Objects class, retain the visibility changes, let the unit tests fail and just not care because the master branch isn't production ready anyway.

The option I've somewhat arbitrarily chosen is to update the visibility all at once and use hackish approach to get the unit tests working for the time being until I have had a chance to more comprehensively revamp those files.

terrafrost: But that said, it is not immediately obvious to me how to best deal with the problem. Maybe when I'm neck deep in SSH2.php I'll come up with a better idea but right now I'm almost finished with my DSA implementation. I want to do some more BigInteger changes and then do ECDSA and EdDSA. And then more ASN1.php changes and then X509.php changes, etc.

But as for why it's not immediately obvious to me...

SCP.php calls `get_channel_packet` in SSH2.php. That is NOT  a method people should normally be calling. That's why `get_channel_packet` is declared private. BUT SCP.php is not a regular file.

System/SSH/Agent.php calls `get_channel_packet` as well. ie. SSH2.php isn't setting SSH forwarding up - System/SSH/Agent.php is.

Now, maybe all of this will be changed when I get around to making the big changes for SSH2.php but until then...  these are functions that /regular/ applications ought not be calling but that some other classes within phpseclib /do/ have a current need to call. In C++ this is where you'd use `friend` but PHP doesn't have anything similar to that.

So what would you recommend in that scenario?

phpseclib/Common/Functions/Objects.php
terrafrost: I've updated the phpdoc lines. Is `@access ...` needed when visibility is declared in the function declaration? I figure the phpdoc blocks are in pretty bad state and at some later date I figure I'll make a PR dedicated to fixing them.

phpseclib/Common/Functions/Objects.php
bantu: Okay. Thanks for the detailed explanation. I am fine with this if it is only a temporary solution. I agree that this is better than having unit tests fail permanently. An alternative would simply having properties and methods public, as you mentioned.

> So what would you recommend in that scenario?

In the long run, I would try to keep things simple only using the tools provided by PHP. That is, if `get_channel_packet` needs to be called on an instance of SSH2, `get_channel_packet` needs to be defined public.
The fact that SCP and SSH Agent need to call get_channel_packet indicates that the architecture might be suboptimal. The same problem would probably also apply to SFTP, but that is worked around by having SFTP extend SSH2. More concretely, maybe there should be a SSHChannel class that can be used by SSH2, SCP, Agent, SFTP, etc. and provides access to the necessary channel mechanisms.

tests/PhpseclibTestCase.php
```diff
@@ -101,4 +101,20 @@ protected static function reRequireFile($filename)
             }
         }
     }
+
+    protected static function getVar($obj, $var)
```
bantu: Copy of methods from Objects class. Probably not too important, though.

bantu:
`@access` is doc block is not required by API generator. I think in conclusion this means that we should not have any of them for 3.0.

phpseclib/Common/Functions/Strings.php
terrafrost: When I view the diff on github.com changes to that file are collapsed by default:

screenshot

Could be why you're not seeing it?

bantu:
Indeed

terrafrost:
> `@access` is doc block is not required by API generator. I think in conclusion this means that we should not have any of them for 3.0.

Sounds good! I'll do another PR to remove it at some later date. Don't want to do it as a commit to this PR because this PR does enough stuff as is and I want to get onto more exciting dev things lol.

Issue #1103, opened 2017-03-07
Namespace issue when assigning several domains to certificate


kAlvaro:
If you try to create a certificate for several domains following the [recommended method](http://phpseclib.sourceforge.net/x509/guide.html#domains) you get a warning:

```
<?php

require_once __DIR__ . '/../vendor/autoload.php';

use phpseclib\Crypt\RSA;
use phpseclib\File\X509;

$keyGenerator = new RSA();
$keys = $keyGenerator->createKey(2048);

$privateKey = new RSA();
$privateKey->loadKey($keys['privatekey']);

$publicKey = new RSA();
$publicKey->loadKey($keys['publickey']);
$publicKey->setPublicKey();

$subject = new X509();
$subject->setDomain('example.com', 'example.net');

$subject->setPublicKey($publicKey);

$issuer = new X509();
$issuer->setPrivateKey($privateKey);
$issuer->setDN($subject->getDN());

$x509 = new X509();
$x509->sign($issuer, $subject);
```

> Warning: array_map() expects parameter 1 to be a valid callback, class 'X509' not found in C:\test\vendor\phpseclib\phpseclib\phpseclib\File\X509.php on line 3467

I'm using phpseclib/2.0.4 on PHP/5.6.21

metaclassing:
Well, I would not consider myself an expert in the use of array_map but I took a stab at fixing it:
https://github.com/metaclassing/phpseclib/blob/fix-array-map-namespace/phpseclib/File/X509.php#L3466-L3468
if this isn't too hateful of a fix then i can send a PR I guess.

terrafrost:
Fixed by #1104

PR #1104, opened 2017-03-07
namespacing the array_map function

metaclassing:
It would seem that issue https://github.com/phpseclib/phpseclib/issues/1103 array_map would prefer a fully qualified namespace.


terrafrost:
I added a unit test for this issue as well:

https://github.com/phpseclib/phpseclib/commit/91a0866fdced1d7b5741b31f0ef2556dfce1d6a5

Thanks!

PR #1111, opened 2017-04-03
Fixed ASN.1 UTCTime parsing

### LINKED ISSUE #1110: ASN.1 parser fails on UTCTime without seconds
A UTCTime like "1401070000Z" is not parsed correctly.

maurice2k:
Fixes #1110 
Added a unit test for the fix.


terrafrost:
Good catch on the int casting.

I tried it on PHP 5.5.15 and 7.1.0 without casting the stuff to int's and got this error:

```
Warning: gmmktime() expects parameter 3 to be integer, string given
```

Weird that it'd be the "$second" parameter (which is actually the third param lol) that'd cause issues. Like all of the other params can be strings but not the third parameter!

Anyway, I cherry picked this into 1.0 and merged that into 2.0 and master.

PR #1115, opened 2017-04-17
Refactor create_function() into anonymous function

langemeijer:
This increases code readability


terrafrost:
I did that for SymmetricKey.php with #1084. X509 is due for a major rewrite but right now I'm working on refactoring large swaths of BigInteger.php.

Anyway, thanks for the PR - it's been merged!

PR #1137, opened 2017-06-02
SFTP: Add a NO_REALPATH option to put()

dhzavann:
A third party we communicate with apparently requires us to use an initial double slash in the path when uploading files, for reasons unknown to me. Never the less, using just one slash does not appear to work.

Since the internal realpath() method strips all redundant slashes we need to bypass it.

### DIFF
--- a/phpseclib/Net/SFTP.php
+++ b/phpseclib/Net/SFTP.php
@@ -162,7 +162,7 @@ class SFTP extends SSH2
      * Current working directory
      *
      * @var string
-     * @see self::_realpath()
+     * @see self::realpath()
      * @see self::chdir()
      * @access private
      */
@@ -1282,7 +1282,7 @@ class SFTP extends SSH2
     /**
      * Returns general information about a file or symbolic link
      *
-     * Determines information without calling \phpseclib\Net\SFTP::_realpath().
+     * Determines information without calling \phpseclib\Net\SFTP::realpath().
      * The second parameter can be either NET_SFTP_STAT or NET_SFTP_LSTAT.
      *
      * @param string $filename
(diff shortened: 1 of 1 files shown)


terrafrost:
I think the changing of the constants from 1 to 0x1 was...  unnecessary. I mean, I get why you're doing it - it could potentially make it more clear that it's serving as a bitmask.

But regardless, the changes otherwise look good. Well, it seems like get perhaps ought to have the same method, but I went ahead and took care of that - thanks!

terrafrost:
This has been merged BTW. I cherry-picked the commit into the 1.0 branch and merged it up to 2.0 and master:

https://github.com/terrafrost/phpseclib/commit/5979ed571f54e6f5bc63ce8e468145ea08388418

terrafrost:
Actually, scratch the above. I did the merge on my local fork but...  `gett` doesn't have a `$mode` parameter. I could add one but it seems like this might be better if it was a global thing? eg. `->disableRealpath()` and `->enableRealpath()`, with the latter being the default option?

dhzavann:
Thank you for your review.

How does this look? I decided to go for "enablePathCanonicalization" instead of disableRealpath as I thought that to be a bit more explanatory, but that can be changed of course.

terrafrost:
I like that a lot better! Imma get this in the 1.0 and 2.0 as well. I'll try to have this done this evening.

Thanks!!

terrafrost:
I rebased your commits into one, changed the name of the first commit and then cherry-picked it from 1.0:

https://github.com/phpseclib/phpseclib/commit/bbf467babde690cf9f074a916d11313f677b2606

From there I merged into 2.0 and master.

PR #1138, opened 2017-06-02
Fix Invalid HMAC for D-link DES-3028

lockland:
When I tried connect to a dlink switch the Net_SSH2 does not choose
the correct algorithms for hmac causing an Invalid HMAC error. However,
after I commented the line 1280 everything work magnificently

### DIFF
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -1289,14 +1289,27 @@ class Net_SSH2
             //'zlib' // OPTIONAL        ZLIB (LZ77) compression
         );
 
+        $removeItemsFromArray = function ($main, $itemsToRemove) {
+            return array_values(array_diff($main, $itemsToRemove));
+        };
+
         // some SSH servers have buggy implementations of some of the above algorithms
+        $problematicAlgorithms = array('hmac-sha1-96', 'hmac-md5-96');
         switch ($this->server_identifier) {
             case 'SSH-2.0-SSHD':
-                $mac_algorithms = array_values(array_diff(
+                $mac_algorithms = $removeItemsFromArray(
                     $mac_algorithms,
-                    array('hmac-sha1-96', 'hmac-md5-96')
-                ));
+                    $problematicAlgorithms
+                );
+                break;
+            case strpos($this->server_identifier, 'SSH-2.0-DLINK') !== false:
+                $mac_algorithms = $removeItemsFromArray(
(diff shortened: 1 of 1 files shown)


terrafrost:
Out of curiosity, what's the full server identifier? I ask because it seems like this might be a simpler implementation:

```php
        // some SSH servers have buggy implementations of some of the above algorithms
        switch ($this->server_identifier) {
            case 'SSH-2.0-SSHD':
            case 'SSH-2.0-DLINK':
                $mac_algorithms = array_values(array_diff(
                    $mac_algorithms,
                    array('hmac-sha1-96', 'hmac-md5-96')
                ));
        }
```

But if SSH-2.0-DLINK was just a part of the server identifier than maybe this would be better:

```php
        // some SSH servers have buggy implementations of some of the above algorithms
        switch (true) {
            case $this->server_identifier == 'SSH-2.0-SSHD':
            case strpos($this->server_identifier, 'SSH-2.0-DLINK') !== false:
                $mac_algorithms = array_values(array_diff(
                    $mac_algorithms,
                    array('hmac-sha1-96', 'hmac-md5-96')
                ));
        }
```

lockland:
Hey Terra,
The full identifier is "SSH-2.0-DLINK Corp. SSH server ver 1.00.000".

I put a check for just a part of a string because I do not have other dlink model to test and this model was the only one showing an error.

I did not put switch(true) because it is not so clear for me, in that case I would prefer using ifs instead, on the other hand with switch statement we can write less code than using ifs.

I guess it is not a problem and the adaption is so easy.

terrafrost:
> I did not put switch(true) because it is not so clear for me, in that case I would prefer using ifs instead, on the other hand with switch statement we can write less code than using ifs.

It's totally subjective lol but personally I prefer the switch approach cause it's more concise.

Anyway, I'll go with that and try to have it merged by the end of the day.

Thanks!!

lockland:
Of Course It is, because of it I sad that the statement is not so clear for me. kkkkkk

So thank you for reply and for your excellent job with this amazing lib. I am using it for about 2 year and it was the only time I got a error even on version 0.3.10 that was the version I was using until last 2 weeks.

terrafrost:
I modified your commit a little bit and it's in the phpseclib repo:

https://github.com/phpseclib/phpseclib/commit/4d36e67beea147ee7d30976bebc4fee27daaddfa

Thanks!

PR #1143, opened 2017-06-27
BigInteger: modularize / modernize class and add new powmod engines

terrafrost:
On phpseclib 1.0 / 2.0 RunKit is needed to test the various different modes. You also cannot test 32-bit integers on a 64-bit system. With this PR you can.

If you want to use GMP or BCMath or whatever you can do this:

```php
BigInteger::setEngine('GMP')
```

If the engine you selected isn't a valid engine an exception will be thrown. This is in contrast to how things work with the various symmetric ciphers, which have `setPreferredEngine`. `setPreferredEngine` was good for 1.0, where exceptions weren't really an option, but with phpseclib 3.0 requiring 5.6+, exceptions now are an option. I'll update the symmetric cipher stuff later to work in a manner more consistent with BigInteger.

Also, in 1.0 / 2.0, if you want to prevent OpenSSL from being used you'd have to do `define('MATH_BIGINTEGER_OPENSSL_DISABLE', true);`. That now longer needs to be done. If you want to enable / disable OpenSSL you can do this:

```php
BigInteger::setEngine('PHP64', ['OpenSSL']);
BigInteger::setEngine('PHP64'); // does not use OpenSSL
```

The code that each "engine" runs is also not decided by switch statements any longer - engines live in their own files.

If you do not explicitly specify an engine the best engine will be chosen for you at runtime.

Two new modular exponentiation engines have been added as well.

- **EvalBarrett for PHP**
  With 64-bit integers with `['PHP64', 'Reductions\Barrett']` mode (ie. what phpseclib used to use for PHP mode) it takes **2.6** seconds to do the test at http://phpseclib.sourceforge.net/math/speed.phps (adapted for [phpseclib 3.0](https://pastebin.com/eY5WCMx6)). With `['PHP64', 'Reductions\Barrett']` mode it takes **1.7** seconds. That's a **35%** speed increase. It was inspired from [petrich's custom runtime generated PHP encryption stuff](https://github.com/phpseclib/phpseclib/pull/65). ie. loop unrolling, array dereferencing and also avoiding a bunch of if conditions (eg. if base == 26 do this)
- **Barrett for BCMath**
  With `['BCMath', 'BuiltIn']`, `bcpowmod` is used (ie. what phpseclib used to use for BCMath mode) it takes **8.6**s to do the above test. With `['BCMath', 'Reductions\Barrett']` it takes **2.3**s. That's a **74%** speed increase. `['PHP64', 'Reductions\EvalBarrett']` is still faster, at 1.7s, but 2.3s is faster than `['PHP32', 'Reductions\EvalBarrett']`. `['BCMath', 'Reductions\EvalBarrett']` does not provide nearly the same speed boost that `['PHP64', 'Reductions\EvalBarrett']` does.

Previously in BigInteger.php there were other modes that were implemented but inaccessible (since there wasn't a way to tell it which modes you wanted to use). Those modes now are accessible:

```php
'PHP64', ['Reductions\Montgomery']
'PHP64', ['Reductions\MontgomeryMult']
'PHP64', ['Reductions\Classic']
```

These modes are slower than the default modes and exist more so that other developers can further improve upon them if they're so inclined.


terrafrost:
> If the engine you selected isn't a valid engine an exception will be thrown. This is in contrast to how things work with the various symmetric ciphers, which have `setPreferredEngine`. `setPreferredEngine` was good for 1.0, where exceptions weren't really an option, but with phpseclib 3.0 requiring 5.6+, exceptions now are an option. I'll update the symmetric cipher stuff later to work in a manner more consistent with BigInteger.

Thinking about it, I'm not actually so sure that it's a good idea to have a publicly accessible `SymmetricKey::setEngine()` method. With `BigInteger` the engine is, by and large, immutable. Once an engine is set there's no reason it'd ever need to change for all practical applications. But that's not the case for `SymmetricKey`. With that one the engine could change depending on any number of parameters. Consider Rijndael for example:

```php
$cipher->setKey('...'); // block size is 128 by default so OpenSSL engine is used
$cipher->setBlockLength(256); // at this point either mcrypt or pure-PHP engine has to be used
```

So let's say you did `$cipher->setEngine(SymmetricCipher::ENGINE_OPENSSL)` before you did `setKey` or `setBlockLength`. Is it then supposed to throw a `BadConfigurationException` exception whilst in `setBlockLength`? That would be...  weird.

So, altho I acknowledge that `SymmetricCipher::setPreferredEngine` and `BigInteger::setEngine` are inconsistent, I'm gonna let it stand.

I did, however, change the parameters `setPreferredEngine` accepts. eg. previously, if you wanted to set the preferredengine to mcrypt you'd have to do this: `setPreferredEngine(SymmetricCipher::ENGINE_MCRYPT)`. I've made it so that you can do this instead: `setPreferredEngine('mcrypt')`. This is more consistent with `BigInteger`'s approach:

https://github.com/terrafrost/phpseclib/commit/4171262b9ee6f905eb2bfc4a942c5f44483efdf8

PR #1151, opened 2017-07-20
Bugfixes

rrran:
Some oblivious bugfixes and typos

### DIFF
--- a/phpseclib/Crypt/TripleDES.php
+++ b/phpseclib/Crypt/TripleDES.php
@@ -258,6 +258,7 @@ class TripleDES extends DES
         switch (strlen($key)) {
             case 16:
                 $key.= substr($key, 0, 8);
+                break;
             case 24:
                 break;
             default:
@@ -438,7 +439,6 @@ class TripleDES extends DES
      * @see \phpseclib\Crypt\Common\SymmetricKey::setPreferredEngine()
      * @param int $engine
      * @access public
-     * @return int
      */
     public function setPreferredEngine($engine)
     {
@@ -448,6 +448,6 @@ class TripleDES extends DES
             $this->des[2]->setPreferredEngine($engine);
         }
 
-        return parent::setPreferredEngine($engine);
+        parent::setPreferredEngine($engine);
     }
 }
--- a/phpseclib/Net/SSH1.php
+++ b/phpseclib/Net/SSH1.php
@@ -547,7 +547,7 @@ class SSH1
     {
         $this->fsock = @fsockopen($this->host, $this->port, $errno, $errstr, $this->connectionTimeout);
         if (!$this->fsock) {
-            throw new \RuntimeException(rtrim("Cannot connect to $host. Error $errno. $errstr"));
+            throw new \RuntimeException(rtrim("Cannot connect to $this->host. Error $errno. $errstr"));
         }
 
         $this->server_identification = $init_line = fgets($this->fsock, 255);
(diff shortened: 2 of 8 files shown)


phpseclib/Net/SSH1.php
```diff
@@ -547,7 +547,7 @@ private function connect()
     {
         $this->fsock = @fsockopen($this->host, $this->port, $errno, $errstr, $this->connectionTimeout);
         if (!$this->fsock) {
-            throw new \RuntimeException(rtrim("Cannot connect to $host. Error $errno. $errstr"));
+            throw new \RuntimeException(rtrim("Cannot connect to $this->host. Error $errno. $errstr"));
```
terrafrost: Interesting. It looks like this is only an issue in the master branch.

When I get around to revamping the API for SSH I will probably be deleting SSH1.php FWIW but, in the interim, I'm ok with having the issue fixed. Nice find!

phpseclib/Net/SSH1.php
```diff
@@ -924,7 +924,7 @@ public function write($cmd)
      * @throws \RuntimeException on connection error
      * @access public
      */
-    public function read($expect, $mode = self::READ__SIMPLE)
+    public function read($expect, $mode = self::READ_SIMPLE)
```
terrafrost: This is an issue in the 2.0 branch as well and ought to be fixed there as well.

phpseclib/Net/SSH1.php
```diff
@@ -936,7 +936,7 @@ public function read($expect, $mode = self::READ__SIMPLE)
 
         $match = $expect;
         while (true) {
-            if ($mode == self::READ__REGEX) {
+            if ($mode == self::READ_REGEX) {
```
terrafrost: This is an issue in the 2.0 branch as well and ought to be fixed there as well.

phpseclib/Crypt/TripleDES.php
```diff
@@ -258,6 +258,7 @@ public function setKey($key)
         switch (strlen($key)) {
             case 16:
                 $key.= substr($key, 0, 8);
+                break;
             case 24:
                 break;
             default:
```
terrafrost: There is no need for this change. Before the change we have this:

```php
        switch (strlen($key)) {
            case 16:
                $key.= substr($key, 0, 8);
            case 24:
                break;
```

After the change we have this:

```php
        switch (strlen($key)) {
            case 16:
                $key.= substr($key, 0, 8);
                break;
            case 24:
                break;
```

They do the same thing. When `strlen($key) == 16` `$key` will be modified and then fallthrough to the `break` for `strlen($key) == 24` case.

https://blog.mariusschulz.com/2015/06/25/the-curious-case-of-switch-statements-in-javascript elaborates. I suppose, as that article suggests, putting `// fallthrough` in comments might not be a bad idea but, in this case, doing `break;` would almost be preferable to that because, despite being redundant, it takes up less bytes. I suppose being concerned about bytes on modern computers is a bit silly tho idk lol

terrafrost:
Most of the changes look good but a few don't. I've commented on the few that don't look good.

Your PR reveals the need for a few unit tests. In particular, for DSA / SSH2 signatures (the issues you found were legit) and for ANSI escape codes (the issues you found there were not legit).

The biggest issues were with the addition of `break`'s. One of the `break`'s you added was for the ANSI escape codes (your change does reveal the need for a unit test) and the other `break` you added...  altho it didn't break anything it's also unnecessary (as elaborated in the comments). Unnecessary tho it is that one, it could be argued, would be good for overall improved readability. I discuss further in the comment for that break.

phpseclib/Crypt/TripleDES.php
rrran: Thanks for the comment. I suspected, but was not sure if it's a correct behavior

terrafrost:
I backported the SSH1 fixes to the 2.0 branch and, in writing the unit test for the DSA / SSH2 signature issues I found an additional issue that I fixed:

https://github.com/phpseclib/phpseclib/commit/4a7d3d27ba553acf289316778f3d6ed0b952af8a

Thanks!

Issue #1125, opened 2017-05-11
[Net/SSH2] Connection closed prematurely when i try to connect to my ssh server


LoickN:
Hello,

I try to use phpseclib to upload a file from my php server to an other server (witch is my personal computer).

When I try to upload a file from my php server in my computer to my ssh server in my computer, it works well but when I use exactly the same code from my prod server, I have the following problem :

`Net\SSH2.php:3702: Notice: Connection closed prematurely [user_error] `

Here is the code I use :

`<?php
    include('Net/SSH2.php');
	define('NET_SSH2_LOGGING', NET_SSH2_LOG_COMPLEX);
    //Send file via sftp to server
    echo "begin";

    echo "now connecting...";

    $ssh = new Net_SSH2($MyIP); //server redacted
    $ssh->login('userName', 'password'); //username and password redacted
    
    echo $ssh->exec('pwd');
    echo $ssh->exec('ls -la');
	echo $ssh->getLog();

?>`

It is the example code, that's why i do not understand why it does not work.

bantu:
Is there any log output?

LoickN:
Yes, I send it now ;)

`<-
00000000  53:53:48:2d:32:2e:30:2d:4f:70:65:6e:53:53:48:5f  SSH-2.0-OpenSSH_
00000010  37:2e:35:0d:0a                                   7.5..

->
00000000  53:53:48:2d:32:2e:30:2d:70:68:70:73:65:63:6c:69  SSH-2.0-phpsecli
00000010  62:5f:31:2e:30:20:28:6d:63:72:79:70:74:2c:20:62  b_1.0 (mcrypt, b
00000020  63:6d:61:74:68:29:0d:0a                          cmath)..

<- NET_SSH2_MSG_KEXINIT (since last: 1, network: 0s)
00000000  e6:56:18:03:86:d4:6a:f2:c0:82:41:af:ed:e7:07:05  .V....j...A.....
00000010  00:00:01:02:63:75:72:76:65:32:35:35:31:39:2d:73  ....curve25519-s
00000020  68:61:32:35:36:2c:63:75:72:76:65:32:35:35:31:39  ha256,curve25519
00000030  2d:73:68:61:32:35:36:40:6c:69:62:73:73:68:2e:6f  -sha256@libssh.o
00000040  72:67:2c:65:63:64:68:2d:73:68:61:32:2d:6e:69:73  rg,ecdh-sha2-nis
[...]

Issue #992, opened 2016-05-24
Login Failed - Unable to connect


AVAW:
I am trying to connect with public key:

```
include('Net/SSH2.php');
include('Crypt/RSA.php');

$ssh = new Net_SSH2('amazon', 22);
$key = new Crypt_RSA();
$key->loadKey(file_get_contents('key'));

if (!$ssh->login('login', $key)) {
    exit('Login Failed');
}

echo $ssh->exec('new_convertedW');
echo $ssh->exec('ls -la');
```

but i get 'Login Failed'. The logs of the connection:

```
<-
00000000  53:53:48:2d:32:2e:30:2d:24:7b:50:4f:4d:2e:41:52  SSH-2.0-${POM.AR
00000010  54:49:46:41:43:54:49:44:7d:2d:24:7b:50:4f:4d:2e  TIFACTID}-${POM.
00000020  56:45:52:53:49:4f:4e:7d:0d:0a                    VERSION}..

->
00000000  53:53:48:2d:32:2e:30:2d:70:68:70:73:65:63:6c:69  SSH-2.0-phpsecli
00000010  62:5f:31:2e:30:20:28:6f:70:65:6e:73:73:6c:2c:20  b_1.0 (openssl, 
00000020  62:63:6d:61:74:68:29:0d:0a                       bcmath)..

<- NET_SSH2_MSG_KEXINIT (since last: 0.2548, network: 0.0696s)
00000000  fd:95:4b:92:87:26:c4:56:a3:c1:2d:5d:d2:43:d7:8b  ..K..&.V..-].C..
00000010  00:00:00:b7:64:69:66:66:69:65:2d:68:65:6c:6c:6d  ....diffie-hellm
00000020  61:6e:2d:67:72:6f:75:70:2d:65:78:63:68:61:6e:67  an-group-exchang
00000030  65:2d:73:68:61:32:35:36:2c:64:69:66:66:69:65:2d  e-sha256,diffie-
00000040  68:65:6c:6c:6d:61:6e:2d:67:72:6f:75:70:2d:65:78  hellman-group-ex
00000050  63:68:61:6e:67:65:2d:73:68:61:31:2c:65:63:64:68  change-sha1,ecdh
00000060  2d:73:68:61:32:2d:6e:69:73:74:70:32:35:36:2c:65  -sha2-nistp256,e
00000070  63:64:68:2d:73:68:61:32:2d:6e:69:73:74:70:33:38  cdh-sha2-nistp38
[...]

PR #1156, opened 2017-08-03
Allow $encryption_algorithms to be configured

ricbra:
Given the amount of issues and bug reports where commenting out the `arcfour256` and `arcfour128` encryption algorithms seems to be the only solution it would make sense to make this variable configurable.

We're talking about this change:

```php
$encryption_algorithms = array(
    // from <http://tools.ietf.org/html/rfc4345#section-4>:
    //'arcfour256',
    //'arcfour128',
);
```

Just a few issues where this is being promoted:

- https://github.com/phpseclib/phpseclib/issues/1125 
- http://www.frostjedi.com/phpbb3/viewtopic.php?f=46&t=168945 
- https://github.com/phpseclib/phpseclib/issues/992

As (hopefully) the majority of PHP users is using composer nowadays, commenting out a piece of code isn't a valid solution anymore. By the upgrade to FreeBSD 11 and proftpd 1.3.6 we were also hit by a bug caused by these algorithms. Would be nice if this could be merged.

I've adapted the code style of the class as much as possible and didn't use visibility for methods and properties.

### DIFF
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -871,6 +871,49 @@ class SSH2
     var $agent;
 
     /**
+     * Contains all possible algorithms
+     *
+     * @see self::_key_exchange()
+     * @var array
+     * @access private
+     */
+    var $encryption_algorithms = array(
+        // from <http://tools.ietf.org/html/rfc4345#section-4>:
+        'arcfour256',
+        'arcfour128',
+
+        //'arcfour',      // OPTIONAL          the ARCFOUR stream cipher with a 128-bit key
+
+        // CTR modes from <http://tools.ietf.org/html/rfc4344#section-4>:
+        'aes128-ctr',     // RECOMMENDED       AES (Rijndael) in SDCTR mode, with 128-bit key
+        'aes192-ctr',     // RECOMMENDED       AES with 192-bit key
+        'aes256-ctr',     // RECOMMENDED       AES with 256-bit key
+
+        'twofish128-ctr', // OPTIONAL          Twofish in SDCTR mode, with 128-bit key
+        'twofish192-ctr', // OPTIONAL          Twofish with 192-bit key
+        'twofish256-ctr', // OPTIONAL          Twofish with 256-bit key
+
+        'aes128-cbc',     // RECOMMENDED       AES with a 128-bit key
+        'aes192-cbc',     // OPTIONAL          AES with a 192-bit key
+        'aes256-cbc',     // OPTIONAL          AES in CBC mode, with a 256-bit key
+
+        'twofish128-cbc', // OPTIONAL          Twofish with a 128-bit key
+        'twofish192-cbc', // OPTIONAL          Twofish with a 192-bit key
+        'twofish256-cbc',
+        'twofish-cbc',    // OPTIONAL          alias for "twofish256-cbc"
+        //                   (this is being retained for historical reasons)
+
+        'blowfish-ctr',   // OPTIONAL          Blowfish in SDCTR mode
+
+        'blowfish-cbc',   // OPTIONAL          Blowfish in CBC mode
+
+        '3des-ctr',       // RECOMMENDED       Three-key 3DES in SDCTR mode
+
+        '3des-cbc',       // REQUIRED          three-key 3DES in CBC mode
+        //'none'         // OPTIONAL          no encryption; NOT RECOMMENDED
+    );
+
+    /**
      * Default Constructor.
      *
      * $host can either be a string, representing the host, or a stream resource.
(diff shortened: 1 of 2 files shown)


bantu:
I am in general support of this. Maybe this should be extended to MACs and
other algorithms as well.

terrafrost:
I like it! I'm in agreement with bantu in that it'd be good to extend this to MACs and other algorithms as well.

Among other advantages of this is that it'd make a more accurate shim for libssh2, which lets you select the algorithms in [ssh2_connect](http://php.net/manual/en/function.ssh2-connect.php).

One thought I do have...  what should happen if you add an algorithm / mode that phpseclib doesn't actually support? By letting this be settable by the user that would be a possibility (whereas right now it isn't). I wonder what `ssh2_connect` does in that situation..

The only other thing I'm thinking is that you use type hinting for `setEncryptionAlgorithms`. I like type hinting and am doing it in the master branch but it's not being done in the 2.0 branch. It could be done in the 2.0 branch but I'm kinda thinking...  let's do it everywhere it could be done or let's not do it at all. So in that regard I think it'd just be easier to do type hinting only in the master branch.to give you a chance make these changes for the other algorithms (MAC,

kevindh89:
I've removed the type hinting in ``setEncryptionAlgorithms`` for the sake of consistency.

I'm not sure what happens if any unsupported algorithms are set. But probably unexpected issues, which you'll want to avoid, will occur.
For that reason a restriction was added for setting supported encryption algorithms only.

I'll open another pull request to make the other algorithms configurable.

terrafrost:
I just made a commit on a branch in terrafrost/phpseclib that should lessen the urgency of this:

https://github.com/phpseclib/phpseclib/issues/1171#issuecomment-325170715

I still do very much like this idea and would like to see it merged but other stuff needs to be done for me to want to merge it. I also think, in-so-far as the "invalid size" error is concerned, this is a better solution. ie. most people who encounter the "invalid size" error will probably just give up and use libssh2 or do SSH via the CLI or something. phpseclib will just be written off and no effort will be made to research or understand the problem. Letting users select which algorithms they'd want to use wouldn't eliminate the need for research whereas the commit I just made would.

Anyway, as I said in the link, I still need to merge it into 2.0 and master. I'll do that another day tho.

cabloo:
Glad to hear that. If I were to put up a PR with that, would it expedite things?

Here's the log:
https://pastebin.com/2ZLUzX52

It looks to me like the client & server send their supported methods, and then the server does not respond - just closes the connection. It's a Dell PowerConnect switch which I suspect has a buggy implementation of arcfour. I commented out arcfour128 & arcfour256 and it started working. I tried sending some dummy data of similar size and that worked so it's nothing wonky to do with packet sizes as far as I can tell.

Unfortunately, the switch is ACLed so I can't give out the IP for testing. I can test any code/ideas you'd like to try though.

terrafrost:
Sorry for not responding sooner.

Anyway, I have no idea why the server would be disconnecting. Normally, after both parties send each other an NET_SSH2_MSG_KEXINIT packet phpseclib sends a NET_SSH2_MSG_KEXDH_INIT.

If there were no supported algorithms that both server and client supported then you'd be getting an error like "No compatible server to client encryption algorithms found" (see #1389, for example), but you're not.

Indeed, here the algorithms that should be being used:

- diffie-hellman-group1-sha1
- ssh-rsa
- arcfour256
- arcfour256
- hmac-sha1-96
- hmac-sha1-96
- none

idk why the server is disconnecting. It may very well be the case that the server doesn't like the algorithms that the key exchange is negotiating but if that were the case then the sever shouldn't be advertising them as being support anyway.

Anyway, I'll try to backport the feature to let you select which algorithms are to be used this weekend. If you want to submit a PR, too, feel free!

Thanks!

cabloo:
> It may very well be the case that the server doesn't like the algorithms that the key exchange is negotiating but if that were the case then the sever shouldn't be advertising them as being support anyway.

Fairly certain this is indeed the case (even though it's buggy as hell) since simply commenting the two arcfour options out resolves the issue.

> Anyway, I'll try to backport the feature to let you select which algorithms are to be used this weekend. If you want to submit a PR, too, feel free!

That would be great! I probably wouldn't have time to work on this until next week, and it would definitely take me some time to familiarize myself with the code base/tests enough to accomplish that.

terrafrost:
I've backported the feature and it is now in the 1.0 and 2.0 branches:

https://github.com/phpseclib/phpseclib/commit/3ad44cb3d60c4c403e7a94c80ec04913ba99631c

I'm currently thinking I'll let it sit for a few days and will then do a release. Maybe late next week or something idk.

terrafrost:
More specifically, I've added `setPreferredAlgorithms()` method to SSH2.php. To configure it you pass an array to it as described at https://www.php.net/manual/en/function.ssh2-connect.php

terrafrost:
The default encryption algorithm is now dependent on what extensions you have installed. If you have mcrypt or OpenSSL installed it'll use aes*-ctr by default.

You can see the algorithms that phpseclib has built in support for by doing `print_r($ssh->getSupportedEncryptionAlgorithms())` or whatever.

This commit does not add support for any new algorithms. Only the 3.0 branch has new algorithm support (aes*-gcm, chacha20-poly1305, ecdsa, ed25519, curve25519, etc) and I will not be backporting the new algorithms.

terrafrost:
I just released 2.0.22, which includes this change.

Thanks!

PR #1157, opened 2017-08-03
Bugfixes

rrran:
Lots of PHPDOC fixes 
Some method visibility fixes 
Method signature fixes 
And more

### DIFF
--- a/phpseclib/File/X509.php
+++ b/phpseclib/File/X509.php
@@ -36,6 +36,7 @@ use phpseclib\File\ASN1\Element;
 use phpseclib\Math\BigInteger;
 use phpseclib\File\ASN1\Maps;
 
+
 /**
  * Pure-PHP X.509 Parser
  *
@@ -1564,7 +1565,7 @@ class X509
      * @param mixed $format optional
      * @param array $dn optional
      * @access public
-     * @return bool
+     * @return array|bool
      */
     public function getDN($format = self::DN_ARRAY, $dn = null)
     {
@@ -2796,7 +2797,7 @@ class X509
         */
         if (strtolower($date) == 'lifetime') {
             $temp = '99991231235959Z';
-            $temp = chr(ASN1::TYPE_GENERALIZED_TIME) . Functions::encodeLength(strlen($temp)) . $temp;
+            $temp = chr(ASN1::TYPE_GENERALIZED_TIME) . ASN1::encodeLength(strlen($temp)) . $temp;
             $this->endDate = new Element($temp);
         } else {
             $this->endDate = @date('D, d M Y H:i:s O', @strtotime($date));
@@ -3117,12 +3118,13 @@ class X509
      *
      * @param string $id
      * @param array $cert optional
+     * @param string $path
      * @access public
      * @return mixed
      */
-    public function getExtension($id, $cert = null)
+    public function getExtension($id, $cert = null, $path)
     {
-        return $this->getExtensionHelper($id, $cert);
+        return $this->getExtensionHelper($id, $cert, $path);
     }
 
     /**
@@ -3430,7 +3432,7 @@ class X509
      * Format a public key as appropriate
      *
      * @access private
-     * @return array
+     * @return array|bool
      */
     private function formatSubjectPublicKey()
     {
@@ -3602,7 +3604,7 @@ class X509
      *
      * @param array $crl optional
      * @access public
-     * @return array
+     * @return array|bool
      */
     public function listRevoked($crl = null)
     {
@@ -3676,7 +3678,7 @@ class X509
      * @param string $serial
      * @param array $crl optional
      * @access public
-     * @return array
+     * @return array|bool
      */
     public function getRevokedCertificateExtensions($serial, $crl = null)
     {
--- a/phpseclib/File/X509.php
+++ b/phpseclib/File/X509.php
@@ -3122,7 +3122,7 @@ class X509
      * @access public
      * @return mixed
      */
-    public function getExtension($id, $cert = null, $path)
+    public function getExtension($id, $cert = null, $path=null)
     {
         return $this->getExtensionHelper($id, $cert, $path);
     }
--- a/phpseclib/Math/BigInteger/Engines/BCMath/Base.php
+++ b/phpseclib/Math/BigInteger/Engines/BCMath/Base.php
@@ -60,7 +60,7 @@ abstract class Base extends BCMath
      * @param string $class
      * @return \phpseclib\Math\BigInteger\Engines\BCMath
      */
-    protected static function powModHelper(BCMath $x, BCMath $e, BCMath $n, $class)
+    public static function powModHelper(BCMath $x, BCMath $e, BCMath $n, $class)
     {
         if (empty($e->value)) {
             $temp = new $class();
--- a/phpseclib/Math/BigInteger/Engines/PHP/Base.php
+++ b/phpseclib/Math/BigInteger/Engines/PHP/Base.php
@@ -80,7 +80,7 @@ abstract class Base extends PHP
      * @param string $class
      * @return \phpseclib\Math\BigInteger\Engines\PHP
      */
-    protected static function powModHelper(PHP $x, PHP $e, PHP $n, $class)
+    public static function powModHelper(PHP $x, PHP $e, PHP $n, $class)
     {
         if (empty($e->value)) {
             $temp = new $class();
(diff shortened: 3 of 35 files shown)


phpseclib/Net/SFTP.php
```diff
@@ -3086,7 +3086,7 @@ public function getSupportedVersions()
      * @return bool
      * @access private
      */
-    private function disconnect_helper($reason)
+    protected function disconnect_helper($reason)
```
GrahamCampbell: Please explain this?

phpseclib/Net/SSH2.php
```diff
@@ -652,7 +652,7 @@ class SSH2
      * @see self::_get_channel_packet()
      * @access private
      */
-    private $curTimeout;
+    protected $curTimeout;
```
GrahamCampbell: Please minimize these changes to those that are required, since the phpdoc changes are bug fixes. :)

phpseclib/Net/SFTP.php
rrran: SFTP class extends SSH2 and overrides method disconnect_helper($reason), also it calls parent's method:

protected function disconnect_helper($reason)
    {
        $this->pwd = false;
        parent::disconnect_helper($reason);
    }

Parent's method was declared as private, so could not be called outside the class.
I changed it's visibility to protected. 
Also I have to change visibility of the successor's method.

phpseclib/File/X509.php
```diff
@@ -2796,7 +2797,7 @@ public function setEndDate($date)
         */
         if (strtolower($date) == 'lifetime') {
             $temp = '99991231235959Z';
-            $temp = chr(ASN1::TYPE_GENERALIZED_TIME) . Functions::encodeLength(strlen($temp)) . $temp;
+            $temp = chr(ASN1::TYPE_GENERALIZED_TIME) . ASN1::encodeLength(strlen($temp)) . $temp;
```
terrafrost: Nice catch! I'll have to add a unit test for this

phpseclib/File/X509.php
```diff
@@ -3117,12 +3118,13 @@ public function removeExtension($id)
...
      * @param array $cert optional
+     * @param string $path
      * @access public
      * @return mixed
      */
-    public function getExtension($id, $cert = null)
+    public function getExtension($id, $cert = null, $path)
     {
-        return $this->getExtensionHelper($id, $cert);
+        return $this->getExtensionHelper($id, $cert, $path);
```
terrafrost: Looks like a unit test will need to be written for this too. This is only called with a third parameter by `getRevokedCertificateExtension`.

`getExtensions` (plural vs singular) probably ought to be updated similarly, as well, and then `getRevokedCertificateExtensions` updated to call `getExtensions` instead of `getExtensionsHelper`.

phpseclib/Math/BigInteger/Engines/PHP/Base.php
```diff
@@ -80,7 +80,7 @@ public static function isValidEngine()
      * @param string $class
      * @return \phpseclib\Math\BigInteger\Engines\PHP
      */
-    protected static function powModHelper(PHP $x, PHP $e, PHP $n, $class)
+    public static function powModHelper(PHP $x, PHP $e, PHP $n, $class)
```
terrafrost: What's the motivation behind this change?

phpseclib/Math/BigInteger/Engines/BCMath/Base.php
```diff
@@ -60,7 +60,7 @@ public static function isValidEngine()
      * @param string $class
      * @return \phpseclib\Math\BigInteger\Engines\BCMath
      */
-    protected static function powModHelper(BCMath $x, BCMath $e, BCMath $n, $class)
+    public static function powModHelper(BCMath $x, BCMath $e, BCMath $n, $class)
```
terrafrost: You made a similar change in https://github.com/phpseclib/phpseclib/commit/f314f9795f2a5c43fc680cd6fee2beaf3e192239#r131527744 . I don't understand the motivation for that change nor do I understand it for this change.

phpseclib/File/X509.php
terrafrost: Also, spaces ought to be added before and after the `=` in the `$path=null` bit. PSR-1/2 do appear to discuss this but in areas where those standards do not provide clarity I'd say, as a general rule of thumb, the PEAR standards apply. Quoting https://pear.php.net/manual/en/standards.funcalls.php ,

> As displayed above, there should be one space on either side of an equals sign used to assign the return value of a function to a variable.

idk how well the phpseclib codebase actually conforms to these standards but, anyway, I can take care of this when I merge.

terrafrost:
The changes look good to me!

I guess I have a few things I need to do:

1. Delete `ASN1::decodeLength` per https://github.com/phpseclib/phpseclib/pull/1157#pullrequestreview-54519828
2. Write a unit test for `X509::setEndDate` per https://github.com/phpseclib/phpseclib/pull/1157#discussion_r131526883
3. Write a unit test for `X509::getRevokedCertificateExtension` and make some changes in `X509::getExtensions` and `getRevokedCertificateExtensions` per https://github.com/phpseclib/phpseclib/pull/1157#discussion_r131527226
4. There's a call to `$this->disconnectHepler` in SSH2.php - that should be changed to `$this->disconnect_helper`.

One question I have that it'd be cool if you could answer...

1. Why did you change the visibility of `powModHelper` from `protected` to `public` in both the PHP and BCMath engines per  https://github.com/phpseclib/phpseclib/pull/1157/commits/f314f9795f2a5c43fc680cd6fee2beaf3e192239#r131527744 and https://github.com/phpseclib/phpseclib/pull/1157/commits/6f36c49baafc8af2821c18088ebec99e4a53a62d#r131527812?

Thanks!

phpseclib/File/X509.php
```diff
@@ -3117,12 +3118,13 @@ public function removeExtension($id)
      *
      * @param string $id
      * @param array $cert optional
+     * @param string $path
      * @access public
      * @return mixed
      */
-    public function getExtension($id, $cert = null)
+    public function getExtension($id, $cert = null, $path)
```
terrafrost: I think that should be `$path = null`. As is not all the calls to `getExtension` have a third parameter:

```php
$issuerAltName = $this->getExtension('id-ce-subjectAltName', $issuer->currentCert);
```

Also, that's how it's defined with `getExtensionHelper`:

```php
    private function getExtensionHelper($id, $cert = null, $path = null)
```

terrafrost: Never mind this - it looks like you fixed this specific issue with https://github.com/phpseclib/phpseclib/pull/1157/commits/eb21fb2bcb781a25c718cdf19f24bc59ea83ae2d

terrafrost:
The changes  I mention that I'd do have been done here:

https://github.com/terrafrost/phpseclib/tree/pre-1157-changes

This branch has failing unit tests that you're changes should fix. Just LMK about the `powModHelper` thing and then I'll merge.

Thanks!

phpseclib/Math/BigInteger/Engines/BCMath/Base.php
rrran: File PHP.php
method:

 protected function powModInner(PHP $e, PHP $n)
    {
        try {
            $class = static::$modexpEngine;
            return $class::powModHelper($this, $e, $n, static::class);
        } catch (\Exception $err) {
            return PHP\DefaultEngine::powModHelper($this, $e, $n, static::class);
        }
    }

As you can see here's a static call which can't be performed if the powModHelper method is not public

phpseclib/Math/BigInteger/Engines/PHP/Base.php
rrran: File PHP.php
method:

protected function powModInner(PHP $e, PHP $n)
{
try {
$class = static::$modexpEngine;
return $class::powModHelper($this, $e, $n, static::class);
} catch (\Exception $err) {
return PHP\DefaultEngine::powModHelper($this, $e, $n, static::class);
}
}

As you can see here's a static call which can't be performed if the powModHelper method is not public

terrafrost: Engines/PHP/Base extends Engines/PHP so visibility shouldn't be an issue. Indeed, if visibility were an issue then one would think that changing the visibility from protected to private wouldn't break any unit tests but it does:

https://travis-ci.org/terrafrost/phpseclib/jobs/261813738

Issue #1046, opened 2016-10-11
Error: Uninitialized string offset 0 \vendor\phpseclib\phpseclib\phpseclib\Net\SSH2.php on line 1110


smtlab:
Getting the following error .Please help!

`PHP Notice:  Uninitialized string offset: 0 in \vendor\phpseclib\phpseclib\phpseclib\Net\SSH2.php on line 1110`

`Notice: Uninitialized string offset: 0 in E:\xampp\htdocs\symfony\vendor\phpseclib\phpseclib\phpseclib\Net\SSH2.php on line 1110
PHP Notice:  Expected SSH_MSG_KEXINIT in \vendor\phpseclib\phpseclib\phpseclib\Net\SSH2.php on line 1111`

`Notice: Expected SSH_MSG_KEXINIT in \vendor\phpseclib\phpseclib\phpseclib\Net\SSH2.php on line 1111`

After diggin into the code found that `_get_binary_packet()` function called on line 1104 in SSH2.php returns empty string which is causing this error.
Please help what is the cause of this error.

terrafrost:
Quoting [RFC4253](https://tools.ietf.org/html/rfc4253#page-5), "Key exchange will begin immediately after sending [the SSH-2.0-xxx identification string]". Later, in the same RFC ([Section 7.1](https://tools.ietf.org/html/rfc4253#section-7.1)) it says "Key exchange begins by each side sending the following packet".

If your SSH server is not sending the key exchange packet immediately than I'd say your server is buggy. I would actually be curious to know what your server identification string is. You can get it by doing `echo $ssh->getServerIdentification();`

That said, it is a fact that some SSH servers are buggy and sometimes workarounds can be implemented to work on those servers. PuTTY has a bunch of workarounds present:

https://the.earth.li/~sgtatham/putty/0.60/htmldoc/Chapter4.html#config-ssh-bugs

phpseclib doesn't need all of these and, in some cases, needs workarounds that PuTTY doesn't, because phpseclib supports ciphers that PuTTY doesn't.

Long story short, try this:

https://github.com/terrafrost/phpseclib/tree/2.0-send-kex-before-rcpt

smt92:
Thank you for replying.

I tested with multiple servers getting same error. Also I can connect with PuTTY with no issues.
Server identification Id: SSH-2.0-OpenSSH_5.3

terrafrost:
Did you try the branch I made?:

[...]

PR #1162, opened 2017-08-07
SSH2: send KEXINIT packet and identification string first or last (for 1.0 branch)

terrafrost:
This branch adds four new methods:

- sendIdentificationStringFirst()
- sendIdentificationStringLast()
- sendKEXINITFirst()
- sendKEXINITLast()

[RFC4253 § 4.2](https://tools.ietf.org/html/rfc4253#section-4.2) says "_when the connection has been established, both sides MUST send an identification string_" but it does not say which side should send it first.

[RFC4253 § 7.1](https://tools.ietf.org/html/rfc4253#section-7.1) says "_key exchange begins by each side sending the_ [SSH_MSG_KEXINIT] _packet_" but again it does not say which side should send it first.

In theory, it shouldn't matter which side sends it first but in practice it sometimes does.

#1046 demonstrates that phpseclib sending the SSH_MSG_KEXINIT packet last can sometimes result in errors so, in lieu of any evidence that sending it first can result in errors, I've changed it so that it sends it first by default now. BUT if that does cause errors then phpseclib can be easily made to send it last.

https://stackoverflow.com/q/40430024/569976 demonstrates that phpseclib sending the identification packet last can result in errors but https://github.com/phpseclib/phpseclib/commit/535104cb842271eb9c76c3739c910477bf16ae75#commitcomment-23493058 demonstrates that sending it first can result in errors as well. This commit preserves the current behavior but makes it so that the current behavior can be overwritten.

### DIFF
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -876,6 +876,22 @@ class Net_SSH2
     var $agent;
 
     /**
+     * Send the identification string first?
+     *
+     * @var bool
+     * @access private
+     */
+    var $send_id_string_first = true;
+
+    /**
+     * Send the key exchange initiation packet first?
+     *
+     * @var bool
+     * @access private
+     */
+    var $send_kex_first = true;
+
+    /**
      * Default Constructor.
      *
      * $host can either be a string, representing the host, or a stream resource.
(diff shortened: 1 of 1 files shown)


dvaeversted:
This solved an issue with connection errors against Cisco devices that we started experiencing after upgrading from a 0.X release of phpseclib, to 2.0.6

terrafrost:
This has been merged.

terrafrost:
It's interesting SSH_MSG_NEWKEYS hasn't caused anyone an issue.

Quoting [RFC4253 § 7.3](https://tools.ietf.org/html/rfc4253#section-7.3), "_Key exchange ends by each side sending an SSH_MSG_NEWKEYS message_".

phpseclib sends SSH_MSG_NEWKEYS first.

PR #1165, opened 2017-08-08
Allowed algorithms to be configured in SSH2

kevindh89:
All algorithms in SSH2 are made configurable as per request in this PR:
https://github.com/phpseclib/phpseclib/pull/1156

### DIFF
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -150,6 +150,61 @@ class SSH2
     const READ_NEXT = 3;
     /**#@-*/
 
+    /**#@+
+     * @access public
+     * @see \phpseclib\Net\SSH2::setKexAlgorithms()
+     */
+    static $SUPPORTED_KEX_ALGORITHMS = array(
+        // Elliptic Curve Diffie-Hellman Key Agreement (ECDH) using
+        // Curve25519. See doc/curve25519-sha256@libssh.org.txt in the
+        // libssh repository for more information.
+        'curve25519-sha256@libssh.org',
+
+        // Diffie-Hellman Key Agreement (DH) using integer modulo prime
+        // groups.
+        'diffie-hellman-group1-sha1', // REQUIRED
+        'diffie-hellman-group14-sha1', // REQUIRED
+        'diffie-hellman-group-exchange-sha1', // RFC 4419
+        'diffie-hellman-group-exchange-sha256', // RFC 4419
+    );
+    /**#@-*/
+
+    /**#@+
+     * @access public
+     * @see \phpseclib\Net\SSH2::setServerHostKeyAlgorithms()
+     */
+    static $SUPPORTED_SERVER_HOST_KEY_ALGORITHMS = array(
+        'ssh-rsa', // RECOMMENDED  sign   Raw RSA Key
+        'ssh-dss'  // REQUIRED     sign   Raw DSS Key
+    );
+    /**#@-*/
+
+    /**#@+
+     * @access public
+     * @see \phpseclib\Net\SSH2::setMACAlgorithms()
+     */
+    static $SUPPORTED_MAC_ALGORITHMS = array(
+        // from <http://www.ietf.org/rfc/rfc6668.txt>:
+        'hmac-sha2-256',// RECOMMENDED     HMAC-SHA256 (digest length = key length = 32)
+
+        'hmac-sha1-96', // RECOMMENDED     first 96 bits of HMAC-SHA1 (digest length = 12, key length = 20)
+        'hmac-sha1',    // REQUIRED        HMAC-SHA1 (digest length = key length = 20)
+        'hmac-md5-96',  // OPTIONAL        first 96 bits of HMAC-MD5 (digest length = 12, key length = 16)
+        'hmac-md5',     // OPTIONAL        HMAC-MD5 (digest length = key length = 16)
+        'none'          // OPTIONAL        no MAC; NOT RECOMMENDED
+    );
+    /**#@-*/
+
+    /**#@+
+     * @access public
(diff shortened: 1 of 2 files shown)


terrafrost:
I got a merge conflict when I tried to merge this and #1156 (since this one doesn't do `$encryption_algorithms`). If you could go ahead and merge them and resolve the conflict that'd be cool :)

I noticed you defined `getMACAlgorithms`, `getCompressedAlgorithms` and `getEncryptionAlgorithms` but not `getServerHostKeyAlgorithm` or `getKexAlgorithms`. I guess the reason that you didn't define them is because they're already defined but the way `getServerHostKeyAlgorithm` and `getKexAlgorithms` currently work there could be some confusion I think. The latter two methods return the servers preferences whereas the former three methods return the clients preferences. What if we add a second parameter to the first three? One that takes either `SSH2::CLIENT_TO_SERVER` (1) or `SSH2::SERVER_TO_CLIENT` (2) as parameters? In 2.0, to preserve BC, we'd still want `getEncryptionAlgorithmsClient2Server`, for BC, even tho this proposed change would make that method redundant, but we could remove the `*Client2Server` and `*Server2Client` methods in the master branch.

I guess to achieve a perfect shim for [ssh2_connect](http://php.net/ssh2-connect) we'd need to be able to set the server_to_client algorithms independently from the client_to_server ones. That said, I do think it makes more sense if they're not defined independently - it's the more common use-case I imagine. PuTTY, for example, doesn't let you select these algorithms independently. BUT... I think we can kinda satisfy both "requirements" if we add a second optional parameter to `setEncryptionAlgorithms`, `setMACAlgorithms` and `setCompressedAlgorithms`. This second parameter would be either `SSH2::CLIENT_TO_SERVER` (1) or `SSH2::SERVER_TO_CLIENT` (2). If the parameter is not explicitly set it's implicitly `SSH2::CLIENT_TO_SERVER | SSH2::SERVER_TO_CLIENT` (3).

Otherwise it's looking like an awesome commit - I'm looking forward to getting these features into phpseclib and I appreciate your patience with me :) If you don't want to do the changes I've proposed I can probably do them myself but it'd be cool if you could do them :)

phpseclib/Net/SSH2.php
```diff
@@ -150,6 +150,61 @@ class SSH2
     const READ_NEXT = 3;
     /**#@-*/
 
+    /**#@+
```
terrafrost: The `#@+`'s are unnecessary. Their main purpose is to tell phpdoc or whatever to propagate the values down until it encounters a `#@=`.

ricbra:
Our main concern is to get the initial change merged into 2.0 (https://github.com/phpseclib/phpseclib/pull/1156/commits/a828656a043524af1b4b35c2494d8143ff8adad4) as we currently are using a private fork of this repository to fix it in production which we want to get rid of. Both @kevindh89 and I are working on these open source contributions during office hours meaning our current client / employer is paying for this work. Thats all cool and stuff but unfortunately it means we're being limited with the time being invested. 

We understand the concerns in your comment but its difficult for us to find the time to discuss and solve all the issues mentioned here. I hope you understand this. It's not that we don't want to help you with this.

That being said, what would be the most easy to way to get our changes merged? Solve the merge conflict you mentioned?

Would be cool to get our work in, as it will help quite some people in the future stumbling on the same problem as us.

terrafrost:
I intend to get these merged in. If the original author won't make the changes I suggested I will as time permits. If you would like faster progress be made on this, however, do feel free to pay.

Some food for thought:

1. I am not paid to work on this during office hours so all my contributions are done in my free time
2. I am on vacation in Nebraska right now to see a total solar eclipse.
3. If you want to pay, I can prioritize this higher than some of the other already outstanding / unresolved phpseclib issues and my own personal commitments, provide an ETA, give weekly or daily status updates, contingent upon how much you pay, etc.
4. Getting the changes merged in and having a release made are two separate things. No release has been made since this PR was originally proposed. Whether or not a commit, alone, would be sufficient for your purposes would depend on your composer.json file (I assume that's how you installed phpseclib).

metaclassing:
Off topic question for terrafrost - what part of NE are you visiting? If you come to Omaha I will buy you steak!

terrafrost:
Oh nice! I'm in Grand Island, NE. Flight leaves at 6am tomorrow. I've never caught a flight that early before :\

Omaha isn't in the path of totality but didja drive to somewhere that was? It was pretty cool!

metaclassing:
I caught the eclipse, 30 minutes south of town got a nearly perfect view, hope you had a good stay in NE. Drop me an email if you ever travel through Omaha, I still owe you thanks for all the help over the past few years.

ricbra:
@terrafrost I totally understand what you're saying (I also have some OSS stuff published which I have to maintain in my free time) and it was absolutely **not** the point that **you** should be the person who's gonna solve our issues unpaid in your free time. So, apologies for the misunderstanding and we're very happy with all the work you've done so far with this library. Hence, we're using it for quite some time now.

We were hoping we could tighten the scope of these changes somehow: what started as a small PR with just one new method it now has become a bit bigger and to be honest we struggle with the latest comments. We're not that familiair with this library and the inner workings of SSH(2) and/or PuTTy so we're getting lost there quite fast.

So the combination of us getting lost and the limited time was the reason I mentioned this. And again, we're not expecting you to code for free for us! We have no expectations of this getting merged in very quickly, we just want to help as much as possible so our work wasn't done for nothing.

terrafrost:
I've added a new `setPreferredAlgorithms()` method to SSH2.php. To configure it you pass an array to it as described at https://www.php.net/manual/en/function.ssh2-connect.php

I'll probably do a new release late next week or something idk,

Thanks!

terrafrost:
I just released 2.0.22, which includes this change.

Thanks!

PR #1164, opened 2017-08-07
Unit tests fixes

rrran:
Now a bit less tests 'ignored'

PR #1180, opened 2017-09-12
Another set of bugfixes

rrran:


.gitignore
```diff
@@ -1,2 +1,4 @@
 /vendor
 /composer.phar
+/.idea/
+.gitignore
```
terrafrost: I'm fine with .idea being removed (I recognize that directory as being one PHP Storm creates) but I'll probably do another commit after I merge this one removing .gitignore. For ignoring .gitignore you could do this:

https://stackoverflow.com/questions/767147/ignore-the-gitignore-file-itself/767213#767213

tests/Unit/Crypt/BlowfishTest.php
```diff
@@ -77,7 +77,15 @@ public function testVectors($engine, $key, $plaintext, $expected)
...
-        if (!$bf->isValidEngine($engine)) {
+
+        $engines = array(
+            'PHP'=>\phpseclib\Crypt\Common\SymmetricKey::ENGINE_INTERNAL,
+            'Eval'=>\phpseclib\Crypt\Common\SymmetricKey::ENGINE_EVAL,
+            'mcrypt'=>\phpseclib\Crypt\Common\SymmetricKey::ENGINE_MCRYPT,
+            'OpenSSL'=>\phpseclib\Crypt\Common\SymmetricKey::ENGINE_OPENSSL,
+        );
+
+        if (!$bf->isValidEngine($engines[$engine])) {
```
terrafrost: I think this is a bad change. Right now there are 206 skipped unit tests:

https://travis-ci.org/phpseclib/phpseclib/jobs/281628707

After this commit there will be 308 tests:

https://travis-ci.org/phpseclib/phpseclib/jobs/274496080

In https://github.com/phpseclib/phpseclib/commit/9d8f014748e4787c4f54a1bb8070cc4c8bb44497#comments I updated phpseclib to take strings instead of ints. This change was inspired by another PR you made - #1164. With this change you're changing what's passed to `isValidEngine` back from strings to ints.

terrafrost: Never mind - I see that you undid this lol. Sorry!

phpseclib/File/X509.php
```diff
@@ -2820,7 +2821,8 @@ public function setEndDate($date)
                 $date = new DateTime($date);
             }
 
-            $this->endDate = $date->format('D, d M Y H:i:s O', new DateTimeZone(@date_default_timezone_get()));
+            $date->setTimezone(new DateTimeZone(@date_default_timezone_get()));
+            $this->endDate = $date->format('D, d M Y H:i:s O');
```
terrafrost: I think this is best added as a second parameter to the `DateTime` constructor. The `DateTime` constructor does accept two parameters, the second of which is `DateTimeZone`:

http://php.net/manual/en/datetime.construct.php

Also, without it being in the constructor, `$date = new DateTime('now');` returns a fatal error when the timezone hasn't been set in php.ini due to an exception not being caught.

It doesn't really serve the same point if set in `$date->setTimezone`.

Anyway, nice job on finding the issue! Altho I think I'll take a different route for fixing the issue, you found it - thanks!

terrafrost:
This has been merged - thanks!

PR #1181, opened 2017-09-13
Replace extension_loaded with forward compatible function exists check

Smitsel:
Since libsodium will be replaced by the new sodium extension which has been accepted into PHP 7.2 the extension loaded will not work for us when we start using the new sodium package.
Reference: https://github.com/jedisct1/libsodium-php

So its replaced by function exists call to allow us to use some kind of polyfill to guarantee compatibility. We created such a polyfill: https://github.com/mollie/polyfill-libsodium

### DIFF
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -1267,7 +1267,7 @@ class SSH2
         $identifier = 'SSH-2.0-phpseclib_2.0';
 
         $ext = [];
-        if (extension_loaded('libsodium')) {
+        if (function_exists('\\Sodium\\library_version_major')) {
             $ext[] = 'libsodium';
         }
 
--- a/tests/Unit/Net/SSH2Test.php
+++ b/tests/Unit/Net/SSH2Test.php
@@ -41,7 +41,7 @@ class Unit_Net_SSH2Test extends PhpseclibTestCase
         $identifier = self::callFunc($this->createSSHMock(), 'generate_identifier');
         $this->assertStringStartsWith('SSH-2.0-phpseclib_2.0', $identifier);
 
-        if (extension_loaded('libsodium')) {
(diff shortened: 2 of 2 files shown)


bantu:
@paragonie-scott What's your take on this?

terrafrost:
For my part...  I'm currently traveling in Paris, France, and will be in Italy for 10-11 days after Paris so I may not be as prompt as one would like during my travels.

paragonie-scott:
This will be more sustainable. We may want to add alternative logic for `if (extension_loaded('sodium'))` to use the new `sodium_*` API instead of `\Sodium\*`, but that's a "sometime before 2018" concern.

Smitsel:
Would be awesome if this could be merged in the near future since it would allow us to use ext-sodium instead of the old libsodium package.

bantu:
Should this go into 2.0 as well? Does it break anything? I don't know to be
very honest.

willemstuursma:
We would really appreciate this being merged.

* It allows phpseclib to work with `paragonie/sodium_compat`
* It allows phpseclib to work with `ext-sodium` using `mollie/polyfill-libsodium`.

terrafrost:
This change makes the libsodium checks more consistent.

One part of phpseclib already does this and has since Feb 11, 2016:

https://github.com/phpseclib/phpseclib/commit/fc4599985ab7617b6565b0de53cf93299192e941#diff-82a5983a6daace4acbcc65edf2dfb605

That check and this check have been done differently since the initial commit that added ECDH support:

https://github.com/phpseclib/phpseclib/commit/58f8affcad26c208cadf1b40221e90b1c54eecf0

I suppose that's neither here nor there tho. I cherry picked this commit into the 2.0 branch and merged the 2.0 branch into the master branch.

Thanks!

PR #1182, opened 2017-09-18
Fix wrong doccomment annotation of return type

eternalharvest:
Fix wrong DocComment annotation of return type declaration.

I'm new to this phseclib project. so I'm not sure if it's correct, but it seems there are many wrong DocComment annotations inconsistents with access modifier of php syntax (public / protected / private).
Must I wrote some patches to fix these wrong DocComment, or just my lack of understanding about this code base?

### DIFF
--- a/phpseclib/Crypt/Common/SymmetricKey.php
+++ b/phpseclib/Crypt/Common/SymmetricKey.php
@@ -1584,7 +1584,7 @@ abstract class SymmetricKey
      *
      * May need to be overwritten by classes extending this one in some cases
      *
-     * @return int
+     * @return string
      * @access private
      */
     protected function openssl_translate_mode()


Sxderp:
> DocComment annotations inconsistents with access modifier of php syntax (public / protected / private). Must I wrote some patches to fix these wrong DocComment, or just my lack of understanding about this code base?

I would recommend not modifying the access inconsistencies. There was a post by @terrafrost in regard to them, but I can't find it. The gist of it is that the methods are often more accessible than necessary in order to facilitate easier debugging / support. If methods that should _technically_ be private are public then they can be called directly and dumped.

eternalharvest:
@Sxderp thank u, I understand about that.

How about return types I commited in this pull-request?
I think this DocComment is wrong...

terrafrost:
@Sxderp - nice memory! That's def the position I've taken for the 1.0 and 2.0 branches but for the master branch I've caved a bit and have succumbed to more commonly held notions of best practices lol.

Anyway, the commit looks good and has been merged.

Thanks!

PR #1190, opened 2017-09-30
Fix BigInteger BadConfigurationException

pcf0:
If you dont have gmp extension installed, you are not able to use BigInteger class because of the BadConfigurationException.

### DIFF
--- a/phpseclib/Math/BigInteger.php
+++ b/phpseclib/Math/BigInteger.php
@@ -138,6 +138,7 @@ class BigInteger implements \Serializable
                     self::setEngine($engine[0], isset($engine[1]) ? $engine[1] : []);
                     break;
                 } catch (\InvalidArgumentException $e) {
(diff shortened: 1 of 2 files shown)


terrafrost:
Good "catch" lol.

It looks like `setEngine` already deals with this situation - just not the constructor. Here's what `setEngine` does:

```php
        foreach ($modexps as $modexp) {
            try {
                $fqmain::setModExpEngine($modexp);
                $found = true;
                break;
            } catch (\Exception $e) {
            }
        }
```

You're doing a different approach for the constructor. I'm ok with either approach but I think the approaches ought to be consistent.

I went ahead and made another commit to make the catch behavior consistent by utilizing the existing `setEngine` approach but if you think the approach you used make a PR and I'll merge it! Like I said, I'm ok with either! My choosing the existing approach was arbitrary lol.

PR #1192, opened 2017-10-04
Add 'cfb8' cipher mode of operation support

eternalharvest:
## What's this ?
This patch add 'cfb8' cipher mode of operation to SymmetricKey class.
This mode was intended to be used with AES128 (Rijndael) encryption.

This patch supports the engines as foollows
* OpenSSL
* mcrypt
* Eval
* PHP

## Why this mode needed?
**Becase I NEED!**

 mcrypt extension is planned to be removed from next release of php version 7.2. And I have some codebase which requires AES-128-CFB8 with mcrypt extension. So, I need to replace the code with mcrypt compatible 'cfb8' mode encryption / decryption library. And phpseclib has internal PHP / Eval engine.

## How to use
```php
<?php
use phpseclib\Crypt\AES;

$key = "VpwGPQ33RrgiEmMV";
$msg = "this is secret message";

$aes = new AES("CFB8");
$aes->enableContinuousBuffer();
$aes->setKey($key);
$aes->setIV($key);

$cipher = $aes->encrypt($msg);
$actual = $aes->decrypt($cipher);

assert($actual === $msg);
echo "    original message: $msg".PHP_EOL;
echo "decrypted message: $actual".PHP_EOL;
```

### Follow up
I carefully tested this patch and add this mode to the test case of AES.
But I'm not sure if this test is enough or not.

And It may contains some problem I didn't noticed.
Please tell me if u found any problem about this patch.

### DIFF
--- a/phpseclib/Crypt/Common/SymmetricKey.php
+++ b/phpseclib/Crypt/Common/SymmetricKey.php
@@ -1128,14 +1128,14 @@ abstract class SymmetricKey
 
                 for ($i = 0; $i < $len; ++$i) {
                     $ciphertext .= ($c = $plaintext[$i] ^ $this->encryptBlock($iv));
-                    $iv = substr($iv, 1, $this->block_size - 1) . $c;
+                    $iv = substr($iv, 1) . $c;
                 }
 
                 if ($this->continuousBuffer) {
-                    if ($len >= $this->block_size) {
-                        $this->encryptIV = substr($ciphertext, -$this->block_size);
+                    if ($len >= $block_size) {
+                        $this->encryptIV = substr($ciphertext, -$block_size);
                     } else {
-                        $this->encryptIV = substr($this->encryptIV, $len - $this->block_size) . substr($ciphertext, -$len);
+                        $this->encryptIV = substr($this->encryptIV, $len - $block_size) . substr($ciphertext, -$len);
                     }
                 }
                 break;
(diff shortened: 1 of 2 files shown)


terrafrost:
I'll try to review this and your other commit this weekend.

Thanks!

phpseclib/Crypt/Common/SymmetricKey.php
```diff
@@ -926,6 +932,16 @@ public function encrypt($plaintext)
...
 
                     return $ciphertext;
+                case self::MODE_CFB8:
+                    $ciphertext = openssl_encrypt($plaintext, $this->cipher_name_openssl, $this->key, OPENSSL_RAW_DATA | OPENSSL_ZERO_PADDING, $this->encryptIV);
+                    if ($this->continuousBuffer) {
+                            if (($len = strlen($ciphertext)) >= $this->block_size) {
+                                    $this->encryptIV = substr($ciphertext, -$this->block_size);
+                            } else {
+                                    $this->encryptIV = substr($this->encryptIV, $len - $this->block_size) . substr($ciphertext, -$len);
+                            }
```
terrafrost: Quoting [PSR-2: Coding Style Guide](http://www.php-fig.org/psr/psr-2/), "_code MUST use 4 spaces for indenting, not tabs_". This piece of code is using 8 spaces for indenting - not 4.

terrafrost:
So there's one test I want to run on this but I need to be getting to work lol. I'll try to get this done this evening - thanks!

phpseclib/Crypt/Common/SymmetricKey.php
eternalharvest: What the hell!! I didn't notice that...
I will fix it!

Should I exec `git rebase -i` to squash the commit and force update the PR?

eternalharvest:
Thank you for taking a time to reviwing my PR!

eternalharvest:
fix indent and bit refactor the code!

terrafrost:
Merged!

I kinda wonder if the motivation behind this PR was to get it in https://github.com/phpseclib/mcrypt_compat , however, that uses phpseclib 2.0 - not phpseclib master. I could backport the changes you did from the master branch to the 2.0 branch but it'd be cool if you did that if that's what you were wanting :)

As far as using the master branch with https://github.com/phpseclib/mcrypt_compat ...  the master branch has an unstable API atm. When the API becomes finalized the 3.0 branch will be created off of it. At that point, creating a 2.0 branch of mcrypt_compat would probably be appropriate. The 2.0 version of mcrypt_compat would be more complicated as the master branch is less forgiving than the 2.0 branch. In the 2.0 branch if you provide a key that's too short phpseclib null pads the key to the correct length, which is what mcrypt does. But in the past year or two there's been a movement in the PHP community (led primarily by a guy named sarciszewski) to make cryptographic primitives such as these less user friendly to discourage sub optimal crypto. eg. if you want to take full advantage of the security AES or whatever has to offer you should be using a key of the proper length and if you're using a password for a key you should be using `setPassword` (algorithm choice not withstanding). So to make mcrypt_compat work with what'll eventually become the 3.0 branch we'd need to implement null padding and stuff like that in it.

eternalharvest:
Thank you for the detailed information!
The motivation behind this PR is not intended to use with mcrypt_compat though this information is very helpfull.

I was not considering about which branch to merge this PR.
But, I am very pleased if this PR is included in the next release of 2.x!

Anyway, Thanks a lot!!

PR #1193, opened 2017-10-05
Fix wrong doccomment annotation and Suggestion

eternalharvest:
Fix wrong doccomment annotations.

## Suggestion
It seems type of first argument of constructor and `setPreferredEngine()` method was changed by the commit 4171262b9

But I would rather wish I can use constants integer value like before.
May I wrote patch which accepts constants value in addition like below?

```php
<?php /* file: phpseclib/Crypt/Common/SymmetricKey.php */

/**
 * Default constructor
 * @param int|string $mode name of mode as string or constant integer value
 * @access public
 * @throws \InvalidArgumentException if an invalid / unsupported mode is provided
 */
 public function __construct($mode)
 {
  if (is_string($mode)) {
    $mode = strtolower($mode);
    // necessary because of 5.6 compatability; we can't do isset(self::MODE_MAP[$mode]) in 5.6
    $map = self::MODE_MAP;
    if (!isset($map[$mode])) {
      throw new \InvalidArgumentException('No valid mode has been specified');
    }
    $mode = self::MODE_MAP[$mode];
  }

  switch ($mode) {
    case self::MODE_ECB:
    case self::MODE_CBC:
      $this->paddable = true;
      break;
    case self::MODE_CTR:
    case self::MODE_CFB:
    case self::MODE_OFB:
    case self::MODE_STREAM:
      $this->paddable = false;
      break;
    default:
      throw new \InvalidArgumentException('No valid mode has been specified');
  }
  
  $this->mode = $mode;
 }
```

```php
<?php

// same all
new AES(AES::MODE_CFB);
new AES('CFB');
new AES('cfb');
```

### DIFF
--- a/phpseclib/Crypt/Common/SymmetricKey.php
+++ b/phpseclib/Crypt/Common/SymmetricKey.php
@@ -481,7 +481,7 @@ abstract class SymmetricKey
      *
      * - ofb
      *
-     * @param int $mode
+     * @param string $mode
      * @access public
      * @throws \InvalidArgumentException if an invalid / unsupported mode is provided
      */
@@ -1795,7 +1795,7 @@ abstract class SymmetricKey
      * If the preferred crypt engine is not available the fastest available one will be used
      *
      * @see self::__construct()
-     * @param int $engine
+     * @param string $engine
      * @access public
      */
     public function setPreferredEngine($engine)


terrafrost:
> But I would rather wish I can use constants integer value like before.

How come? I mean, I guess I could add it as some undocumented feature but idk I like the strings better lol.

Like what if you're doing a switch statement? eg.

```
case '3des-cbc':
    $cipher = new TripleDES(TripleDES::MODE_CBC);
    break;
case 'aes128-cbc':
    $cipher = new AES(AES::MODE_CBC);
    break;
```

idk I just think it looks cleaner as strings. And I guess you could do `SymmetricKey::MODE_CBC` instead of `TripleDES::MODE_CBC` but idk - it just seems long and verbose lol.

I guess I'm not opposed to it but if you do it add a unit test because that's not the main way I'll be using it internally and as such the possibility that things might get broken with a subsequent commit would be higher.

PR #1197, opened 2017-10-23
Sometimes at php7 uniqid() return same result as their previous call.…

lzylinl:
… Here generates few equal labels wich makes error on eval. Does not know why uniqid() return same values but this is fact and my construction is more correct than previous. No need to call label id generation and use dirty goto :')


Sxderp:
Go-to isn't dirty, just underappreciated.

terrafrost:
I'll try to take a look in the next few days.

Thanks!

PR #1204, opened 2017-11-02
Unveil received value in unexpected value exceptions (SFTP)

bittner:
From SFTP servers that are configured in a way this library can't handle we should show the unexpected value we receive when it doesn't correspond to one of the constants we expect.

This PR adds the missing information to all the related exceptions in SFTP handling.

_Made with :heart: from @comsolit, Switzerland._ :smiley:

### DIFF
--- a/phpseclib/Net/SFTP.php
+++ b/phpseclib/Net/SFTP.php
@@ -487,7 +487,8 @@ class SFTP extends SSH2
 
         $response = $this->get_sftp_packet();
         if ($this->packet_type != NET_SFTP_VERSION) {
-            throw new \UnexpectedValueException('Expected SSH_FXP_VERSION');
+            throw new \UnexpectedValueException('Expected NET_SFTP_VERSION. '
+                                              . 'Got packet type: ' . $this->packet_type);
         }
 
         if (strlen($response) < 4) {
@@ -693,7 +694,8 @@ class SFTP extends SSH2
                     $this->logError($response);
                     return false;
                 default:
-                    throw new \UnexpectedValueException('Expected SSH_FXP_NAME or SSH_FXP_STATUS');
+                    throw new \UnexpectedValueException('Expected NET_SFTP_NAME or NET_SFTP_STATUS. '
+                                                      . 'Got packet type: ' . $this->packet_type);
             }
         }
 
@@ -769,7 +771,8 @@ class SFTP extends SSH2
                 $this->logError($response);
                 return false;
             default:
-                throw new \UnexpectedValueException('Expected SSH_FXP_HANDLE or SSH_FXP_STATUS');
+                throw new \UnexpectedValueException('Expected NET_SFTP_HANDLE or NET_SFTP_STATUS' .
+                                                    'Got packet type: ' . $this->packet_type);
         }
 
         if (!$this->close_handle($handle)) {
@@ -903,7 +906,8 @@ class SFTP extends SSH2
                 $this->logError($response);
                 return false;
             default:
-                throw new \UnexpectedValueException('Expected SSH_FXP_HANDLE or SSH_FXP_STATUS');
+                throw new \UnexpectedValueException('Expected NET_SFTP_HANDLE or NET_SFTP_STATUS. '
+                                                  . 'Got packet type: ' . $this->packet_type);
         }
 
         $this->update_stat_cache($dir, []);
@@ -969,7 +973,8 @@ class SFTP extends SSH2
                     }
                     break 2;
                 default:
-                    throw new \UnexpectedValueException('Expected SSH_FXP_NAME or SSH_FXP_STATUS');
+                    throw new \UnexpectedValueException('Expected NET_SFTP_NAME or NET_SFTP_STATUS. '
+                                                      . 'Got packet type: ' . $this->packet_type);
             }
         }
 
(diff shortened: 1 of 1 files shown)


terrafrost:
`$this->packet_type` is an integer so the exceptions will wind up reading like "expected NET_SFTP_VERSION. Got package type 5". Also, I think you should be doing "packet" - not "package".

Finally, adding your name to the list of authors because of one commit...  not gonna happen. The AUTHORS section reflects everyone who has commit access to the repo. You do not.

phpseclib has had 70+ individuals make commits: https://github.com/phpseclib/phpseclib/graphs/contributors I can't, in good conscious, add you whilst excluding everyone else.

bittner:
Thanks for the hint! I fixed the package/packet wording issue.

Yes, it should be clear to people dealing with SFTP the the upper case words are constants that correspond to numbers (as usual with protocols). I think that's okay. Better than no number at all.

I'll restore the original AUTHORS file, no problem. (Other projects _require_ you to add yourself to that file. It's always a roulette game decision.)

bittner:
Anything missing to get this PR merged?

terrafrost:
No - it looks better. I'll merge it this weekend!

terrafrost:
I note you replaced `SSH_FXP_*` with `NET_SFTP_*`. I actually kinda prefer `SSH_FXP_*` since that's how the packets are referred to in the [SFTP specs](https://tools.ietf.org/html/draft-ietf-secsh-filexfer-02#page-6). That's why SSH2.php exceptions say stuff like "_Expected SSH_MSG_KEXINIT_" instead of "_Expected NET_SSH2_MSG_KEXINIT_" - because that's how the [SSH2 specs](https://tools.ietf.org/html/rfc4254#section-9) refer to them as.

That said, I've decided to go ahead and merge this. When I get around to significantly refactoring SSH2.php the ideas illustrated in #1160 will change how these constants all work anyway. At that point I could name the constants SSH_FXP_WHATEVER instead of NET_SFTP_WHATEVER.

Are you going to make similar changes to SSH2.php?

Thanks!

bittner:
Thanks for merging!

No, I don't plan to make similar changes to other parts of the code. We had an issue with an SFTP server responding with a value this library doesn't handle, thus aborting the operation. The changes now merged help in diagnosing the specific case.

In general, though, I think I understand and agree: There are [quite a few more places](https://github.com/phpseclib/phpseclib/search?utf8=%E2%9C%93&q=%22Exception%28%27Expected%22&type=) that need the same or a similar fix. We should probably come up with a customized version of the `UnexpectedValueException` to do away with the ever duplicating error message, and pass the expected values and the actual value as parameters to it. And then use that exception for all occurrences in the whole library.

As for the constant names I'm a bit confused. My goal was to match the error messages and the source code. The changed parts read as if there had been copy+paste mistakes, which I tried to address together with making the overall exception message more useful. Sorry for any additional work this caused.

I feel, it's important that the constant names in the code and in the SFTP specs match. Now it's probably worth taking the time to refactor all affected parts of this library and match the real world with all parts of the code.

terrafrost:
> The changes now merged help in diagnosing the specific case.

The way I normally diagnose stuff is by looking at the SSH or SFTP logs. You can enable them by doing `define('NET_SSH2_LOGGING', 2);` and then `$ssh->getLog();` or `define('NET_SFTP_LOGGING', 2);` and `$ssh->getSFTPLog();`.

When people encounter errors I generally ask them to provide the logs and I figure out changes - if any - need to be done based on that.

> We should probably come up with a customized version of the `UnexpectedValueException` to do away with the ever duplicating error message, and pass the expected values and the actual value as parameters to it. And then use that exception for all occurrences in the whole library.

That's a good idea.

> As for the constant names I'm a bit confused. My goal was to match the error messages and the source code. The changed parts read as if there had been copy+paste mistakes, which I tried to address together with making the overall exception message more useful. Sorry for any additional work this caused.

I used NET_SFTP_WHATEVER style constants because the old PEAR CS required constants be "namespaced". The package was in the [Networking category](http://pear.php.net/packages.php?catpid=16&catname=Networking) hence the prefix NET. The package name is SFTP hence NET_SFTP_WHATEVER. Doing `NET_SFTP_SSH_FXP_WHATEVER` seemed redundant so I did `NET_SFTP_WHATEVER` for the constant names and `SSH_FXP_WHATEVER` in the error messages.

It's more of a personal preference like camelCase vs. snake_case.

> Now it's probably worth taking the time to refactor all affected parts of this library and match the real world with all parts of the code.

Right now what I'm working on (between my job, personal activities and other phpseclib stuff) is ECDSA. After that I'll be working on completely refactoring X509. Then I'll be doing SSH2, making it support multiple channels, changing how the constants work (as done in #1160), etc. By the time I'm through with everything the API won't look anything like the API for 1.0 / 2.0.

bittner:
> The way I normally diagnose stuff is by looking at the SSH or SFTP logs. You can enable them by doing define('NET_SSH2_LOGGING', 2); and then $ssh->getLog(); or define('NET_SFTP_LOGGING', 2); and $ssh->getSFTPLog();.

We did that. What I intended to say and to do was, future users shouldn't have to suffer the same pain. The exception should tell what's the root cause right away. For diagnosing it will suffice to read the error message in future.

PR #1206, opened 2017-11-07
Fix for PHP 7 global variable changes

JudeRosario:
As described here http://php.net/manual/en/migration70.incompatible.php

### DIFF
--- a/phpseclib/Net/SFTP/Stream.php
+++ b/phpseclib/Net/SFTP/Stream.php
@@ -197,7 +197,7 @@ class Net_SFTP_Stream
 
         if ($host[0] == '$') {
             $host = substr($host, 1);
-            global $$host;
+            global ${$host};
             if (!is_object($$host) || get_class($$host) != 'Net_SFTP') {
                 return false;
             }


timledlie:
+1, we're preparing to upgrade to PHP7 and would like to see this fixed.

terrafrost:
Give me a few days. I'd like to write a unit test that reproduces the error and then I'll get it merged. It looks like the unit testing for the SFTP Stream class is lacking:

https://github.com/phpseclib/phpseclib/blob/master/tests/Unit/Net/SFTPStreamTest.php
https://github.com/phpseclib/phpseclib/blob/master/tests/Functional/Net/SFTPStreamTest.php

terrafrost:
First, I'm going to assume that you're referring to [global only accepts simple variables](http://php.net/manual/en/migration70.incompatible.php#migration70.incompatible.variable-handling.global). Quoting it,

> [Variable variables](http://php.net/manual/en/language.variables.variable.php) can no longer be used with the [_global_](http://php.net/manual/en/language.variables.scope.php#language.variables.scope.global) keyword. The curly brace syntax can be used to emulate the previous behaviour if required:
> 
> ```php
> <?php
> function f() {
>     // Valid in PHP 5 only.
>     global $$foo->bar;
> 
>     // Valid in PHP 5 and 7.
>     global ${$foo->bar};
> }
> ?>
> ```
> 

Anyway, that doesn't seem to be strictly true:

```php
<?php
function test($var)
{
    global $$var;
    echo $$var;
}

$zzz = 'hello, world!';
test('zzz');
```

That works without even so much as an E_NOTICE on PHP 7.2:

https://3v4l.org/4hqZ4

Now, doing what they did in their example doesn't work in PHP 7 whereas it does in PHP 5:

```php
<?php

class test
{
    var $var;

    function __construct($var)
    {
        $this->var = $var;
        global $$this->var;
        echo $$var. "\n";
    }
}

$zzz = 'hello, world!';
$test = new test('zzz');
```

https://3v4l.org/6CWsM

Of course, what they did in their example isn't what phpseclib is doing here.

I mean, I can still merge this, I guess, but it really doesn't seem to be necessary. And the PHP docs do seem to be a bit misleading.

coderintherye:
@terrafrost Would be great if you could merge it so it is compliant with the official PHP docs / suggested syntax.

terrafrost:
@KonstantinCodes You can already upgrade to PHP 7.2. This commit was only ever a cosmetic change. https://github.com/phpseclib/phpseclib/pull/1206#issuecomment-343739964 elaborates. https://3v4l.org/4hqZ4 shows it running on PHP 7.2 just fine.

That said, I am planning on releasing 2.0.10, probably this weekend, because of #1248, but I'd never do a release based solely on this change because this change is about as impactful as changing the spelling of a word in a comment.

Issue #971, opened 2016-04-14
validateSignature can't find certificate


klatoszewski:
I have problem with validateSignature() function in X509.php file. This function check whether one certificate sign other certificate. I have two certificate that cause validateSignature() function fail. The problem is in this line:
`
if ($this->currentCert['tbsCertificate']['issuer'] === $ca['tbsCertificate']['subject'])
`

In first certificate all properties have type printableString in other certificate all properties have type utf8String and this cause that validateSignature() function can't find parent of certificate. My solution of this problem is repleace this condition by:
`
if($this->getDN(self::DN_STRING, $this->currentCert['tbsCertificate']['issuer']) == $this->getDN(self::DN_STRING, $ca['tbsCertificate']['subject']))
`

In this condition i compare only two string not array with types. See screenshot: http://imgur.com/F10dVvL

terrafrost:
Can you post the certificates?

klatoszewski:
Here is: http://pastebin.com/Mr4YS3DP

terrafrost:
I'll take a look tonight - this was kinda a busy weekend for me.

Thanks!

klatoszewski:
Have you done any investigation about that?

terrafrost:
Yes.

So there are a couple of issues.

First, the intermidate cert's subject DN and in the third certs subject DN the types don't match. The strings have the same content but in the CA cert the OU and CN are utf8String's whereas in the third cert they're printableString's.

In the past it would have validated but that is now no longer the case. The commit that changed this behavior:

https://github.com/phpseclib/phpseclib/commit/340ee0cd2d7bf33316238fed8e9ab2e459d676ce

The reasoning behind that change:

https://github.com/phpseclib/phpseclib/pull/25#issuecomment-9596411

Second, the subject DN of the third cert has a weird DN. postalAddress (2.5.4.16). Here's the ASN.1 definition for the DN's:

```
   Certificate  ::=  SEQUENCE  {
        tbsCertificate       TBSCertificate,
        signatureAlgorithm   AlgorithmIdentifier,
        signatureValue       BIT STRING  }

   TBSCertificate  ::=  SEQUENCE  {
        version         [0]  EXPLICIT Version DEFAULT v1,
        serialNumber         CertificateSerialNumber,
        signature            AlgorithmIdentifier,
        issuer               Name,
        validity             Validity,
        subject              Name,
        subjectPublicKeyInfo SubjectPublicKeyInfo,
[...]

PR #1213, opened 2017-11-23
validateSignature if subjectKeyIdentifier is not set and check serial number if is set

theki:
validateSignature failed in case when subjectKeyIdentifier extension is not set.

According to openssl implementation both (authorityKey.keyIdentifier and subjectKeyIdentifier) needs to be set to perform validation on it. Also if authorityKey.authorityCertSerialNumber is set then needs to be validated.


phpseclib/File/X509.php
```diff
@@ -1137,7 +1138,11 @@ public function validateSignature($caonly = true)
                                 $subjectKeyID = $this->getExtension('id-ce-subjectKeyIdentifier', $ca);
                                 switch (true) {
                                     case !is_array($authorityKey):
+                                    case !$subjectKeyID:
```
terrafrost: What if the signed cert has authority key identifier and no authority cert serial number but the candidate CA cert doesn't have a subject key identifier? I can write a test to be sure but it looks like the old code would have continued looping whereas it looks like this could would think such a cert was the CA cert when it probably ought not be.

phpseclib/File/X509.php
```diff
@@ -1137,7 +1138,11 @@ public function validateSignature($caonly = true)
...
                                     case is_array($authorityKey) && isset($authorityKey['keyIdentifier']) && $authorityKey['keyIdentifier'] === $subjectKeyID:
+                                        if (isset($authorityKey['authorityCertSerialNumber']) && $authorityKey['authorityCertSerialNumber'] != $ca['tbsCertificate']['serialNumber']) {
```
terrafrost: What if `$authorityKey` isn't an array (ie. it isn't set)? It'll then do the `isset($authorityKey['authorityCertSerialNumber'])`, trying to access the index of a null. It seems like this could produce a PHP error (E_NOTICE) but it actually doesn't appear to do so: https://3v4l.org/UFH9D 

```php
<?php
error_reporting(E_ALL);

$a = null;
echo isset($a['test']) ? 'yes' : 'no';
```
Be that as it may I still don't think it ought to basically be doing `isset(null['key'])`.

terrafrost:
> According to openssl implementation both (authorityKey.keyIdentifier and subjectKeyIdentifier) needs to be set to perform validation on it.

I assume you mean the signed cert needs to have both an authority key identifier and a subject key identifier. Here's how I'd implement that:

```php
$authorityKey = $this->getExtension('id-ce-authorityKeyIdentifier');
$CASubjectKeyID = $this->getExtension('id-ce-subjectKeyIdentifier', $ca);
$subjectKeyID = $this->getExtension('id-ce-subjectKeyIdentifier');
switch (true) {
    case !is_array($authorityKey):
    case is_array($authorityKey) && $subjectKeyID && isset($authorityKey['keyIdentifier']) && $authorityKey['keyIdentifier'] === $CAsubjectKeyID:
        $signingCert = $ca; // working cert
        break 3;
}
```

> if authorityKey.authorityCertSerialNumber is set then needs to be validated

Here's how I'd implement that (building on the above):

```php
$authorityKey = $this->getExtension('id-ce-authorityKeyIdentifier');
if (!is_array($authorityKey)) {
    $signingCert = $ca;
    break 2;
}
$CASubjectKeyID = $this->getExtension('id-ce-subjectKeyIdentifier', $ca);
$subjectKeyID = $this->getExtension('id-ce-subjectKeyIdentifier');
if (!$subjectKeyID || !isset($authorityKey['keyIdentifier']) || $authorityKey['keyIdentifier'] !== $CAsubjectKeyID) {
    break;
}
if (isset($authorityKey['authorityCertSerialNumber']) && $authorityKey['authorityCertSerialNumber'] !== $ca['tbsCertificate']['serialNumber']) {
    break;
}
$signingCert = $ca;
break 2;
```

I want to consult the X.509 RFC and see if that says anything on the subject as well.

terrafrost:
> According to openssl implementation both (authorityKey.keyIdentifier and subjectKeyIdentifier) needs to be set to perform validation on it.

Neither [RFC5280 § 4.2.1.1 Authority Key Identifier](https://tools.ietf.org/html/rfc5280#section-4.2.1.1) nor [RFC5280 § 4.2.1.2 Subject Key Identifier](https://tools.ietf.org/html/rfc5280#section-4.2.1.2) say anything about this. Can you post a link to OpenSSL source where this restriction is described or implemented? I didn't check the errata for the RFC.

Also, I think err'ing on the side of caution is a good thing. Worst case: this yields false negatives. Say you have a cert with an authority key identifier and no subject key identifier. Using the current approach the cert would fail validation with phpseclib. You're saying that in OpenSSL the cert would pass validation? That can be put to the test but in lieu of having conducted such a test I'd say that err'ing on the side of caution is the best approach in-so-far as X.509 is concerned.

> if authorityKey.authorityCertSerialNumber is set then needs to be validated

I concur with this. I'll try to whip something up in the next few days as time permits (for my own reference, see File/ASN1/Maps/AuthorityKeyIdentifier.php)

Thanks!

terrafrost:
So here's what I'd do (without doing the subject key identifier thing, which will require some verification):

```php
$authorityKey = $this->getExtension('id-ce-authorityKeyIdentifier');
if (!is_array($authorityKey)) {
    $signingCert = $ca;
    break 2;
}
$CASubjectKeyID = $this->getExtension('id-ce-subjectKeyIdentifier', $ca);
switch (true) {
    case isset($authorityKey['keyIdentifier']) && $authorityKey['keyIdentifier'] !== $CAsubjectKeyID:
    case isset($authorityKey['authorityCertSerialNumber']) && $authorityKey['authorityCertSerialNumber'] !== $ca['tbsCertificate']['serialNumber']:
    case !defined('FILE_X509_IGNORE_TYPE') && $authorityKey['authorityCertIssuer'] === $ca['tbsCertificate']['subject']:
    case defined('FILE_X509_IGNORE_TYPE') && $this->getDN(self::DN_STRING, $authorityKey['authorityCertIssuer']) === $this->getDN(self::DN_STRING, $ca['tbsCertificate']['subject']):
        break 2;
}
$signingCert = $ca;
break 2;
```

phpseclib/File/X509.php
theki: That case will never happen because we are checking first:
if ca.subject == cert.issuer (lines 1135 and 1136)
so if ca.subject and cert.issuer are the same we just have to perform additional checks on fields like serial if are set.

phpseclib/File/X509.php
theki: It's just shortcut to avoid another condition in if statement widely used. But I don't mind to perform another check if will make it more readable

theki:
Long story short is I have ~300k certs and only 80% passed signature validation via phpseclib and all of them passed validation with openssl. So was digging in openssl implementation and found exact function for that:

`int X509_check_akid(X509 *issuer, AUTHORITY_KEYID *akid)
{

    if (!akid)
        return X509_V_OK;

    /* Check key ids (if present) */
    if (akid->keyid && issuer->skid &&
        ASN1_OCTET_STRING_cmp(akid->keyid, issuer->skid))
        return X509_V_ERR_AKID_SKID_MISMATCH;
    /* Check serial number */
    if (akid->serial &&
        ASN1_INTEGER_cmp(X509_get_serialNumber(issuer), akid->serial))
        return X509_V_ERR_AKID_ISSUER_SERIAL_MISMATCH;
    /* Check issuer name */
    if (akid->issuer) {
        /*
         * Ugh, for some peculiar reason AKID includes SEQUENCE OF
         * GeneralName. So look for a DirName. There may be more than one but
         * we only take any notice of the first.
         */
        GENERAL_NAMES *gens;
        GENERAL_NAME *gen;
        X509_NAME *nm = NULL;
        int i;
        gens = akid->issuer;
        for (i = 0; i < sk_GENERAL_NAME_num(gens); i++) {
            gen = sk_GENERAL_NAME_value(gens, i);
            if (gen->type == GEN_DIRNAME) {
                nm = gen->d.dirn;
                break;
            }
        }
        if (nm && X509_NAME_cmp(nm, X509_get_issuer_name(issuer)))
            return X509_V_ERR_AKID_ISSUER_SERIAL_MISMATCH;
    }
    return X509_V_OK;
}`

https://github.com/openssl/openssl/blob/master/crypto/x509v3/v3_purp.c

In the end it works fine when we just add
`case !$subjectKeyID:`

But to be more compliant I've added checking serial (if present) as well.

I skipped last step because I discovered that my leaf certificates have info about Root certificate there - not CA parent. It makes sense to don't check it if we want to verify signature against parent only. OpenSSL requires whole CA chain to verify leaf certificate.
I hope that makes sense.

phpseclib/File/X509.php
terrafrost: > so if ca.subject and cert.issuer are the same we just have to perform additional checks on fields like serial if are set.

That's what I'm saying. Your code isn't doing that. And lines 1135 and 1136 aren't doing the additional checks you speak of.

Here's an example.

The signing cert:

```
-----BEGIN CERTIFICATE-----
MIIB4zCCAUygAwIBAgIUSyCTib61O8k8TM5JA5n3zDYDFs8wDQYJKoZIhvcNAQEF
BQAwHDEaMBgGA1UECgwRcGhwc2VjbGliIGRlbW8gQ0EwIBcNMTcxMjEwMTYxMDIw
WhgPOTk5OTEyMzEyMzU5NTlaMBwxGjAYBgNVBAoMEXBocHNlY2xpYiBkZW1vIENB
MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQCn5XdGJUd9PY9zpksbJ2kD57PL
nQ3ISK+qbeSe2H+Uj0AEDnm611tvXlr9a7zBPNwZ2ali9bczirQkxtu/sftSHCGv
LL6sz1vdYP0Un792o5h+PkbQLnDe1lf/W8Q87GY9cRxUQVVfA7QUfgsl4iWgi7tv
qz3dRmxzBpjpgWvp/QIDAQABoyAwHjALBgNVHQ8EBAMCAQYwDwYDVR0TAQH/BAUw
AwEB/zANBgkqhkiG9w0BAQUFAAOBgQBQJpF/uakVjNIOz4Ve+duZYeeghpVxVNVK
t4OFi8DQzAsdFBpDDg/ZdAoXYOI/VHwNR+jKS2bPLmvNR2dP+OuZVdSTYu4/dZtI
Tp9ju17bym9oyWjX8CApa32+4N4WtcYLaVHOgyGA9f1VRHBtLTc8xcuHnJg6v3m9
TYVkR8c3IQ==
-----END CERTIFICATE-----
```

The signed cert:

```
-----BEGIN CERTIFICATE-----
MIIB5TCCAU6gAwIBAgIUVCrbYZRButS1ka5fyUe2iOzFAmwwDQYJKoZIhvcNAQEF
BQAwHDEaMBgGA1UECgwRcGhwc2VjbGliIGRlbW8gQ0EwIBcNMTcxMjEwMTYxMDIw
WhgPOTk5OTEyMzEyMzU5NTlaMB4xHDAaBgNVBAoME3BocHNlY2xpYiBkZW1vIGNl
cnQwgZ8wDQYJKoZIhvcNAQEBBQADgY0AMIGJAoGBANkmWIAsKkL1Bv3PiFA65IRQ
EOSc65rui70M5tvrDpka1jl6QgvGY/jI/6D6h8sZBxyYyncgWaCA4dVv9dmdZm8Z
JPXACUfHTJuk7ApPq2oYjyheyQkSLp8kYP47cidbNfWk3RkTRqwleh3ByG4e1Qxh
sfdf5owA1yuW5JSygppvAgMBAAGjIDAeMA4GA1UdIwQHMAWAA3p6ejAMBgNVHQ4E
BQQDenp6MA0GCSqGSIb3DQEBBQUAA4GBACMLNye3bEzTnHJAroAiV8Rw9rL6QG+x
wrFDBq173MXTAnKar/TujUxfmOzNZCxjPHh6oKRosOWZuI5v8/0wG9I6irYwLXOw
kXMvXT91/m5Vk7rLCqQ+zsFncxg12leHRgglWh1wsJOhPI/B0mYgCA04RNHsm+iC
/XjRBlAt1kMb
-----END CERTIFICATE-----
```

phpseclib doesn't recognize this cert as being valid (and I'm not sure it ought to) but in your changes it is recognized as valid. The signing cert doesn't have a subject key identifier while the signed cert does have an authority key identifier (and a subject key identifier).

That said, to be fair, OpenSSL seems to think the cert is valid:

```
openssl verify -verbose -CAfile signer.pem signed.pem
```

That said, idk that I agree with OpenSSL's verdict. In light of this I've posed the question on crypto.stackexchange.com:

https://crypto.stackexchange.com/q/53856/4520

terrafrost: So I've re-evaluated things after the crypto.stackexchange.com and I'll go ahead and merge this. I'll update it based on my comment at https://github.com/phpseclib/phpseclib/pull/1213#discussion_r153625564 , I'll backport to 1.0 and I'll see if this should be added for CRL's (`tbsCertList`).

theki: Great! Thanks a lot!

terrafrost:
So the unit test that is failing with this PR was added as a result of #971. As I explained in the comment I just posted in that ticket I'll be removing the failing unit test.

terrafrost:
I backported this to 1.0 (squashing the commits and then cherry-picking into 1.0), added this to the CRL signature validation routines and this is now in the main phpseclib repo (all branches).

Thanks!

theki:
Thank you :)

Issue #1092, opened 2017-02-03
Consider Static Analysis as part of Continuous Integration


paragonie-scott:
In some of my projects (especially when crypto/security is involved), I've added [psalm](https://github.com/vimeo/psalm) to my Travis configuration to ensure type safety before any code is changed.

* https://github.com/paragonie/random_compat/pull/123
* https://github.com/paragonie/sodium_compat/pull/15

The downside is that some of the things phpseclib does right now (mostly `extract()`) will have to be replaced with more explicit variable declarations.

terrafrost:
I'll take a look as time permits! I'm in Florida for the next to weeks. Conference this week, Disneyworld next week :)

bantu:
I'd be much in favour of getting rid of `extract()`.

PR #1222, opened 2017-12-04
Add PhpDoc @var for extract() functions

rrran:


phpseclib/Math/BigInteger/Engines/PHP64.php
```diff
@@ -192,7 +192,7 @@ public function extendedGCD(PHP64 $n)
      */
     public function gcd(PHP64 $n)
     {
-        extract($this->extendedGCD($n));
+        $gcd=$this->extendedGCD($n)['gcd'];
         return $gcd;
```
terrafrost: I think I'd prefer just doing `return $this->extendedGCD($n)['gcd'];`. I can change that myself for both PHP32 / PHP64.

terrafrost:
So I think some of the changes are good but some of the changes I'm more on the fence about.

Like the change that caused the unit test to fail for example. That could be fixed by putting the following above `compact` in Crypt/RSA/Keys/PuTTY::load:

```php
        if (isset($components['comment'])) {
            $comment = $components['comment'];
        }
```
But we'd have to do that for Crypt/DSA/Keys/PuTTY::load as well. It's like the change is increasing the amount of boilerplate that's needed and idk that I'm a big fan of that.

On the flip side, less reliance on `extract` would be well received. See, for example, #1092.

I guess I'm willing to do away with the extract and replace it with extra boilerplate.

You also seem to be using two different standards for extract's with one parameter. From Net/SSH2.php:

```php
        /** @var integer $first_kex_packet_follows */
```

...and...

```php
                /**
                 * @var integer $length
                 */
```

I'd kinda prefer it if their was a single unified standard. ie. all one parameter extract's have a one-line comment or all one parameter extract's have a multi-line comment. Not half use one standard, half use another, but all use the same standard.

To summarize, there are two things that I think it'd be cool if you could do before I merge this:

1. Fix the unit test by adding the additional boilerplate to both RSA/Keys/PuTTY.php and DSA/Keys/PuTTY.php that I spoke of
2. Make all the comments for the one parameter extract's take up just one line or three lines. Make it consistent.

There was one other personal preference thing I found but I can fix that myself:

https://github.com/phpseclib/phpseclib/pull/1222/commits/c705e7d9c60061091f4ca29aafdb92b65178e79e#r157345472

rrran:
Done! 
/**
* ......
*/
converted to /* ..... */
Tests passed.
And thumbs up to the getting rid of extract()

PR #1238, opened 2018-01-04
Get rid of Classes and all that slow Reflection stuff.

rrran:
Getting rid of Reflection makes code a bit faster and more readable


terrafrost:
This is not a change I will be making at this time. https://github.com/phpseclib/phpseclib/pull/1084#pullrequestreview-17805854 elaborates but to provide a more succinct summary...  this approach will not be used when the master branch is tagged 3.0.0 or the 3.0 branch created from it. This approach is more of a place holder until a better approach can be found.

SSH2.php (which is what primarily uses this) has not been significantly rewritten. Portions of the master branch have already been significantly rewritten (eg. RSA.php). A significant rewrite is on the horizon for SSH2.php but right now my priority is implementing ECDSA (prime finite fields over short weiestrass curves are mostly done; still need to do twisted edwards curves and binary finite fields). After that my focus will be X509.php. _Then_ I'll do SSH2.php and that's when I'll come up with a better approach for dealing with this. Until then -and as I said - it's more of a place holder. A way to make sure that variables that should not be public are not public whilst not breaking unit tests.

In my earlier elaboration I suggest that one possibility is that the unit tests for SSH2.php could simply be deleted but I think they're still useful as they help identify merge conflicts arising from merging the 2.0 branch.

rrran:
Cool! Looking forward for 3.0 :)

PR #1242, opened 2018-01-17
add php nightly tests

andreybolonin:


terrafrost:
Something I'm curious about...

In https://github.com/phpseclib/mcrypt_compat/commit/33b727da70895e160f04194fc749f53821ce18be nightly tests were added but failures were allowed in them. In this one you add nightly tests but aren't allowing failures.

Do you have any thoughts on that approach vs this approach?

PR #1261, opened 2018-04-03
PHP Fixes found by Static Analysis Tool

craiglondon:


craiglondon:
I think the only commit related to Amy performance (supposed) improvements is this one.  

https://github.com/phpseclib/phpseclib/commit/57c87fc9bc4bb66175a08fdb8ee23d531a19d7ef  

I haven't done any benchmarks before but would be willing to do some given a little direction as a best tool to use?   like using Blackfire and PHPunit?

terrafrost:
I've never used Blackfire lol. The way I normally benchmark stuff is just to loop over it a million times lol. And in this case I'm not sure a loop would be the most effective way to do the benchmark since I don't really expect anyone to do that stuff in a loop.

I guess two possible ideas for benchmarking:

1. Upload a 500mb file to an SFTP server with the changes and without.
2. Maybe try to decode a really big CRL with File/X509.php with and without the changes.

yarkm13:
Why test performance of comparison operators? It's known that strict comparison is faster and benchmark proves it (approx 15% faster) https://github.com/8ctopus/php-benchmark

Typehint performance difference is negligible 0.4 ms per 100M cycles

PR #1287, opened 2018-08-13
Ensure SFTP::rawlist gives same results regardless of statcache

dsas:
Previously SFTP::rawlist gave different results depending on whether the
stat cache was enabled or not. With the stat cache turned off it did not
treat directories correctly, as it relied on the stat cache even though
it was not populated.


terrafrost:
I cherry-picked this into the 1.0 branch and merged this into 2.0 and master.

I also fixed the failing unit test (changed `stat` to `lstat`; the issue is that an earlier method creates a symlink in `self::$scratchDir`)

Thanks!

PR #1297, opened 2018-09-11
Fixes #1296 parsing invalid certificate

### LINKED ISSUE #1296: Infinite loop in ASN1::decode_ber()
If the certificate passed in is random data (or generally invalid) recursive calls of self::decode_ber can return false.  The response of the recursive call is not type checked leading to an infinite loop breaking on a memory limit exceeded.

xepa:
When parsing invalid (ie corrupted) certificate ASN1::decode_ber() will go into an infinite loop as the result of recursive calls to decode_ber could be false (and not checked).  This will lead to a memory limit exception.

Tests and fixes Infinite loop in ASN1::decode_ber() #1296

### DIFF
--- a/phpseclib/File/ASN1.php
+++ b/phpseclib/File/ASN1.php
@@ -333,6 +333,9 @@ abstract class ASN1
                 $remainingLength = $length;
                 while ($remainingLength > 0) {
                     $temp = self::decode_ber($content, $start, $content_pos);
+                    if (false == $temp) {
+                        break;
+                    }
                     $length = $temp['length'];
                     // end-of-content octets - see paragraph 8.1.5
                     if (substr($content, $content_pos + $length, 2) == "\0\0") {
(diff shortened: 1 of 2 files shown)


phpseclib/File/ASN1.php
```diff
@@ -333,6 +333,9 @@ private static function decode_ber($encoded, $start = 0, $encoded_pos = 0)
                 $remainingLength = $length;
                 while ($remainingLength > 0) {
                     $temp = self::decode_ber($content, $start, $content_pos);
+                    if (false == $temp) {
+                        break;
```
terrafrost: No where else in the phpseclib are [yoda conditions](https://en.wikipedia.org/wiki/Yoda_conditions) used. As such, for consistency purposes, I don't think they ought to be used anywhere. ie. they ought to be used everywhere or nowhere.

Also, every other condition does `===` but this one does `==`.

Finally, I think `return false` would be better than `break`.

That said, don't worry about these changes - I cherry-picked this branch into the 1.0 branch and made the changes there. I then merged to 2.0 and master.

Thanks!

terrafrost:
I cherry picked this and merged into 1.0 / 2.0 / master.

Thanks!

phpseclib/File/ASN1.php
xepa: Ah .. good catch and thnx for fixing it and the pull.

Issue #1298, opened 2018-10-02
Connections in long running processes time out, but isConnected() stays true


sebsel:
This is probably related to #1126 and thus #985.

A more detailed description of my problem can be found in this issue on the Flysystem SftpAdapter, which uses this library to provide easy sFTP access:
https://github.com/thephpleague/flysystem-sftp/issues/66

The main point is that, when used in a long running process, a SSH connection can drop due to inactivity. Before every attempt, Flysystem checks the connection. The Flysystem adapter fails to reconnect, because `->isConnected()` (in `\phpseclib\Net\SFTP` which inherits the method from `\phpseclib\Net\SSH2`) still returns `true` after the connection has timed out.

I don't think sending a ping (like suggested in #1126) would be a good solution to this. If there is some way to detect the disconnection in `->isConnected()`, the problem would be solved already.

terrafrost:
I think something that would be better would be something like adding `$this->bitmap = 0;` every time `user_error('Connection closed by server');` is called. That's what `->isConnected()` looks at.

Another possibility. You could do something like this (this is for phpseclib 1.0 but can be easily adapted to work on phpseclib 2.0):

```php
<?php
include('Net/SSH2.php');

define('NET_SSH2_CHANNEL_KEEP_ALIVE', 100);

class Net_SSH2_Ping extends Net_SSH2 {
    /**
     * Authentication Credentials
     *
     * @var array
     * @access private
     */
    var $auth = array();

    /**
     * Login
     *
     * The $password parameter can be a plaintext password, a Crypt_RSA object or an array
     *
     * @param string $username
     * @param mixed $password
     * @param mixed $...
     * @return bool
     * @see self::_login()
     * @access public
     */
    function login($username)
    {
        $args = func_get_args();
        $this->auth[] = $args;
        return call_user_func_array(array(&$this, 'parent::login'), $args);
    }

    /**
     * Pings a server connection, or tries to reconnect if the connection has gone down
     *
     * Inspired by http://php.net/manual/en/mysqli.ping.php
     *
     * @return string
     * @access public
     */
    function ping()
    {
        if (!$this->isAuthenticated()) {
            return false;
        }

        $this->window_size_server_to_client[NET_SSH2_CHANNEL_KEEP_ALIVE] = $this->window_size;
        $packet_size = 0x4000;
        $packet = pack(
            'CNa*N3',
            NET_SSH2_MSG_CHANNEL_OPEN,
            strlen('session'),
            'session',
[...]

PR #1313, opened 2018-11-09
fix order of user_error() and bitmap reset

sebsel:
In some configurations of PHP (at least in mine), execution stops after a `user_error()`, so the bitmap is never set to 0 in that case. I think it's best just always first set the bitmap and then do the error.

### DIFF
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -3335,8 +3335,8 @@ class SSH2
     function _get_binary_packet($skip_channel_filter = false)
     {
         if (!is_resource($this->fsock) || feof($this->fsock)) {
-            user_error('Connection closed prematurely');
             $this->bitmap = 0;
+            user_error('Connection closed prematurely');
             return false;
         }
 
@@ -3379,8 +3379,8 @@ class SSH2
         while ($remaining_length > 0) {
             $temp = stream_get_contents($this->fsock, $remaining_length);
             if ($temp === false || feof($this->fsock)) {
-                user_error('Error reading from socket');
                 $this->bitmap = 0;
+                user_error('Error reading from socket');
                 return false;
             }
             $buffer.= $temp;
@@ -3398,8 +3398,8 @@ class SSH2
         if ($this->hmac_check !== false) {
             $hmac = stream_get_contents($this->fsock, $this->hmac_size);
             if ($hmac === false || strlen($hmac) != $this->hmac_size) {
-                user_error('Error reading socket');
                 $this->bitmap = 0;
+                user_error('Error reading socket');
                 return false;
             } elseif ($hmac != $this->hmac_check->hash(pack('NNCa*', $this->get_seq_no, $packet_length, $padding_length, $payload . $padding))) {
                 user_error('Invalid HMAC');
@@ -3936,8 +3936,8 @@ class SSH2
     function _send_binary_packet($data, $logged = null)
     {
         if (!is_resource($this->fsock) || feof($this->fsock)) {
-            user_error('Connection closed prematurely');
             $this->bitmap = 0;
+            user_error('Connection closed prematurely');
             return false;
         }
 


bantu:
Hmm. If execution stops, the value of bitmap is irrelevant. Can you elaborate?

bantu:
0c26415f932aedc5b77e9dae39c20e3e45f4b5ad went into 1.0, so this should too.

bantu:
Feel free to rebase onto 1.0. Patch applies cleanly.

sebsel:
Steps to reproduce:
1. create a fresh install of the Laravel framework (with the [Laravel installer](https://laravel.com/docs/5.7/installation#installing-laravel) this is `laravel new fresh`, for a project called 'fresh')
2. do a `composer require league/flysystem-sftp` to enable Sftp support (just found out that this is all you need in the new Laravel 5.7, yaay)
3. open `config/filesystems.php` and add this item in the array under `'disks' => [`:

    ```
    'sftp' => [
        'driver' => 'sftp',
        'root' => '/home/user/',
        'host' => 'example.com',
        'username' => 'user',
        'password' => 'password',
    ],
    ```

    ... with the right configuration for your server.

4. Do a `php artisan tinker` to open the build-in REPL. Then call `Storage::disk('sftp')->files();`. It will list your files.

5. SSH into the machine yourself, do a `ps faux`. There will be two SSH-sessions, one running `ps faux`. Kill the other with `sudo kill 123` where 123 is the pid.

6. Call `Storage::disk('sftp')->files();` again: it will fail with 'Connection closed prematurely'. Call `Storage::disk('sftp')->getAdapter()->isConnected();`: it will be `true`. Call `->files()` again: it will fail again.

Now, exit `tinker`, apply patch from PR in the `vendor` folder, and start again at 4. Step 6 will now look like:

- Call `Storage::disk('sftp')->files();` again: it will fail with 'Connection closed prematurely'. Call `Storage::disk('sftp')->getAdapter()->isConnected();`: it **will be `false`**. Call `->files()` again: it will **list your files again**.

Which means: in a long-running queue worker when connection has dropped due to a night of no-activity, failed jobs that are retried will start working again. Also: the first job of the day fails, but next jobs will have connection again.

bantu:
@sebsel Thanks a lot for your detailed explanation.

> I'm a bit confused by how the branches relate. First off: I can't rebase onto 1.0 cleanly. It gives me a lot of conflicts to resolve. It would be faster if I got a new branch from 1.0 and manually fix the thing. I can do that.

The idea is to forward merge 1.0 into 2.0 into master.

> But next: on 2.0 there are, as you can see in this PR, 4 places where there is this order of error and reset, but on 1.0 there is just one. Fixing this for 1.0 won't fix those cases in 2.0, and the package I'm using in my project is using 2.0 ([per it's composer.json](https://github.com/thephpleague/flysystem-sftp/blob/master/composer.json#L14)).

You're absolutely right. I was confused because cherry-picking the commit onto 1.0 does not result in conflicts. I've submitted PR #1314 for 1.0.

> The problem I try to solve is described in short in #1298 and more in dept in [thephpleague/flysystem-sftp#66](https://github.com/thephpleague/flysystem-sftp/issues/66).
> 
> I think there is no harm in first doing the `$this->bitmap = 0` and then the `user_error()`. I'm not adding, only rearranging. But I do think it has to be like this for all the 4 places in 2.0 and the 1 remaining place in 1.0.

I am fine with this. However ...

> It's not really that the execution stops, but it throws, which means the next line won't run. If the throw is caught, `$bitmap` remains `15` (in my case) so `isConnected()` will stay `true`, meaning flysystem-sftp will not try to reconnect.

I think, if the behaviour of user_error() is not reliable, we should not use it in the first place.

terrafrost:
> I think, if the behaviour of user_error() is not reliable, we should not use it in the first place.

The issue is presumably due to the use of a custom error handler.

Consider the following code:

```php
<?php

function error_handler($errno, $errstr, $errfile, $errline)
{
    exit("all done!\n");
}

set_error_handler('error_handler');

trigger_error('sample error', E_USER_NOTICE);

echo "this far\n";
```

The "this far" never gets output whereas the "all done!" does.

Issue #892, opened 2015-12-07
Windows OpenSSH navigation


mwienk:
Hi,

I cannot use phpseclib as an SFTP client to a Windows server. This is is caused by the fact that the SFTP::_realpath() uses a / in front of the path. Commenting out 

```
if ($path[0] != '/') {
    $path = $this->pwd . '/' . $path;
}
```

and changing

```
return '/' . implode('/', $new);
```

into

```
return implode('/', $new);
```

Lets it all work. But I don't think those statements were made just for fun. Is there any possibility on checking if the root path (/) is actually that system's root path?

terrafrost:
What are you using as your SFTP server software? I've tried phpseclib with a number of different SFTP clients but I don't have a comprehensive list.

mwienk:
I installed sftp as per https://winscp.net/eng/docs/guide_windows_openssh_server, which uses the https://github.com/PowerShell/Win32-OpenSSH sftp server

terrafrost:
The server's implementation of SFTP is buggy. I created a ticket for them. See https://github.com/PowerShell/Win32-OpenSSH/issues/48 . In theory, I could look at the server identifier and handle their server differently but that doesn't work either per https://github.com/PowerShell/Win32-OpenSSH/issues/49 .

So for the time being I'd just stick with your solution until they fix their bugs.

terrafrost:
What I didn't mention in the PowerShell bug reports is that I also tested this on http://www.freesshd.com/ (which identifies as SSH-2.0-WeOnlyDo 2.4.3) and https://www.bitvise.com/ssh-server-download (which identifies as SSH-2.0-SilverSHielD), both of which run on Windows, and neither of them behave as https://github.com/PowerShell/Win32-OpenSSH does.

ie. PowerShell's behavior is departing even from more established Windows SFTP servers.

mwienk:
Ohw wow, thanks for your effort! I'll install another server, I never guessed that the server would've been the problem.

PR #1328, opened 2018-12-20
Makes the SFTP protocol more permissive(AWS SFTP-S3) like other implementations

g-ferreira-dev:
if we find an error status keep reading till the current package ends before calling it quits, fixes AWS SFTP-S3 service and likely some others as well

[Example from SFTP-S3](https://i.imgur.com/ZFZlbPO.png)

### DIFF
--- a/phpseclib/Net/SFTP.php
+++ b/phpseclib/Net/SFTP.php
@@ -2255,16 +2255,11 @@ class SFTP extends SSH2
                 break;
             }
 
-            $clear_responses = false;
+            $errorPacket = false;
             while ($i > 0) {
                 $i--;
-
-                if ($clear_responses) {
-                    $this->get_sftp_packet();
-                    continue;
-                } else {
-                    $response = $this->get_sftp_packet();
-                }
+                
+                $response = $this->get_sftp_packet();
 
                 switch ($this->packet_type) {
                     case NET_SFTP_DATA:
@@ -2278,9 +2273,9 @@ class SFTP extends SSH2
                         $temp = null;
                         break;
                     case NET_SFTP_STATUS:
-                        // could, in theory, return false if !strlen($content) but we'll hold off for the time being
                         $this->logError($response);
-                        $clear_responses = true; // don't break out of the loop yet, so we can read the remaining responses
+                        // maybe check for SSH_FX_OK and not set the error flag
+                        $errorPacket = true;
                         break;
                     default:
                         if ($fclose_check) {
@@ -2292,7 +2287,8 @@ class SFTP extends SSH2
                 $response = null;
             }
 
-            if ($clear_responses) {
+            // if there was an errorPacket stop reading
+            if ($errorPacket) {
                 break;
             }
         }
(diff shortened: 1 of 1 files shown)


terrafrost:
I don't have a problem with the change per say but could you provide the SFTP logs showing the necessity of the change? The screenshot gives me a general idea but I'd rather see the log files.

Overall it's a bit disconcerting that an SFTP server would respond with a NET_SFTP_DATA packet for a 32kb read request, x+y bytes into a given file whilst responded with an NET_SFTP_STATUS packet for a 32kb read request, x bytes into the same file. If it's gonna be doing that then it kinda makes one wonder if even breaking out of the loop that sends the read requests is appropriate. How are you to know if there's gonna be a mysterious NET_SFTP_DATA packet sent back if you sent an a read request x+y+z bytes into the file?

Most SFTP clients, I imagine, send 1x read request and read the 1x response immediately. They'd start from offset 0 and would, I imagine, do this until either they had read x bytes, where x was the size of the file, or until they got an error (NET_SFTP_STATUS packet) back. How would a regular SFTP client deal with the scenario this PR is attempting to make phpseclib work with? My expectation is that it wouldn't work at all.

I mean, maybe there's a valid use case for AWS SFTP-S3 service to behave this way but it almost seems more likely that this is a bug. This wouldn't be the first time I got a ticket about something that turned out to be a bug about a third parties implementation of SFTP. See #892 for example.

gadd-ferreira:
Thanks for taking your time to review this.

Let me start by saying that i am familiar with the SFTP spec, and i usually prefer the more strict approach to protocol implementations, but we needed it working before the holidays.

I agree this is likely a bug, i've spoken with an AWS representative about it and he will route that to the right people. However phpseclib was the only implementation of the ones we use that displayed this behaviour while downloading files from their endpoint.

This patch was based on some fast sampling of other open source implementations found on google and after running some tests, we decided that it was good enough for our use case.

If this change does not break existing implementations and supports misbehaving servers maybe there is no harm other than enabling these bad implementations, but that could be the worst option on the long run.

I will be happy to study the spec and address your techinical points if you choose to implement this and i understand if you decide against it.

Happy Holidays.

terrafrost:
I'd still like to see the SFTP logs to get a better understanding of what's going on. If you could provide them that'd be helpful!

To get them you can do `define('NET_SFTP_LOGGING', 2);` at the top and then `$sftp->getSFTPLogs();` after the issue has manifested itself.

The logs will give me more info into the issues and will better enable me to see if the proposed fix is the best fix.

Thanks!

terrafrost:
I think https://github.com/phpseclib/phpseclib/commit/fce6063de6844904b0a3facf3ba63a17f6f7b69d likely fixes the issue this was attempting to resolve.

g-ferreira-dev:
Yeah the referenced commit likely solves this issue, makes sense in hindsight.

I am not sure about the correct protocol here but i think we can close this pull request?

Anyway, thanks a lot for your work on phpseclib!

PR #1341, opened 2019-02-05
Whitelist OID 1.3.6.1.4.1.11129.2.4.2

stayallive:
This OID is used in some intermediate certificates from the Dutch government and allows for parsing and saving/validating these certificates with phpseclib.

This is the certificate I encountered: [crt.sh](https://crt.sh/?id=54959165) / [Censys](https://censys.io/certificates/85363a24cb1b66e6cf6244e87d243dbb8306f607357c614cb9c4c224a0e04358)

Happy to add a test for this if needed.

Or if it's possible to disable the validation of the OID's that would also be a great solution. But this patches it for now.

### DIFF
--- a/phpseclib/File/X509.php
+++ b/phpseclib/File/X509.php
@@ -907,7 +907,7 @@ class X509
             // https://tools.ietf.org/html/rfc6962
             case '1.3.6.1.4.1.11129.2.4.2':
             // "Qualified Certificate statements"
-            // https://tools.ietf.org/html/draft-ietf-pkix-qc-03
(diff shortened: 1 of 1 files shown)


terrafrost:
I squashed these two commits into one commit, cherry picked into 1.0 and then merged 1.0 into 2.0 and 2.0 into master:

https://github.com/phpseclib/phpseclib/commit/9a0a9a009669f2d7496b0361653d94c6afedc259

Thanks!

stayallive:
Great, thanks!

PR #1346, opened 2019-02-26
Call xml_parser_free and unset to avoid memory leaks

zacherystuart:
In php7+ it's required to unset the parser after calling xml_parser_free to avoid memory leaks.

I added xml_parser_free and the unset to force proper behavior around the xml parser.

https://secure.php.net/manual/en/function.xml-parser-free.php
https://bugs.php.net/bug.php?id=76874


terrafrost:
I squashed this into one commit and then cherry-picked it into the 1.0 branch and then merged that branch into 2.0 and master:

https://github.com/phpseclib/phpseclib/commits/1.0

Thanks!

PR #1349, opened 2019-03-08
Fix x509 OpenSSL format when multiple OU's are present

miniyarov:
When formatting DN in OpenSSL format Multiple Organizational Unit Names (OU) in certificates throw `PHP Notice:  Undefined index: id-at-organizationalUnitName` and returned array only contains the last OU.


terrafrost:
I cherry picked this into the 1.0 branch and merged it into the 2.0 branch and the master branch:

https://github.com/phpseclib/phpseclib/commit/7c894b2b99f57c0dddb54c0f1af87a1cd9383e5b

Thanks!

Issue #397, opened 2014-06-29
Implement chacha20-poly1305@openssh.com SSH2 transport cipher


bantu:
Available since OpenSSH 6.5 - http://www.openssh.com/txt/release-6.5

This is probably almost as fast as arcfour128 while being a lot more secure and should possibly replace all the "rather weak" arcfour ciphers. I have not taken a too deep look into this, though.

Also see #396

bantu:
RFC7465 now prohibits the use of RC4 in TLS.

paragonie-scott:
It's also side-channel resistant. :+1:

Do you need help with this one?

bantu:
> Do you need help with this one?

Feel free to (help) implement this via libsodium. I believe this will be a bit more complex than https://github.com/phpseclib/phpseclib/pull/755/files?w=0 though.

terrafrost:
I'd prefer this be implemented as it's own stand-alone cipher. eg. Crypt/ChaCha20.php or some such. Certainly libsodium could be used if it's available on the system (as indeed `sha256()` is used in the 1.0 branch on systems using suhosin) but - and especially in the case of symmetric ciphers - I don't want support for symmetric ciphers being added to SSH2.php that don't natively exist in phpseclib.

I don't mind libsodium being used as much for #755 since there's currently no framework in place for any sort of DH / ECDH within phpseclib. But in this case there is a framework for symmetric ciphers to be added. They go in the Cipher/ directory.

It's definitely more work but half the idea of phpseclib is that it doesn't require dependencies. Like mcrypt's increasing obsolescence isn't a huge issue for phpseclib since it doesn't /need/ mcrypt. I suppose libsodium wouldn't be a dependency per say since it could very well only be used if it's available but I also don't want to hack up Net/SSH2.php, making it so it either uses a Crypt/*.php file or a libsodium instance.

Basically, I think it ought to be done right or not done at all and I don't think bypassing the existing framework and just hacking in a one-off solution into Net/SSH2.php is doing it right.

mcarbonneaux:
a first approch ?!
https://github.com/fpoirotte/pssht/blob/master/src/ChaCha20.php 
https://github.com/fpoirotte/pssht/blob/master/src/Poly1305.php

terrafrost:
Looks impressive! Not just ChaCha20 / Poly1305 but the whole thing! I'm actually most impressed by the compression. I didn't even think it'd be possible without a pure-PHP implementation of the algorithm:

[...]

Issue #396, opened 2014-06-29
Add Salsa/ChaCha support


bantu:
http://en.wikipedia.org/wiki/Salsa20#ChaCha_variant

PR #1354, opened 2019-03-21
add Salsa20 / ChaCha20 / Poly1305 support

terrafrost:
This addresses #396 / #397.

Here's an example of how to use it:

```php
$c = new ChaCha20;
$c->setKey('...');
$c->setNonce('...');
//$c->setCounter(...);
echo bin2hex($c->encrypt('...'));
```

Note how it uses `setNonce()` instead of `setIV()`.

Poly1305 can be enabled by either doing `$c->enablePoly1305();` (at which point the key is auto-determined using the AEAD construction described in [RFC8439](https://tools.ietf.org/html/rfc8439#section-2.8)) or `$c->setPoly1305Key('...');` (the latter is necessary for SSH's use of ChaCha20-Poly1305).

AAD's can be set by doing `$c->setAAD()` and the tag can be set (for verification during decryption) by doing `$c->setTag()`. To retrieve the tag after encryption you'd do `$c->getTag()`.

### DIFF
--- a/phpseclib/Crypt/ChaCha20.php
+++ b/phpseclib/Crypt/ChaCha20.php
@@ -1,0 +1,783 @@
+<?php
+
+/**
+ * Pure-PHP implementation of ChaCha20.
+ *
+ * PHP version 5
+ *
+ * @category  Crypt
+ * @package   ChaCha20
+ * @author    Jim Wigginton <terrafrost@php.net>
+ * @copyright 2019 Jim Wigginton
+ * @license   http://www.opensource.org/licenses/mit-license.html  MIT License
+ * @link      http://phpseclib.sourceforge.net
+ */
+
+namespace phpseclib\Crypt;
+
+use phpseclib\Crypt\Common\StreamCipher;
+use phpseclib\Exception\InsufficientSetupException;
+use phpseclib\Exception\BadDecryptionException;
+
+/**
+ * Pure-PHP implementation of ChaCha20.
+ *
+ * @package ChaCha20
+ * @author  Jim Wigginton <terrafrost@php.net>
+ * @access  public
+ */
+class ChaCha20 extends Salsa20
+{
+    /**
+     * The OpenSSL specific name of the cipher
+     *
+     * @var string
+     */
+    protected $cipher_name_openssl = 'chacha20';
+
+    /**
+     * Test for engine validity
+     *
+     * This is mainly just a wrapper to set things up for \phpseclib\Crypt\Common\SymmetricKey::isValidEngine()
+     *
+     * @see \phpseclib\Crypt\Common\SymmetricKey::__construct()
+     * @param int $engine
+     * @access protected
+     * @return bool
+     */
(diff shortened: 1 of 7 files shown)


terrafrost:
One kinda unexpected thing about this: one of the big reasons to use chacha20-poly1305 is, traditionally, speed. But due to the way way SSH uses poly1305 libsodium can't be used. I mean, libsodium can be used for the chacha20 part of chacha20-poly1305 but not for the poly1305 bit. OpenSSL can't be used, either, because, in my experiments, OpenSSL's PHP bindings do not support poly1305 (as https://stackoverflow.com/q/54809652 would seem to confirm).

As such it's faster, with SSH to use AES, even if OpenSSL or mcrypt or whatever aren't available. The thing is...  AES uses a traditional HMAC. eg. sha256, sha1, md5, whatever. These traditional HMACs can make use of http://php.net/hash (which is an enabled by default extension,) whereas poly1305 can't. That fact means that even a pure-PHP AES implementation with a traditional HMAC is a lot faster than a pure-PHP poly1305 with anything.

bantu:
Some thoughts:
* Are nonce and IV the same concept?
* Can we get native support for those algorithms from libsodium-php? Is that supported by the implementation?

terrafrost:
> Are nonce and IV the same concept?

Somewhat lol.

In block ciphers an IV has a very specific meaning and function. ChaCha20, however, is a stream cipher. It works surprisingly similar to CTR mode but none-the-less it's designed as a stream cipher.

The literature for ChaCha20 refers to nonces - not to IVs.

The other big AEAD algorithm, AES-GCM, is basically AES in CTR mode with the GHASH algorithm providing authentication. In that mode the IV for CTR mode is derived from the nonce. In the case of ChaCha20...  it's different lol. You concatenate the key and the nonce and other stuff and then you run it through the 10x odd / even rounds (10 * 2 = 20, hence the 20 in ChaCha20). Each subsequent "block" increments the counter. idk. I guess it's because of the documentation that I'm ultimately going with the term nonce idk lol

> Can we get native support for those algorithms from libsodium-php? Is that supported by the implementation?

I think you're referring to sodium_compat. We could but we can't use it in SSH with the way libsodium / sodium_compat implement it. In particular, SSH calculates the poly1305 result differently than libsodium calculates it. In particular, with SSH the poly1305 is just the concatenation of the AAD with the ciphertext. With libsodium you null pad the AAD and ciphertext and then you append to that the length of both. So altho libsodium can be and is used for the ChaCha20 stuff it can't be used for the Poly1305 stuff.

bantu:
> I think you're referring to sodium_compat. 

I think I was referring to https://github.com/jedisct1/libsodium-php

terrafrost:
> I think I was referring to https://github.com/jedisct1/libsodium-php

I assume that's pretty much the same libsodium extension that's now included in PHP by default?: http://php.net/manual/en/book.sodium.php

bantu:
Yes. I think so.

Issue #1132, opened 2017-05-26
text removed


lsmith77:
relates to https://github.com/paragonie/EasyRSA/issues/12

PR #1369, opened 2019-05-19
make RSA / DSA / ECDSA immutable and add support to SSH2 / X509

terrafrost:
This change makes it so that RSA / DSA / ECDSA keys are immutable. Previously you'd load keys thusly:

```php
$key = new ECDSA;
$key->load('...');
```

Now you'll do so thusly:

```php
$key = PublicKeyLoader::load('...');
```

Previously you needed to know the type of the key you were loading. Now you don't have to. Further, if you pass an X.509 certificate to `PublicKeyLoader::load()` it'll extract the public key portion of the cert, which is something people do periodically try to do ([example](https://stackoverflow.com/q/53618486/569976)).

Also, previously, `RSA::createKey()` would return an array. Now it just returns the private key. If you want to get the public key you'll need to do `$key->getPublicKey()`.

To identify whether or not a key is a public or private key you can do `$key instanceof \phpseclib\Crypt\Common\PublicKey` or `$key instanceof \phpseclib\Crypt\Common\PrivateKey`.

As a consequence of RSA / DSA / ECDSA being immutable you can no longer do stuff like `$key->setHash('sha512')`. Now, you have to do stuff like `$key = $key->withHash('sha512')`. And sure, if all you're doing is setting one parameter, the `withHash` approach is more clunky, but if you're setting multiple parameters, it suddenly becomes a lot more convenient. eg. instead of `$key->setHash('sha512'); $key->setPadding(RSA::SIGNATURE_PKCS1);` now you do `$key = $key->withHash('sha512')->withPadding(RSA::SIGNATURE_PKCS1)`.

One of the big benefits of immutability is that certain types of bugs become non-issues. eg. #1132.

DSA / ECDSA support has also been added to SSH2.php for logging in and to X509.php for verifying and signing certs.

### DIFF
--- a/phpseclib/Crypt/Common/AsymmetricKey.php
+++ b/phpseclib/Crypt/Common/AsymmetricKey.php
@@ -15,11 +15,13 @@
 
 namespace phpseclib\Crypt\Common;
 
+use phpseclib\Exception\UnsupportedFormatException;
+use phpseclib\Exception\NoKeyLoadedException;
 use phpseclib\Math\BigInteger;
 use phpseclib\Crypt\Hash;
-use ParagonIE\ConstantTime\Base64;
-use phpseclib\Exception\UnsupportedOperationException;
-use phpseclib\Exception\FileNotFoundException;
+use phpseclib\Crypt\RSA;
+use phpseclib\Crypt\DSA;
+use phpseclib\Crypt\ECDSA;
 
 /**
  * Base Class for all stream cipher classes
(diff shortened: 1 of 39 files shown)


terrafrost:
I guess I can't respond to the commits in the PR now that it's been merged lol.

Anyway, I've thought about the toString() thing and here's what I'm thinking:

```php
$key = $key->withFormat('PKCS8', [
    'encryptionAlgo' => 'id-PBES2',
    'encryptionScheme' => 'aes128-CBC-PAD',
    'iterationCount' => 1000,
    'PRF' => 'id-hmacWithSHA512',
]);
echo $key->withPassword('password');
```

PR #1382, opened 2019-06-21
fix compat with 5.3

remicollet:
As composer.json still states it is compatible with 5.3.3


terrafrost:
Good catch! It's easy to forget about the limitations of older versions of PHP when you don't work with them very much lol

bantu:
Shouldnt code sniffer find this?

terrafrost:
> Shouldnt code sniffer find this?

It didn't seem to on 5.5, which is the version that Travis CI runs it on:

https://travis-ci.org/phpseclib/phpseclib/jobs/548029710#L850

PR #1397, opened 2019-08-23
Fix SymmetricKey and cipher class constructor mode types, and mapping execution order

### LINKED ISSUE #1396: Inconsistent types in constructor of Rijndael and its superclass
Why does the constructor of `Rijndael` [want's a mode](https://github.com/phpseclib/phpseclib/blob/b89c488f4338d74f0fec05be29c0deae2dc50b97/phpseclib/Crypt/Rijndael.php#L175) in form of an `int` like `Rijndael::MODE_CBC` and checks if `$mode == self::MODE_STREAM`, but the constructor of it's superclass `SymmetricKey` [performs operations](https://github.com/phpseclib/phpseclib/blob/b89c488f4338d74f0fec05be29c0deae2dc50b97/phpseclib/Crypt/Common/SymmetricKey.php#L609) based on a string?

### `phpseclib/Crypt/Rijndael.php`
``` php
    /**
     * Default Constructor.
     *
     * @param int $mode
     * @access public
     * @throws \InvalidArgumentException if an invalid / unsupported mode is provided
     */
    public function __construct($mode)
    {
        if ($mode == self::MODE_STREAM) {
            throw new BadModeException('Block ciphers cannot be ran in stream mode');
        }
        parent::__construct($mode);
    }
```
### `phpseclib/Crypt/Common/SymmetricKey.php`
``` php
    public function __construct($mode)
    {
        $mode = strtolower($mode);
        // necessary because of 5.6 compatibility; we can't do isset(self::MODE_MAP[$mode]) in 5.6
        $map = self::MODE_MAP;
        if (!isset($map[$mode])) {
            throw new BadModeException('No valid mode has been specified');
        }
```

gocom:
Addresses findings brought up in #1396. Ciphers extending the base SymmetricKey class expect integers in their constructor param and checks, but then pass the given value to the inherited constructor that expects string labels. This leads to a case where either the check, or the inherited mapper, can't function.

This PR addresses that by changing the SymmetricKey to accept both strings and the constants themselves, and fixes execution order; the inherited mapper is ran before the validation check so that it will function if a label is given instead of the constant's value.

Fixes #1396

### DIFF
--- a/phpseclib/Crypt/Blowfish.php
+++ b/phpseclib/Crypt/Blowfish.php
@@ -288,17 +288,17 @@ class Blowfish extends BlockCipher
     /**
      * Default Constructor.
      *
-     * @param int $mode
+     * @param int|string $mode
      * @access public
      * @throws \InvalidArgumentException if an invalid / unsupported mode is provided
      */
     public function __construct($mode)
     {
+        parent::__construct($mode);
+
         if ($mode == self::MODE_STREAM) {
             throw new \InvalidArgumentException('Block ciphers cannot be ran in stream mode');
         }
-
-        parent::__construct($mode);
     }
 
     /**
--- a/phpseclib/Crypt/Common/SymmetricKey.php
+++ b/phpseclib/Crypt/Common/SymmetricKey.php
@@ -602,20 +602,22 @@ abstract class SymmetricKey
      *
      * - gcm
      *
-     * @param string $mode
+     * @param string|int $mode
      * @access public
      * @throws BadModeException if an invalid / unsupported mode is provided
      */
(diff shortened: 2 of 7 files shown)


terrafrost:
Nice catch! That said, idk that I'm interested in publicly supporting both strings (eg. `new AES('ctr')`) and numeric constants (`new AES(AES::MODE_CTR)`) tho. I mean, internally, it makes sense - it's faster (albeit a bit of a micro optimization) to compare integers than it is to compare arbitrary length strings, but externally it's a different matter..

bantu:
The key argument against strings is having one mechanism and only one.

PR #1409, opened 2019-09-16
Added test SFTPWrongServerTest

gajdamaka:


terrafrost:
I went ahead and went with #1410 instead of this one. The submitter of that one submitted his post before yours and altho I do like the fact that yours has a unit test (whereas the one I did merge doesn't) the other one had a slightly better fix and he got there first.

Maybe submit a unit test as a separate PR?

Thanks!

PR #1418, opened 2019-10-06
Listing directory as a generator

StefanPrintezis:
I've been trying to find a way to list the directory, without the list having to be entirely in an array, which uses allot of memory.

I've looked into the protocol and the listing is being send incrementally therefor it should be possible. https://tools.ietf.org/html/draft-ietf-secsh-filexfer-13#section-8.1.2

I've created a generator that yields the attributes of every file in a directory.

This could be some logic the Stream::_dir_opendir function could use, this would make more sense as it's supposed to stream the data/listing.

### DIFF
--- a/phpseclib/Net/SFTP.php
+++ b/phpseclib/Net/SFTP.php
@@ -812,6 +812,17 @@ class SFTP extends SSH2
     }
 
     /**
+     * Yields information about files in the given directory.
+     *
+     * @param string $dir
+     * @return \Generator
+     */
+    public function nlistGenerator($dir = '.')
+    {
+        yield from $this->readlistGenerator($dir);
+    }
+
+    /**
      * Returns a detailed list of files in the given directory
      *
      * @param string $dir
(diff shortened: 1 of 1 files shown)


phpseclib/Net/SFTP.php
```diff
@@ -811,6 +811,17 @@ private function nlist_helper($dir, $recursive, $relativeDir)
...
 
+    /**
+     * Yields information about files in the given directory.
+     *
+     * @param string $dir
+     * @return \Generator
+     */
+    public function nlistGenerator($dir = '.')
+    {
+        yield from $this->readlistGenerator($dir);
```
StefanPrintezis: An alternative for 5.5 should not be an issue if compatibility is required.

phpseclib/Net/SFTP.php
```diff
@@ -961,6 +972,108 @@ private function readlist($dir, $raw = true)
...
+                                $this->update_stat_cache($dir . '/' . $shortname, []);
+                            } else {
+                                if ($shortname == '..') {
+                                    $temp = $this->realpath($dir . '/..') . '/.';
+                                } else {
+                                    $temp = $dir . '/' . $shortname;
+                                }
+                                $this->update_stat_cache($temp, (object) ['lstat' => $attributes]);
+                            }
+                            yield $content;
```
StefanPrintezis: Instead of filling the array yield the value for processing and pause the function until next() is called.

phpseclib/Net/SFTP.php
terrafrost: To be included in the 3.0 branch 5.6 compatibility is a requirement but the master branch should always be assumed to have an unstable API and is not at all aiming to maintain BC with itself. Overall, I'm okay bumping the master branch version reqs to 7.0.

I'll try to review your PR more thoroughly this evening - thanks!

phpseclib/Net/SFTP.php
StefanPrintezis: I think this function needs to be more DRY, but that would probably require some refactoring of existing code. My trail of thought was to introduce a new feature, but maybe a generator for these kind of requests should be default for all functions. Implementation thereof could be of any kind. Filling an array with the attributes, which costs allot of memory would merely be a helper, but not advised, since the default would be a generator/Iterable.

terrafrost:
First thoughts:

It looks like this is more analagous to rawlist() as opposed to nlist().

There's a huge amount of code re-use between readlist() and readlistGenerator(). The main differences are...  instead of doing `$contents[$shortname] = ...` you're doing `$content = .... You're also doing `yield $content` after the stat cache is updated and you're removng the `count($this->sortOptions) bit. Outside of that the code is 100% the same. 

I suppose in the pursuit of lowered cyclomatic complexity having two different methods might be the better approach but from an "eliminating duplicate code" I feel like it'd be better if there was one method with a few new if conditions than two nearly 100% identical methods.

I feel like it might also be better if nlist() were updated to return an [ArrayObject](https://www.php.net/manual/en/class.arrayobject.php) object instead of an array. I feel like that would eliminate the need for using yield. Another advantage of this approach vs yield is that you could see how many "files" were in the directory, you could access the n'th file, etc, altho in reviewing the SFTP specs, even to count all the "files" would still require you iterate through all of them. `__debugInfo` could also be utilized to make this rewritten `print_r($ssh->nlist())` work as it currently does (mostly) as well.

Should this hypothetical ArrayObject thing be an explicitly defined class or an anonymous class? atm idk.

One thing I've also been mulling over for SFTP is removing the stat cache but that's another matter entirely I suppose lol.

StefanPrintezis:
I'm not entirely sure how using an ArrayObject would eliminate the need for using yield.

Yielding in a function returns a [Generator](https://www.php.net/manual/en/class.generator.php) object, which is what I'm aiming for. We could use the Generator to fill the array as it is filled now to reduce dupe code.

I wouldn't want the result to be countable, since the Generator object does not implement Countable.

Lowering the cyclomatic complexity is going to be a challenge since i'm not very familiar with this library. But since it's a mayor release, breaking (backwards compatibility) I'll give it a try.

terrafrost:
> I'm not entirely sure how using an ArrayObject would eliminate the need for using yield.

Here's a demo script that does just that:

https://pastebin.com/Vftyfn4D

It creates an [iterable](https://www.php.net/is-iterable) using three different techniques.

- with [yield](https://www.php.net/manual/en/language.generators.syntax.php)
- with [range()](https://www.php.net/range)
- with [IteratorAggregate](https://www.php.net/manual/en/class.iteratoraggregate.php) (which is one of the four interfaces that [ArrayObject](https://www.php.net/manual/en/class.arrayobject.php) implements)

[memory_get_usage](http://php.net/memory-get-usage) is used to observe the memory each technique requires. You can see the results here:

If you [run the script](https://3v4l.org/JQsSv) you'll get the following results (in PHP 7.4):

- yield uses 512 bytes
- range() uses 33,952,128 bytes
- IteratorAggregate uses 376 bytes.

The numbers vary slightly depending on the PHP version you're using but one thing is consistent: yield and IteratorAggregate are always neck to neck with one another. It wasn't until PHP 7.4 that IteratorAggregate actually became more memory efficient than yield but even on PHP 7.3 and earlier ArrayObject provides a ton more versatility than yield does. ArrayObject enables random access (via ArrayAccess), enables `count` (via Countable) and allows for custom serialization.

You can tell from the memory utilization that the entire "range" isn't being loaded all at once with yield or IteratorAggregate, which is the property you're looking for.

rmikalkenas:
Hi! What's the final output of this MR? I'm facing similar issue with `\phpseclib3\Net\SFTP::rawlist` method, where I have an sftp with a directory containing tens of thousand of files and it's causing a drastically increased memory usage.
@StefanPrintezis maybe you found a way to workaround it somehow?

terrafrost:
I haven't done any work on this, sadly.

The closest I've gotten to doing work on this was more for opening files - not for directories:

https://github.com/terrafrost/phpseclib/commit/17a1a4100cc6dea9939d85a4daab20da3037d5e2

I'm not sure when I'll be able to get to this...

StefanPrintezis:
> Hi! What's the final output of this MR? I'm facing similar issue with `\phpseclib3\Net\SFTP::rawlist` method, where I have an sftp with a directory containing tens of thousand of files and it's causing a drastically increased memory usage. @StefanPrintezis maybe you found a way to workaround it somehow?

Hey, this PR was my way to work around it. The protocol already sends the data in a batches, so instead of gathering it in an array before returning(which causes memory to go nuts, because it's an array in php), i've used a generator function to return "yield" the content of each response.

mnapoli:
For the record: https://github.com/phpseclib/phpseclib/pull/2122

Issue #1423, opened 2019-10-19
RSA PSS doesn't support non power of 2 public keys


masterzen:
The RFC3447 doesn't mention that public keys modulus size should be a power of 2.
Though, the RSA PSS verification code only supports keys whose modulus has a size that is a power of 2.

I discovered this while trying to validate the JWS PS256 signature of a token that had been signed with a public key of 522 bits.

terrafrost:
Your fix has been merged.

Thanks!

PR #1424, opened 2019-10-19
(#1423) make PSS verification work for non power of 2 keys

masterzen:
RFC3447 doesn't require RSA public keys to be power of 2 keys.
The actual validation code doesn't work when trying to verify
a PSS signature generated with a non power of two key.
This small patch adds support for such keys.

### DIFF
--- a/phpseclib/Crypt/RSA/PublicKey.php
+++ b/phpseclib/Crypt/RSA/PublicKey.php
@@ -212,7 +212,7 @@ class PublicKey extends RSA implements Common\PublicKey
         // if $m is larger than two million terrabytes and you're using sha1, PKCS#1 suggests a "Label too long" error
         // be output.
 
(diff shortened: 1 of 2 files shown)


terrafrost:
Looks good! Altho I see no reason why this change can't be in the 1.0 / 2.0 / 3.0 branches lol. I think I'll cherry pick this into the 1.0 branch (which _will_ generate conflicts) but then I'll adapt it it to the 1.0 branch (this way you'll get credit for the change) and I'll then merge it into the 2.0 / 3.0 / master branches, making the appropriate adjustments along the way.

terrafrost:
I've gone ahead and done as I described:

https://github.com/phpseclib/phpseclib/commit/207cd0f4ea0ae9b8065975925848184a39c3cf67

Thanks!

Issue #1427, opened 2019-11-04
Laravel - No compatible key exchange algorithms found


jonlee90:
Hi,

I'm trying to pull a file from a different server using SFTP but I keep getting "No compatible key exchange algorithms found" errorExeption whenever I try to connect. I'm currently using Laravel 5. 8 in a CentOS Redhat server.

Thank you.

terrafrost:
Can you post your SSH logs? You can get them by doing `define('NET_SSH2_LOGGING', true);` at the top and `$ssh->getLog();` after the failed login attempt.

In lieu of having seen them my best guess is that the SSH server in question just uses more modern algorithms and that the 3.0 branch would solve the issue for you. I'm in the process of documenting how the 3.0 branch works and will then release 3.0.0 but, in the mean time, it should be stable enough to use. It passes all the unit tests!

jonlee90:
Hi Terrafrost,

I was able to call the getLog right before the kex_algorithm check and I got this as a response:
array:4 [
  0 => "<-"
  1 => "->"
  2 => "-> NET_SSH2_MSG_KEXINIT (since last: 0.1728, network: 0s)"
  3 => "<- NET_SSH2_MSG_KEXINIT (since last: 0.0001, network: 0s)"
]

I also tried to install 3.0 branch but I keep getting this error:
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - league/flysystem-sftp 1.0.6 requires phpseclib/phpseclib ~0.3.8 -> satisfiable by phpseclib/phpseclib[0.3.10, 0.3.8, 0.3.9] but these conflict with your requirements or minimum-stability.
    - league/flysystem-sftp 1.0.5 requires phpseclib/phpseclib ~0.3.8 -> satisfiable by phpseclib/phpseclib[0.3.10, 0.3.8, 0.3.9] but these conflict with your requirements or minimum-stability.
    - league/flysystem-sftp 1.0.4 requires phpseclib/phpseclib ~0.3.8 -> satisfiable by phpseclib/phpseclib[0.3.10, 0.3.8, 0.3.9] but these conflict with your requirements or minimum-stability.
    - league/flysystem-sftp 1.0.3 requires phpseclib/phpseclib ~0.3.8 -> satisfiable by phpseclib/phpseclib[0.3.10, 0.3.8, 0.3.9] but these conflict with your requirements or minimum-stability.
[...]

PR #1431, opened 2019-11-13
rename phpseclib/phpseclib to phpseclib/phpseclib3 for 3.0 branch

terrafrost:
The motivation for this change is so that phpseclib 2.0 (or at least a shim of 2.0) and 3.0 can co-exist so that problems like #1427 can be avoided. In order to be an effective shim the shim has to have the namespace as phpseclib 2.0 (eg. `\phpseclib\Crypt\RSA`) which means that if the shim is going to be using phpseclib 3.0 that phpseclib 3.0 can't make use of that same namespace as phpseclib 2.0 because otherwise the PHP will think files are already loaded when they're in fact not (and if it did try to force load them you'd get a "class already exists" error)

Since phpseclib 3.0 is a huge BC breaking change, anyway, one more very minor BC breaking change isn't a huge deal imho

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -72,7 +72,7 @@
             "phpseclib/bootstrap.php"
         ],
         "psr-4": {
-            "phpseclib\\": "phpseclib/"
+            "phpseclib3\\": "phpseclib/"
         }
     }
 }
--- a/phpseclib/Common/Functions/Strings.php
+++ b/phpseclib/Common/Functions/Strings.php
@@ -13,10 +13,10 @@
  * @link      http://phpseclib.sourceforge.net
  */
 
-namespace phpseclib\Common\Functions;
+namespace phpseclib3\Common\Functions;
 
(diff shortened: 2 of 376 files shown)


composer.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "phpseclib/phpseclib",
+    "name": "phpseclib/phpseclib3",
```
BinaryKitten: should the package name be updated?  I'm for the rest, but this is beyond a code change and has deeper effects. 
If the package name is to change, then maybe phpseclib3 needs to migrate to a new repo?

terrafrost: Oh - good catch. I did a `sed` on the codebase and I guess that got caught up in it lol.

I'm not currently in a place where I can change it (I'm traveling and saw the Pyramids today!) but I'll update this as soon as I'm able.

Thanks!

PR #1437, opened 2019-12-08
rename phpseclib/phpseclib to phpseclib/phpseclib3 for 3.0 branch

terrafrost:
#1431 targeted the master branch - not the 3.0 branch.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -72,7 +72,7 @@
             "phpseclib/bootstrap.php"
         ],
         "psr-4": {
-            "phpseclib\\": "phpseclib/"
+            "phpseclib3\\": "phpseclib/"
         }
     }
 }
--- a/phpseclib/Common/Functions/Strings.php
+++ b/phpseclib/Common/Functions/Strings.php
@@ -13,10 +13,10 @@
  * @link      http://phpseclib.sourceforge.net
  */
 
-namespace phpseclib\Common\Functions;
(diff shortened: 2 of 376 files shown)


GrahamCampbell:
Won't this make a v4 a pain?

terrafrost:
> Won't this make a v4 a pain?

v4 will just have a new `\phpseclib4` namespace.

phpseclib v3 had huge BC breaking changes. I fully expect phpseclib v4 will as well.

This was done to facilitate https://github.com/phpseclib/phpseclib2_compat . Composer doesn't let you check out multiple versions of the same package but if you have one package using v2 of phpseclib and another using v3 then they both can work with the shim.

The fact that v2 and v3 have slightly different namespaces also means that one could, in theory, fork phpseclib and use the fork for v2 and the main version for v3 but the shim offers additional improvements that this forking approach wouldn't provide: namely, it lets you use all the new algorithms introduced in v3 in v2. Like you can't use an Ed25519 private key with v2 but you should be able to, without issue, with the shim (I need to add unit tests for these use cases; the unit tests right now are just to ensure BC with v2).

GrahamCampbell:
Right. Something guzzle did was make another git repo and package, so Guzzle 3 existed, and then Guzzle 4 could be installed as well as it, since it was another repo and package. This is probably overkill though, and not what we want if there are going to be more major changes in v4 too.

PR #1444, opened 2020-01-10
Remove error suppression from stream_select()

tomsommer:
Suppressing errors from stream_select() makes debugging timeouts extremely hard.

### DIFF
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -1245,7 +1245,7 @@ class SSH2
                     $usec = 1000000 * ($this->curTimeout - $sec);
                     // on windows this returns a "Warning: Invalid CRT parameters detected" error
                     // the !count() is done as a workaround for <https://bugs.php.net/42682>
-                    if (!@stream_select($read, $write, $except, $sec, $usec) && !count($read)) {
+                    if (stream_select($read, $write, $except, $sec, $usec) === false && !count($read)) {
                         $this->is_timeout = true;
                         return false;
                     }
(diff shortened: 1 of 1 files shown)


terrafrost:
According to https://bugs.php.net/42682 the issue that motivated me to add error suppression has been fixed. I'll try to confirm when I get home and if it is indeed no longer an issue I'll merge it.

Thanks!

tomsommer:
I think you are referring to the count() workaround in the same line :)

terrafrost:
So it looks like error suppression (which has been in place since the timeout code was added on [Jun 23, 2012](https://github.com/phpseclib/phpseclib/commit/9e69344d60e911b67e5f7f2b38e10e12c55e4bb6)) actually preceded the ` && !count($read)` bit (which has been in place since [Aug 29, 2012](https://github.com/phpseclib/phpseclib/commit/5de5d45bd5163b1c974eaab0eae410fdd643645c#diff-82a5983a6daace4acbcc65edf2dfb605)). The https://bugs.php.net/42682 link apparently applies to the ` && !count($read)` bit specifically.

That said, I'm not able to reproduce either issue on Windows. I tested it as low as PHP 5.0.5. It looks like a lot of people had issues reproducing the bug even in the bug report. As such I'm not even sure unit tests against https://www.appveyor.com/ would have helped.

Anyway, can you resubmit this PR but target the 3.0 branch instead? They're pretty much identical so it shouldn't be a ton of effort on your part (whereas on my part I'd have to add you as a remote then I'd have to cherry-pick your commit into the 3.0 branch).

Thanks!

terrafrost:
Oh - and while you're at it, might as well remove the ` && !count($read)` bit as well. Just for the 3.0 branch. If that causes the unit tests to fail (which I don't expect) then readd it.

Thanks!

tomsommer:
Done

PR #1452, opened 2020-02-14
read larger chunks from sftp in stream wrapper

icewind1991:
This speeds up reading from the stream wrapper about 5x in my local testing, increasing the chunk size further doesn't seem to make a difference

### DIFF
--- a/phpseclib/Net/SFTP/Stream.php
+++ b/phpseclib/Net/SFTP/Stream.php
@@ -17,6 +17,7 @@
 
 namespace phpseclib3\Net\SFTP;
 
+use phpseclib3\Common\Functions\Strings;
 use phpseclib3\Crypt\RSA;
 use phpseclib3\Net\SFTP;
 use phpseclib3\Net\SSH2;
@@ -339,8 +340,7 @@ class Stream
             $this->buffer .= $chunk;
         }
 
-        $result = substr($this->buffer, 0, $count);
-        $this->buffer = substr($this->buffer, $count);
+        $result = Strings::shift($this->buffer, $count);
 
         if (empty($result)) { // ie. false or empty string
             $this->eof = true;
@@ -360,6 +360,9 @@ class Stream
      */
     private function _stream_write($data)
     {
+        // invalidate any readahead buffer on write
+        $this->buffer = '';
+
         switch ($this->mode) {
             case 'r':
                 return false;
@@ -425,6 +428,9 @@ class Stream
      */
     private function _stream_seek($offset, $whence)
     {
+        // invalidate any readahead buffer on seek
+        $this->buffer = '';
+
         switch ($whence) {
             case SEEK_SET:
                 if ($offset >= $this->size || $offset < 0) {
@@ -738,6 +744,9 @@ class Stream
      */
     private function _stream_truncate($new_size)
     {
+        // invalidate any readahead buffer on truncate
+        $this->buffer = '';
+
         if (!$this->sftp->truncate($this->path, $new_size)) {
             return false;
         }
--- a/phpseclib/Net/SFTP/Stream.php
+++ b/phpseclib/Net/SFTP/Stream.php
@@ -114,6 +114,14 @@ class Stream
     private $notification;
 
     /**
+     * Buffer of data being read
+     *
+     * @var string
+     * @access public
+     */
+    private $buffer = '';
+
+    /**
      * Registers this class as a URL wrapper.
      *
      * @param string $protocol The wrapper name to be registered.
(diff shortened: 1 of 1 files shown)


bantu:
@icewind1991 Hi Robin. I've not been active in this project for quite some time, so please bear with me.

My understanding is that, essentially, you're introducing a buffer between (external) SFTP requests and (internal) PHP stream requests, correct? That seems to make sense to me.

phpseclib/Net/SFTP/Stream.php
```diff
@@ -315,16 +323,25 @@ private function _stream_read($count)
...
 
-        $result = $this->sftp->get($this->path, false, $this->pos, $count);
-        if (isset($this->notification) && is_callable($this->notification)) {
-            if ($result === false) {
-                call_user_func($this->notification, STREAM_NOTIFY_FAILURE, STREAM_NOTIFY_SEVERITY_ERR, $this->sftp->getLastSFTPError(), NET_SFTP_OPEN, 0, 0);
-                return 0;
+        if (strlen($this->buffer) < $count) {
+            // read 80k chunks from sftp even though php will only request 8k chunk
+            // this allows us to minimize the overhead of making sftp requests
+            $chunk = $this->sftp->get($this->path, false, $this->pos, 81920);
```
bantu: Can we make the 81920 a parameter? I think there is a mechanism for that on stream construction.

terrafrost:
For the most common use case this does indeed appear to speed things up a good amount. A much less common use case, however, would be slowed down by this. eg. randomly seeking / reading 1kb chunks of a file.

That said, I noticed you targeted the master branch with this PR. Maybe 3.0 might not be a bad idea? I say that because long term I do want to do something like https://github.com/phpseclib/phpseclib/pull/655#issuecomment-92600348 , which could very well eliminate any speedups your PR provides.

phpseclib/Net/SFTP/Stream.php
```diff
@@ -315,16 +323,25 @@ private function _stream_read($count)
...
+                call_user_func($this->notification, STREAM_NOTIFY_PROGRESS, STREAM_NOTIFY_SEVERITY_INFO, '', 0, strlen($chunk), $this->size);
             }
-            // seems that PHP calls stream_read in 8k chunks
-            call_user_func($this->notification, STREAM_NOTIFY_PROGRESS, STREAM_NOTIFY_SEVERITY_INFO, '', 0, strlen($result), $this->size);
+
+            $this->buffer .= $chunk;
         }
 
+        $result = substr($this->buffer, 0, $count);
+        $this->buffer = substr($this->buffer, $count);
```
terrafrost: This is nitpicky but...  you could do `$result = \phpseclib3\Common\Functions\Strings::shift($this->buffer, $count)` instead. Also, I'd do `$this->buffer.= $chunk` instead of `$this->buffer .= $chunk`.

phpseclib/Net/SFTP/Stream.php
terrafrost: I think what bantu is talking about is this:

```
<?php
require __DIR__ . '/vendor/autoload.php';

use phpseclib3\Net\SFTP\Stream;
use phpseclib3\Net\SFTP;

Stream::register();

$sftp = new SFTP('website.com');
$sftp->login('username', 'password');

$context = [
    'sftp' => ['sftp' => $sftp]
];
$context = stream_context_create($context);

$fp = fopen('sftp://dummy.com/home/vagrant/1mb', 'r', false, $context);
echo fread($fp, 10);
```
Other options are discussed at https://www.php.net/manual/en/wrappers.ssh2.php

bantu: Yes, "stream context" is what I was referring to.

bantu:
What if
* multiple chunks are loaded into the buffer
* the first one is read from the stream, modified, and written back to the underlying file of the stream at chunk 2

Does the stream still behave as expected?

terrafrost:
@bantu - good points. Maybe it'd just be easier if I implemented the ideas discussed in https://github.com/phpseclib/phpseclib/pull/655#issuecomment-92600348 . I could do that in the 1.0 / 2.0 / 3.0 / master branches. That's not a BC breaking change, after all.

bantu:
@terrafrost Maybe call the parameter something like "readahead length" and make it optional. This way Robin can use it for his use case. If the parameter is not supplied, it should behave as before.

@icewind1991 What do you think?

icewind1991:
made the readahead length configurable and added cache invalidation logic to prevent reading stale chunks.

bantu:
> added cache invalidation logic to prevent reading stale chunks

In the buffering problem I described in https://github.com/phpseclib/phpseclib/pull/1452#issuecomment-590016219, I was (implicitly) considering two independent instances accessing the file. Your additional changes do not work in this case. Hence my suggestion for not changing the behavior in case the read ahead buffer length is not supplied.

phpseclib/Net/SFTP/Stream.php
```diff
@@ -113,6 +114,16 @@ class Stream
...
 
+    private $readAheadLength = 81920;
+
+    /**
+     * Buffer of data being read
+     *
+     * @var string
+     * @access public
+     */
+    private $buffer = '';
```
amayer5125: Should this say `@access private`?

terrafrost:
Here's my idea for an alternate way to deal with this:

https://github.com/terrafrost/phpseclib/tree/1.0-sftp-file-handling

I'm thinking that this - and making `nlist` and `rawlist` return an ArrayObject instead of an actual array (see #1418) - could create new versions of phpseclib - 1.1.0, 2.1.0 and 3.1.0.

What I've done can be used as a starting point. If no one else wants to take this task on I can do so in the fullness of time.

PR #1466, opened 2020-04-02
Fix logic error reading random bytes from /dev/urandom

TysonAndre:
Fixes a bug introduced in c2be7e648

Previously, this would return those bytes if the number of bytes read
was **less than** the number of bytes this was trying to read.

In practice, I believe this would mean bytes from /dev/urandom would never
get used.  (Noticed when upgrading phpseclib)

Also, I think that `if (!\is_bool($fp))` might be even faster than `if ($fp !== true && $fp !== false) {` or `if (!is_bool($fp))` - PHP Opcache can convert the is_bool call to an efficient ZEND_TYPE_CHECK opcode instead of a function call, but only if opcache is enabled and the call is fully qualified (e.g. with `\`).

This looks like it only impacts php 5 applications without openssl support

### DIFF
--- a/phpseclib/Crypt/Random.php
+++ b/phpseclib/Crypt/Random.php
@@ -98,7 +98,7 @@ class Random
             }
             if ($fp !== true && $fp !== false) { // surprisingly faster than !is_bool() or is_resource()
                 $temp = fread($fp, $length);
-                if (strlen($temp) != $length) {
+                if (strlen($temp) == $length) {
                     return $temp;
                 }
             }


terrafrost:
I cherry picked this to the 1.0 branch and then merged that to 2.0 / 3.0 / master (well, 3.0 / master didn't really see any changes but merging into 3.0 / master would have resulted in a merge conflict so I did the merge and resolved the conflict by telling it to use the 3.0 file, unmodified).

Thanks!

Issue #1471, opened 2020-04-21
Methods have different return types than their PHPDoc suggest


func0der:
This line causes a notice (`Trying to access array offset on value of type bool`) and is not developed defensively enough:
https://github.com/phpseclib/phpseclib/blob/34620af4df7d1988d8f0d7e91f6c8a3bf931d8dc/phpseclib/Net/SSH2.php#L3471

The switch case above that line provides multiple cases where there is not 'return' to abort method execution.
In those switch cases, there are call so `\phpseclib\Net\SSH2::_get_binary_packet()` which internally also calls `\phpseclib\Net\SSH2::_filter()`. Both of these methods falsely advertise that their return type is a `string` when they multiple instances boolean returns in their method body.

The real problem therefore is burried in the wrong return types of the two methods mentioned above.

I see that `\phpseclib\Net\SSH2::_filter()` is only called from `\phpseclib\Net\SSH2::_get_binary_packet()` so changing `_filter()` should be an easy fix.
`\phpseclib\Net\SSH2::_get_binary_packet()` on the other hand relies on returning false.
I'll try to address that in a PR, but since I am not very familier with SFTP, I am not sure, I will get this right without your help.

terrafrost:
> Both of these methods falsely advertise that their return type is a `string` when they multiple instances boolean returns in their method body.

The DocBlock comments do indeed need to be improved. Unfortunately, that's not really my highest priority. Right now I'm working on the documentation for v3.0. If someone wants to do a PR for the return types they're very welcome to do so.

As I noted in your PR you're proposed fix is problematic because throwing exceptions in the 2.0 branch would be a PR break. So here's what I'll do.

In the 1.0 / 2.0 branches I'll add a line to return false if `$payload` is false, right above the line you included in your post.

In the 3.0 / master branches I'll look for more `return false`'s that can be replaced with exceptions. Like I said in your PR the 3.0 / master branches already do throw exceptions so throwing a few more ain't no thing.

terrafrost:
Looking at the code (1.0) it is not clear to me how `_filter()` could get anything other than a string for `$payload`. As you note, `_filter()` is only called from `_get_binary_packet()`.

`$payload` is set thusly:

```php
$payload = $this->_string_shift($raw, $packet_length - $padding_length - 1);
```
Here's the definition of `_string_shift`:

```php
    /**
     * String Shift
     *
     * Inspired by array_shift
     *
     * @param string $string
     * @param int $index
     * @return string
[...]

PR #1472, opened 2020-04-21
[fix_SSH2_class] Introduce a fix for inconsistent return type problem."

func0der:
Proposed fix for #1471

### DIFF
--- a/phpseclib/Exception/PhpSecLibException.php
+++ b/phpseclib/Exception/PhpSecLibException.php
@@ -1,0 +1,9 @@
+<?php
+
+namespace phpseclib\Exception;
+
+use Exception;
+
+class PhpSecLibException extends Exception implements PhpSecLibExceptionInterface
+{
+}
--- a/phpseclib/Exception/PhpSecLibExceptionInterface.php
+++ b/phpseclib/Exception/PhpSecLibExceptionInterface.php
@@ -1,0 +1,7 @@
+<?php
+
+namespace phpseclib\Exception;
+
+interface PhpSecLibExceptionInterface
+{
+}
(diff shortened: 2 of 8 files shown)


terrafrost:
Throwing exceptions in the 2.0 branch is a non starter as it's a BC break. The 3.0 branch, however, does throw exceptions.

As for the issue that this is attempting to fix...  I'll take a look in the next few days as time permits!

Thanks!

func0der:
> Throwing exceptions in the 2.0 branch is a non starter as it's a BC break.

The only changes made are those to "private" methods (those with an _ (underscore) in front of it). If someone was to use those methods, they would use the library wrong.
Also the `@access` PHPDoc states that those methods are private.

It would be possible to have the internal code throw exceptions and the external api to keep returning the falsely documented results like "false" etc.

terrafrost:
> It would be possible to have the internal code throw exceptions and the external api to keep returning the falsely documented results like "false" etc.

The public functions would need to catch the exceptions thrown by the private exceptions and return false, if appropriate, instead. At least for the code changes not to be BC breaking.

Like consider this example:

```php
function a()
{
    throw new \Exception('demo');
}

function b()
{
    a();
}

b();
```
The exception `a()` throws will kinda float up to `b()`.

And make no mistake the changes are BC breaking. Exceptions do offer a lot of advantages over doing `user_error()` but `user_error()` won't kill a program whereas an uncaught exception will.

Also, I'd do one or the other (`user_error()` or throwing an exception) - not both.

Finally, I'm not the biggest fan of including the full SSH packet in the exceptions. The full SSH packet is going to be binary encoded and ran over a command line it could mess up how stuff displays. As a simple demonstration...

```php
for ($i = 128; $i <= 255; $i++) {
    echo chr($i);
}
```
Does that actually output anything for you? It doesn't output anything for me.

PR #1503, opened 2020-07-30
_parseKey should return false by default

VincentLanglet:
If you try to use `setKey` with a non supported type, you handle with an exception.
Indeed the method `_parseKey` will return `null` but the check made after is `=== false`.

This fix, should be made in 2.0 too. I assume you'll merge 1.0 into 2.0.


terrafrost:
Merged to 2.0!

Thanks!

PR #1510, opened 2020-08-14
Resource not closed

sykorax:
Hey there,

We might have been identified a problem with the SFTP (2.x) version only. This file has been used by League Flysystem SFTP adapter itensively. We found this in the League package itself as well, however the problem could be solved by changing just a few lines of code - see my pull request.

Since we're having a strong dependency on the adapter pattern - provided by the League package - we cannot get rid of it. This has led us to change these lines.

About the problem: We faced with an issue, when transferring files from a local <-> SFTP destination (and vice-versa) and some of the read streams are just being stuck / not closed

### DIFF
--- a/phpseclib/Net/SFTP.php
+++ b/phpseclib/Net/SFTP.php
@@ -2084,14 +2084,16 @@ class SFTP extends SSH2
             return false;
         }
 
-        if ($mode & self::SOURCE_LOCAL_FILE) {
-            if ($this->preserveTime) {
-                $stat = fstat($fp);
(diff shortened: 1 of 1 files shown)


terrafrost:
I went ahead and did this here:

https://github.com/phpseclib/phpseclib/commit/f4e017f383074b6bbf99b256a229b0bc74480a6a

I opted not to use yours because it wasn't targeting 1.0, among other reasons. I could have squashed and cherry picked but it was just too much effort idk lol.

Thanks!

PR #1543, opened 2020-10-21
ProgressCallback should report the actual downloaded amount of data and not the requested packets.

thirsch:
$progressCallback should be called after a specific amount of data has been received and not, if its just requested, as the requested packets might be more as the available and received data.

This does not only solve the problem for $length = -1, because it requests 32 "slots" of max_packet_size packets nevertheless if the requested file is that big or not, it also solves the issue that requesting the data is not having the data delivered. Placing the call to $progressCallback later in the script, it reports the actual received amount of data.

### DIFF
--- a/phpseclib/Net/SFTP.php
+++ b/phpseclib/Net/SFTP.php
@@ -2248,9 +2248,6 @@ class SFTP extends SSH2
                 }
                 $packet = null;
                 $read+= $packet_size;
-                if (is_callable($progressCallback)) {
-                    call_user_func($progressCallback, $read);
-                }
                 $i++;
             }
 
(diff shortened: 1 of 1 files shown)


terrafrost:
I'll try to take a look at this in the next few days - I've been without internet in Phantom Ranch (bottom of the Grand Canyon) and have one more days of hiking to do before I return to my hometown!

thirsch:
You should definitely use your remaining time at the Grand Canyon instead of reviewing this pr :) But happy to get feedback, after you are back to business.

terrafrost:
I cherry picked this into the 1.0 branch and merged into 2.0 / 3.0 / master:

https://github.com/phpseclib/phpseclib/commit/b2ae60f0a91adac5eedd190cbfe4bd6b3df327e5

Thanks!

thirsch:
Thanks, are there any plans for the next release?

PR #1549, opened 2020-12-08
fix docblock : decrypt is not internal

fezfez:


terrafrost:
The `@internal` annotation doesn't mean that the method in question is for "internal" use only. Quoting https://manual.phpdoc.org/HTMLSmartyConverter/HandS/phpDocumentor/tutorial_tags.internal.pkg.html, "_Use @internal to facilitate the creation of two sets of documentation, one for advanced developers or for internal company use, and the other for the general PHP public_".

Here's maybe a better example of `@internal` in use.

```
    /**
     * Sort Array
     *
     * @access public
     * @param string $iv
     * @internal Implements https://en.wikipedia.org/wiki/Quicksort
     */
```
An end user doesn't need to know the sorting algorithm that's being used but it might be relevant for a developer trying to further develop the library.

In the specific line your PR is modifying...  it's more for people who would try to implement their own crypto algorithms into the phpseclib framework by extending the Base object - it's not intended for the average end-user.

fezfez:
@terrafrost you refer a old version of phpdoc, see : https://docs.phpdoc.org/latest/guide/references/phpdoc/tags/internal.html

terrafrost:
I'd say the use still fits. Quoting that page (emphasis mine):

> The `@internal` tag is used to denote that associated _Structural Elements_ are internal to the application or library. It may also be used inside a long description to insert a piece of text that is **only applicable for the developers of the software.**

I'd say that applies to this use of the `@internal` tag.

fezfez:
Doc say :  @internal is about mark method as internal of project and {@internal} Include a note in the Description that would only show in Developer Docs

Thoses tag have different signification

terrafrost:
I'm just not seeing that in the link you provided. The closest thing to that that I'm seeing is this:

> Authors MAY use this tag to indicate that an element with public visibility should be regarded as exempt from the API - for example: _Library authors MAY regard breaking changes to internal elements as being exempt from semantic versioning._ Static analysis tools MAY indicate the use of internal elements from another library/package with a warning or notice.

In the 1.0 / 2.0 branches, I use `@access private` to denote that, even if the method is implicitly public. But more than that, pursuant to their use of the word "may", it's not a requirement.

I'd be more willing to move it over to `{@internal}` then to remove the note all together. But that said, there are other instances of `@internal` being used. If we're going to remove one then why not remove them all? `grep -r "@internal" . | wc -l` suggests that there are 54 instances of it in the 1.0 branch. Sure, some of them may be in methods that aren't intended to be publicly accessible, but all of them?

fezfez:
> I'm just not seeing that in the link you provided. The closest thing to that that I'm seeing is this:

See example section : 

image

If think its indeed better to check all @internal that's is not @internal but {@internal}, my PR does not intend to fix all problems.

fezfez:
also if you use an @internal method, phpstan and psalm will report a bug/problem.

terrafrost:
Based on https://docs.phpdoc.org/latest/guide/references/phpdoc/tags/internal.html I think phpstan and psalm are in the wrong. I'd quote that URL but I already did in an earlier post. I am not seeing anywhere in that URL it says that `{@internal}` and `@internal` behave differently. That the examples don't show `@internal` with comments following it doesn't mean that they can't be used in that capacity. If `@internal` (as opposed to `{@internal}`) is supposed to be a flag and only a flag then why do the docs for it, at the top, say `@internal [description]` in the Syntax section?

> my PR does not intend to fix all problems

1. Well your PR also "fixes" this issue not by changing it to `{@internal}` but by deleting it, which I've already commented on
2. I think consistency is virtuous. phpseclib 3.0 / master use `[]` for brackets, everywhere, and not `array()`. phpseclib uses spaces, per PSR1/2, and not tabs. I don't mix and match standards. It's all or none.

Anyway, I don't think we're going to see eye to eye here. From my vantage point this is just making a change for the sake of making a change. In light of this, I've just gone straight to the horse's mouth and posted https://github.com/phpDocumentor/phpDocumentor/issues/2736

Thanks!

terrafrost:
Well the phpDocumentor issue I posted wasn't super helpful so I did some tests.

## Without any @internal

```php
<?php

/**
 * Sorts Array
 *
 * @access public
 * @param string $iv
 */
function mySort()
{
}
```



## With @internal

```
<?php

/**
 * Sorts Array
 *
 * @access public
 * @param string $iv
 * @internal Implements https://en.wikipedia.org/wiki/Quicksort
 */
function mySort()
{
}
```



## With {@internal}

```
<?php

/**
 * Sorts Array.
 * {@internal Implements https://en.wikipedia.org/wiki/Quicksort}
 *
 * @access public
 * @param string $iv
 */
function mySort()
{
}
```



It also looks like `{@internal}` needs to be on it's own line. If you do `Sorts Array. {@internal Implements https://en.wikipedia.org/wiki/Quicksort}` it'll show up like this:



imho this behavior is not well documented in the phpDocumentor docs but whatever. I'm still not going to merge this PR in. The proper solution isn't to just delete `@internal` but to replace it with `{@internal}`. And I also think it ought to be done all in one go. I'll do it myself in the fullness of time.

I appreciate your persistence / patience in the matter!

terrafrost:
https://github.com/phpseclib/phpseclib/commit/a19b5b4ca85523317a72556ed1c2a10ccd99dc8f should fix this. I just did it for 3.0 - not for 2.0. If someone wants to do it for 2.0 - in it's entirety - they're free to do so. Or maybe I'll do it at some point idk lol

PR #1556, opened 2020-12-21
Fix docblock

lbausch:
While bundling phpseclib into a phar for a third-party app, an error was thrown because of the invalid doc block. This PR fixes the docblock and removes an unneeded (?) comment.

### DIFF
--- a/phpseclib/System/SSH/Agent/Identity.php
+++ b/phpseclib/System/SSH/Agent/Identity.php
@@ -50,7 +50,6 @@ class Identity implements PrivateKey
      */
     const SSH_AGENT_RSA2_256 = 2;
     const SSH_AGENT_RSA2_512 = 4;
-    /**#@-*/
 
     /**
      * Key Object
--- a/phpseclib/System/SSH/Agent/Identity.php
+++ b/phpseclib/System/SSH/Agent/Identity.php
@@ -41,7 +41,7 @@ class Identity implements PrivateKey
 {
     use \phpseclib3\System\SSH\Common\Traits\ReadBytes;
 
-    /**@+
+    /**
      * Signature Flags
      *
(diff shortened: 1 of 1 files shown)


terrafrost:
They're called [dockblock templates](https://devdocs.magento.com/guides/v2.4/coding-standards/docblock-standard-general.html#docblock-templates) but I guess phpDoc no longer supports them.

There are a lot more instances of them then the one you removed. I'll go ahead and clean it up.

lorenzbausch:
TIL and thank you!

terrafrost:
I removed them from the 3.0 and master branches:

https://github.com/phpseclib/phpseclib/commit/7b7d254a6c64719d13595e81ea45ec330f02c950

If someone wants them removed from the 1.0 or 2.0 branches they're free to submit a PR. Or maybe I'll get bored and do it one day myself idk lol

hellopablo:
@terrafrost I also encountered a similar situation - I might be wrong but I think the syntax is wrong rather than not supported? I believe the format is `/**#@+` rather than `/**@+` (note the missing `#`).

FWIW, I encountered the issue in `phpseclib\System\SSH\Agent` (Agent forwarding address).

Happy to submit a PR, but wanted to check with you first before doing so as you may have it covered already!

terrafrost:
> FWIW, I encountered the issue in `phpseclib\System\SSH\Agent` (Agent forwarding address).

That namespace is used by the 2.0 branch. I only updated the 3.0 and master branches. I can update the 2.0 branch too, however.

> I believe the format is `/**#@+` rather than `/**@+` (note the missing `#`).

What I did for the 3.0 branch was that I just removed it all together. Unless I see documentation suggesting that `/**#@+` is the correct format I'd just opt for removing it.

PR #1559, opened 2020-12-23
Add PHP 8 support for PublicKeyLoader::load()

kylekatarnls:
The following example taken from documentation
https://phpseclib.com/docs/rsa#raw-rsa-public-keys

Pass in PHP 7.4
But not in the stable current version of PHP 8.0
```
There was 1 error:

1) Unit_Crypt_RSA_LoadKeyTest::testLoadModulusAndExponent
TypeError: strpos(): Argument #1 ($haystack) must be of type string, array given

C:\Users\kylek\PhpstormProjects\phpseclib\phpseclib\Crypt\EC\Formats\Keys\PKCS1.php:64
C:\Users\kylek\PhpstormProjects\phpseclib\phpseclib\Crypt\Common\AsymmetricKey.php:162
C:\Users\kylek\PhpstormProjects\phpseclib\phpseclib\Crypt\PublicKeyLoader.php:42
C:\Users\kylek\PhpstormProjects\phpseclib\tests\Unit\Crypt\RSA\LoadKeyTest.php:44
```
https://travis-ci.com/github/phpseclib/phpseclib/jobs/464235441#L3282

### DIFF
--- a/phpseclib/Crypt/EC/Formats/Keys/PKCS1.php
+++ b/phpseclib/Crypt/EC/Formats/Keys/PKCS1.php
@@ -61,7 +61,7 @@ abstract class PKCS1 extends Progenitor
     {
         self::initialize_static_variables();
 
-        if (strpos($key, 'BEGIN EC PARAMETERS') && strpos($key, 'BEGIN EC PRIVATE KEY')) {
+        if (is_string($key) && strpos($key, 'BEGIN EC PARAMETERS') && strpos($key, 'BEGIN EC PRIVATE KEY')) {
             $components = [];
 
             preg_match('#-*BEGIN EC PRIVATE KEY-*[^-]*-*END EC PRIVATE KEY-*#s', $key, $matches);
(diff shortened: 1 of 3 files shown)


terrafrost:
I took your first commit and cherry-picked that to the 3.0 branch and then merged that into master.

I went with a somewhat different solution for the second commit:

https://github.com/phpseclib/phpseclib/commit/02fa3b142e34e90c37d53e41b599477d1353c391

I prefer this method over your method because this method, in theory, takes into account objects that can be cast to strings via `__toString`.  It's also more consistent with the other classes:

https://github.com/phpseclib/phpseclib/blob/3.0.1/phpseclib/Crypt/EC/Formats/Keys/PKCS8.php#L85
etc

Another difference between my approach and your approach is that mine throws an exception whereas yours does not. For `PublicKeyLoader::load(...)` it doesn't matter much because the load method captures all exceptions thrown, but if you do `\phpseclib3\Crypt\EC::loadFormat('PKCS1', [])`...  well, it still doesn't matter a ton, because the parent class does that check, but still...  idk lol

Nice find and thanks!

kylekatarnls:
Yes this also matches closer the PHP 7 behavior, it sounds legit.

PR #1578, opened 2021-01-07
Allow to pass end date as DateTime

kylekatarnls:
While the method intents to support date objects, it's actually made impossible due to `strtolower()` throwing an error. This preliminary check will actually allow to enter into the `else` without triggering the `if (!is_object($date) || !($date instanceof DateTimeInterface))` e.g. support `DateTimeInterface` as parameter.

### DIFF
--- a/phpseclib/File/X509.php
+++ b/phpseclib/File/X509.php
@@ -3061,7 +3061,7 @@ class X509
 
           -- http://tools.ietf.org/html/rfc5280#section-4.1.2.5
         */
-        if (strtolower($date) == 'lifetime') {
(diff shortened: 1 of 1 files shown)


terrafrost:
Good catch!

I suppose we could make it check to see if it was an object that could be cast to a string with the `__toString` magic method ([Strings::is_stringable](https://github.com/phpseclib/phpseclib/blob/d812384c2dced7e6ff39206b333a12fb2630d5ad/phpseclib/Common/Functions/Strings.php#L383) does that) but it also seems like that's such an extreme edge case that it's not even worth considering atm lol

PR #1585, opened 2021-01-18
Allow newer versions of paragonie/random_compat

KacerCZ:
Allows to use version of paragonie/random_compat which is NOOP for PHP 7.0 or newer.


terrafrost:
I cherry-picked this into the 3.0 branch and merged that into master:

https://github.com/phpseclib/phpseclib/commit/e8b4e4d4df51a5ed901ec31a22dcf04c994dc6e8

Thanks!

PR #1615, opened 2021-02-21
fix undefined index notice in stream touch()

lightray22:
I was getting a PHP notice in Stream.php.  It would happen when using touch( ) with the stream wrapper and no atime/mtime given.            

`vendor/phpseclib/phpseclib/phpseclib/Net/SFTP/Stream.php(450): Undefined offset: 0`

### DIFF
--- a/phpseclib/Net/SFTP/Stream.php
+++ b/phpseclib/Net/SFTP/Stream.php
@@ -447,7 +447,7 @@ class Stream extends SFTP
         //     and https://github.com/php/php-src/blob/master/main/php_streams.h#L592
         switch ($option) {
             case 1: // PHP_STREAM_META_TOUCH
-                return $this->sftp->touch($path, $var[0], $var[1]);
+                return $this->sftp->touch($path, $var[0] ?? null, $var[1] ?? null);
             case 2: // PHP_STREAM_OWNER_NAME
             case 3: // PHP_STREAM_GROUP_NAME
                 return false;
--- a/phpseclib/Net/SFTP/Stream.php
+++ b/phpseclib/Net/SFTP/Stream.php
@@ -447,11 +447,9 @@ class Stream extends SFTP
         //     and https://github.com/php/php-src/blob/master/main/php_streams.h#L592
         switch ($option) {
             case 1: // PHP_STREAM_META_TOUCH
-            {
                 $time = isset($var[0]) ? $var[0] : null;
                 $atime = isset($var[1]) ? $var[1] : null;
                 return $this->sftp->touch($path, $time, $atime);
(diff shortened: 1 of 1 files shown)


terrafrost:
It's not clear to me under what circumstances the two optional parameters wouldn't be set. I tried calling `touch()` with one parameter and `stream_metadata` was called with three parameters. Sample code:

```php
<?php

class VariableStream {
    public function __call($name, $arguments)
    {
        echo $name . "\n";
        var_dump($arguments);
        $name = '_' . $name;
        if (!method_exists($this, $name)) {
            return false;
        }
        return $this->$name(...$arguments);
    }
}

stream_wrapper_register("var", "VariableStream");
    
touch("var://myvar");
```
https://3v4l.org/UeVTQ

If these parameters aren't being passed to `stream_metadata` under as yet unknown conditions then we should probably make the second parameter default to `time()` since that's what https://www.php.net/touch by default does. The third parameter would also need to be set to whatever the second parameter was set to if it wasn't explicitly set.

But what I just don't understand is why it's not getting all three parameters for you whereas in the 3v4l.org thing it's getting all three parameters in every single version of PHP that they have...

lightray22:
I use both PHP 7.4 and 8.0 for development, not sure which I was using at the time.  I'll do more digging.

This change, while seeming like it should be unnecessary, is at least a safe one that shouldn't cause any side effects.  The actual touch() implementation in SFTP.php looks like it takes care of the params being null. so we're good there.

```
    public function touch($filename, $time = null, $atime = null)
    {
        .......
        if (!isset($time)) {
            $time = time();
        }
        if (!isset($atime)) {
            $atime = $time;
        }
```

lightray22:
Also it looks like in the link you sent, it's displaying the behavior I described.

It says the arguments it's sending are... 
```

array(3) {
  [0]=>
  string(11) "var://myvar"
  [1]=>
  int(1)
  [2]=>
  array(0) {
  }
}
```

The third argument (the array with zero length) is what it's sending to stream_metadata() so it makes sense that you would get a notice about indices 0 and 1 being undefined.

terrafrost:
Oh right - good call on both points.

> Also it looks like in the link you sent, it's displaying the behavior I described.

I think I just saw that it was two parameters and assumed that the two parameters were atime and mtime after a super quick glance. Sorry about that!

Anyway, I'm going to squash this into one commit, cherry-pick that one commit to the 1.0 branch and then merge up to master.

terrafrost:
See https://github.com/phpseclib/phpseclib/commit/488db53bf7cbd90d5910838162b275f43e16363b

Thanks!

PR #1624, opened 2021-03-04
PHPDoc fixes for Net/SFTP class

MathObsessed:
Added some PHPDoc fixes according to use cases from the docs (https://phpseclib.com/docs/sftp) to satisfy PHPStan

### DIFF
--- a/phpseclib/Net/SFTP.php
+++ b/phpseclib/Net/SFTP.php
@@ -430,7 +430,7 @@ class SFTP extends SSH2
      * Login
      *
      * @param string $username
-     * @param string[] ...$args
+     * @param string[]|string ...$args
      * @throws \UnexpectedValueException on receipt of unexpected packets
      * @return bool
      * @access public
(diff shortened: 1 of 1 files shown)


phpseclib/Net/SFTP.php
```diff
@@ -430,7 +430,7 @@ public function __construct($host, $port = 22, $timeout = 10)
      * Login
      *
      * @param string $username
-     * @param string[] ...$args
```
terrafrost, line 433: So technically, the second parameter can also be ]a \phpseclib3\Crypt\Common\PrivateKey object](https://phpseclib.com/docs/auth#public-key). It could also be an [array of arrays](https://phpseclib.com/docs/auth#keyboard-interactive)

MathObsessed: Made some adjustments. Hope I understood your intention correctly

terrafrost:
I squashed the commits into one and cherry-picked them into the 3.0 branch:

https://github.com/phpseclib/phpseclib/commit/906a5fafabe5e6ba51ef3dc65b2722a677908837

I then merged 3.0 into master.

Probably should get the phpDoc fixes for the login method applied to SSH2::login as well

PR #1635, opened 2021-04-06
cleanup RSA PKCS#1 v1.5 signature verification (CVE-2021-30130)

terrafrost:


MrPetovan:
Is this going to be fixed in the version 2.0 branch? We're using version 2.0.4 over at https://github.com/friendica/friendica-addons and I would prefer not have to either downgrade to version 2.0.31 or figure out if we can use version 3.0.7.

terrafrost:
First, the fix is in the 2.0 branch. The release that contains this has been tagged as 2.0.31, which is not a downgrade from 2.0.4 - it's an _upgrade_.

Second, the only vulnerabilities are in the 3.0 branch. Quoting the disclosure (which I can send to you if you'd like; I kinda wish the author would make it public but whatever):

> In short, we have found 4 leniencies in phpseclib v3 (relaxed mode)'s RSA PKCS#1 v1.5 signature verification (2 vulnerabilities with Bleichenbacher-style low public exponent RSA Signature Forgery and 2 bugs causing interoperability issue [accepting invalid signature value that should have been rejected]). We have also found incompatibility issue in phpseclib v1, v2, v3 (strict mode)'s RSA PKCS#1 v1.5 signature verification suffering from rejecting valid signatures whose encoded message uses implicit hash algorithm's NULL parameter.

So to sum it up, there were five issues total.

1. Two were vulnerabilities in v3.0
2. Two were bugs in v3.0
3. One was a bug in v1.0, v2.0 and v3.0.

In other words, v2.0 does not have a vulnerability, according to the original paper.

MrPetovan:
Thank you so much for the answer, for some reason I read 2.0.31 as 2.0.3.1 which would have made it a downgrade from 2.0.4. Sorry about the confusion.

khallouki:
Could this also be fixed in version ^1.0?

terrafrost:
Altho phpseclib isn't mentioned in it, [Analyzing Semantic Correctness with Symbolic Execution: A Case Study on PKCS#1 v1.5 Signature Verification](https://www.ndss-symposium.org/ndss-paper/analyzing-semantic-correctness-with-symbolic-execution-a-case-study-on-pkcs1-v1-5-signature-verification/) talks about the kinds of issues that this PR addresses. The reporter of these issues is one of the authors of that paper

edit (11/22/2025): phpseclib _is_ mentioned in [Morpheus: Bringing The (PKCS) One To Meet the Oracle](https://dl.acm.org/doi/pdf/10.1145/3460120.3485382)

PR #1646, opened 2021-04-23
Allow string for BigInteger

kylekatarnls:
In real life BigInteger are used to store number too big to be `int` so `string` is actually the natural type to use


terrafrost:
I cherry-picked this to the 3.0 branch and merged to master:

https://github.com/phpseclib/phpseclib/commit/0f77cf3a59e2e6dfbb743eb9a6c3791d979726f9

Thanks!

Issue #1670, opened 2021-06-10
enableDatePreservation on SFTP not working


paolosupena:
Hi! Am unsure if some people have encountered this before. However, I am having trouble with getting the data preservation working on version 3 when uploading a file.  The connection and the file gets uploaded ok. It is just the date modified for the file that is not preserved.  

This code is a rough proof of concept

```
require_once __DIR__ . '/vendor/autoload.php';
use phpseclib3\Net\SFTP;
define('NET_SFTP_LOGGING', SFTP::LOG_COMPLEX);

$server = "xxx.com";
$username = "xxxx";
$password = "xxxxxx";

$sftp = new SFTP($server);
$sftp->enableDatePreservation();

if($sftp->login($username, $password)){

$test1 = $sftp->put("public_html/Content/". basename(filename.remote), filename.local, SFTP::SOURCE_LOCAL_FILE);

if ($test1) { echo "upload ok"; } else { echo "not ok"; }
```

I also tried putting the date preservation after the PUT but it did not work. Anyone having the same issues?

terrafrost:
Does https://github.com/terrafrost/phpseclib/tree/3.0-date-preservation/phpseclib work for you?

paolosupena:
Hi! Thanks for the response. Apologies for the delay on my response. Had other things to work on.

I did test out your version, unfortunately it isn't working on our setup. 

Here is a screenshot of the simple log though: https://www.screencast.com/t/6b4cyovAW .  
Here is the complex log: https://www.screencast.com/t/dubvwW9YI8b (right at the part of NET_SFTP_FSETSTAT) where I think we do see a failure. 

We are running on PHP 7.3.28.

Hope that does give some insight.

Thanks!

terrafrost:
I made another commit to that branch. LMK if it works for you?

paolosupena:
Hi! 

I did test out the latest changes.  While, on the logs, I do see a success: https://www.screencast.com/t/OncOvlNgwHMP , 
checking on the server, I still see the uploaded date as the last modified date, seen here: https://www.screencast.com/t/vBUbieINbyjx . 

I did do around 4 tests, just in case my files were recently updated or anything like that, but the last test, I did make sure that it was only modified/generated last year definitely.

Hopefully that helps a bit.
Thanks a lot!

terrafrost:
That's actually consistent with my own observations. The access time is modified but the modified time isn't. I suspect the same will be true for you if you do `stat()` on the file.

[...]

PR #1671, opened 2021-06-14
fix exception thrown because of undefined constant with php7.4

adrienHoareau:
Hello,

When using php7.4 and symfony an exception is thrown while using SFTP class :
`WARNING   [php] Warning: Use of undefined constant NET_SFTP_CHANNEL - assumed 'NET_SFTP_CHANNEL' (this will throw an Error in a future version of PHP)`

This is because of this line : 
`if ($this->channel_status[NET_SFTP_CHANNEL] === NET_SSH2_MSG_CHANNEL_CLOSE) {`
NET_SFTP_CHANNEL is not defined as a constant.

### DIFF
--- a/phpseclib/Net/SFTP.php
+++ b/phpseclib/Net/SFTP.php
@@ -2940,7 +2940,7 @@ class SFTP extends SSH2
         while (strlen($this->packet_buffer) < 4) {
             $temp = $this->get_channel_packet(self::CHANNEL, true);
             if ($temp === true) {
-                if ($this->channel_status[NET_SFTP_CHANNEL] === NET_SSH2_MSG_CHANNEL_CLOSE) {
+                if ($this->channel_status[self::CHANNEL] === NET_SSH2_MSG_CHANNEL_CLOSE) {
                     $this->channel_close = true;
                 }
                 $this->packet_type = false;


terrafrost:
Yup - found that the other day myself and included it in https://github.com/terrafrost/phpseclib/commit/cee6849f2e57e920ac6ec625caf1aa27c7eff6f5 . That's not in the 3.0 branch yet tho - was hoping that the person who submitted #1670 could confirm that it worked. It wasn't working for me but neither was the regular touch either so I'm thinking that the issue was something specific to my system and I didn't want to dig into it any farther at that point.

Anyway, I can merge your thing and then merge that one. I don't believe it'd create a merge conflict since both commits are modifying the same bit of code in the exact same way but who knows.

Maybe this evening or maybe in a few days!

adrienHoareau:
ok thanks for your reply.
While waiting  for the merge if other people have the same problem a quick fix is to just create a constant :
```
if (! defined('NET_SFTP_CHANNEL')) {
    define('NET_SFTP_CHANNEL', 0x100);
}
```

terrafrost:
I cherry-picked this into the 3.0 branch and then merged that into master:

https://github.com/phpseclib/phpseclib/commit/f1c41646878eff641f8ad066f2d94566a229119d

Thanks!

PR #1678, opened 2021-06-30
Throw an error when a packet can not be read to prevent infinite loop

ashokadewit:
The method SSH2::_get_channel_packet() returns false when
the connection has been closed. This would cause an infinite loop
in SFTP::_get_sftp_packet because it keeps reading until the packet
buffer length is 4. To solve this a runtime exception is thrown when the
return value is false.

### DIFF
--- a/phpseclib/Net/SFTP.php
+++ b/phpseclib/Net/SFTP.php
@@ -3108,6 +3108,9 @@ class SFTP extends SSH2
                 $this->packet_buffer = '';
                 return false;
             }
+            if ($temp === false) {
+                throw new \RuntimeException('Packet can not be read');
+            }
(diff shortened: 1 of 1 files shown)


ashokadewit:
This change is not needed in ^3.0 because there the method never returns false but throws an exception instead.

terrafrost:
The 2.0 branch doesn't throw exceptions - it does user_error(). Exceptions are better, which is why 3.0 adopted them, but, in-so-far as 2.0 is concerned, they'd be a BC break, in-so-far a I'm concerned.

Here's what I did instead:

https://github.com/phpseclib/phpseclib/commit/a02526e5e0e17f6d15438a773e9cd3d930fb5048

Thanks!

PR #1684, opened 2021-07-26
timeout in _get_channel_packet during  _init_sftp_connection should be a failure

ziktar:
If using SFTP to connect to a server that timed out during the initial connection, you'll get weird error messages, as the _init_sftp_connection will ignore the timeout and continue to try to connect, but now the underlying SSH connection is in a weird state. Instead, it should treat that as a failure.

The underlying issue can be reproduced by changing the timeout to a really small value (say 1 second) and then connecting to a server far enough away that it'll timeout during connection.

### DIFF
--- a/phpseclib/Net/SFTP.php
+++ b/phpseclib/Net/SFTP.php
@@ -467,6 +467,8 @@ class SFTP extends SSH2
         $response = $this->_get_channel_packet(self::CHANNEL, true);
         if ($response === false) {
             return false;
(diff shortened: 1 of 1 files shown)


terrafrost:
I cherry-picked this to the 1.0 branch and then merged it to 2.0, 3.0 and master:

https://github.com/phpseclib/phpseclib/commit/dab514faffcc1adc5d34078b2e0804add1f90e4c

Thanks!

PR #1689, opened 2021-08-13
PublicKeyLoader: make all methods static

claudepache:
Those methods are intended to be used statically.

However, as they are not static, calling them statically will trigger a deprecation warning in PHP 7 and throw an error in PHP 8.

### DIFF
--- a/phpseclib/Crypt/PublicKeyLoader.php
+++ b/phpseclib/Crypt/PublicKeyLoader.php
@@ -72,7 +72,7 @@ abstract class PublicKeyLoader
      * @param string|array $key
      * @param string $password optional
      */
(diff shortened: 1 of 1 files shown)


claudepache:
The same issue plagues the class phpseclib3\Crypt\Common\AsymmetricKey

terrafrost:
I noticed this a while ago and made a commit to fix it:

https://github.com/phpseclib/phpseclib/commit/a5fcb44e280ac69b2c12ec8f0e8d3de22a25559a

I guess I didn't catch all of them!

Anyway, I cherry-picked your commit into the 3.0 branch and merged that into master and I made a new commit for the other methods in AsymmetricKey.

Thanks!

PR #1691, opened 2021-08-23
add option to allow arbitrary length packets

NickLassondeAtLandmarkGlobal:
The company OpenText has their own SFTP server, and when navigating a directory with a large number of files, the entire directory listing is in one packet. As such, the packet can get arbitrarily long. Add a flag to allow any length packets. This probably should still default to being off, as this could maybe be a security issue where the server could claim to send any length packet. For our needs, we trust the server enough to allow them to send any length, and the directory could have any number of files, so putting any "reasonable" limit on it might not work, thus the request is to allow arbitrarily long packets.

### DIFF
--- a/phpseclib/Net/SFTP.php
+++ b/phpseclib/Net/SFTP.php
@@ -270,6 +270,21 @@ class SFTP extends SSH2
     var $preserveTime = false;
 
     /**
+     * Arbitrary Length Packets Flag
+     *
+     * Determines whether or not packets of any length should be allowed,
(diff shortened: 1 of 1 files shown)


terrafrost:
I cherry picked this to the 1.0 branch and merged to 2.0 and 3.0 and master.

Made a few other tweaks to your changes as well in a second commit:

https://github.com/phpseclib/phpseclib/commit/fc40727cfbbc8c997ffdcef9511a3b22db290bd8

Thanks!

PR #1699, opened 2021-09-29
mbstring.func_overload is deprecated in php 7.2 and removed in php 8.0

ym2048:
According to the official PHP manual  https://www.php.net/manual/en/migration72.deprecated.php and https://www.php.net/manual/en/migration80.incompatible.php, mbstring.func_overload is deprecated in php 7.2 and removed in php 8.0, So I think we should add PHP version judgment here.

### DIFF
--- a/phpseclib/bootstrap.php
+++ b/phpseclib/bootstrap.php
@@ -11,7 +11,8 @@
  */
 if (extension_loaded('mbstring')) {
     // 2 - MB_OVERLOAD_STRING
-    if (ini_get('mbstring.func_overload') & 2) {
+    // mbstring.func_overload is deprecated in php 7.2 and removed in php 8.0.
+    if ((PHP_VERSION_ID < 80000) && (ini_get('mbstring.func_overload') & 2)) {
         throw new UnexpectedValueException(
             'Overloading of string functions using mbstring.func_overload ' .
             'is not supported by phpseclib.'


terrafrost:
Are there any drawbacks to calling `ini_get('mbstring.func_overload')`? I tried it on a PHP 8 system with mbstring installed and didn't get any warnings or notices or errors or anything like that. Seems to behave like `ini_get('mbstring.zzz')` does in that there's no harm in calling it.

I mean, I don't have a problem changing it up but neither am I seeing an issue with keeping it as is? Unless there's something I'm missing?

ym2048:
According to your suggestion, I conducted a detailed test. The conclusion is that there is no harm effect if this judgment is not added. The only function of the  PHP version judgment  is to clearly remind users that mbstring.func_overload entry is removed in PHP>=8.0 and there is no need to worry about the following exception.

Here is the test script:

<?php
$multiByteStrings = 'hello,世界!';
$len_strlen = strlen($multiByteStrings);
$len_mb_strlen = mb_strlen($multiByteStrings);

echo "value of mbstring.func_overload in php.ini is :\n";
var_dump(ini_get('mbstring.func_overload'));
echo "strlen    result: $len_strlen \n";
echo "mb_strlen result: $len_mb_strlen \n";

In PHP >= 8.0, ini_get ('mbstring.func_overload') will always return false, regardless of whether mbstring.func_overload is set and value is set in php.ini.
In php 7.2~7.4, mbstring.func_overload=2 entry  in php.ini will result in an E_DEPRECATED level notice.

ym2048:
Here is the test script:

<?php
$multiByteStrings = 'hello,世界!';
$len_strlen = strlen($multiByteStrings);
$len_mb_strlen = mb_strlen($multiByteStrings);

echo "value of mbstring.func_overload in php.ini is :\n";
var_dump(ini_get('mbstring.func_overload'));
echo "strlen    result: $len_strlen \n";
echo "mb_strlen result: $len_mb_strlen \n";

In PHP >= 8.0, ini_get ('mbstring.func_overload') will always return false, regardless of whether mbstring.func_overload is set and value is set in php.ini.
In php 7.2~7.4, mbstring.func_overload=2 entry  in php.ini will result in an E_DEPRECATED level notice.

terrafrost:
Well I'm not getting any errors.

```php
error_reporting(E_ALL);

echo extension_loaded('mbstring') ? 't' : 'f';
echo "\n";
echo ini_get('mbstring.func_overload');
```
Running that on PHP 7.4.25 on Windows 10 yields no error.

I also tried that same script on PHP 7.4.25 on Ubuntu 20.04 without issue. More specifically, I did it in a Docker container:

```
FROM php:7.4

RUN printf "\nmbstring.func_overload=2" >> /usr/local/etc/php/php.ini
```
The output of the test script is the same in both cases:

```
t
2
```
No errors are output...

---
**Sema Reaction:** :hammer_and_wrench: This code needs a fix

ym2048:
You cannot call error_reporting function in code.You need to set `error_reporting = E_ALL` directly in php.ini.

Here is the source code of ext/mbstring/mbstring.c (begin line is1727 ) from PHP 7.4.5

PHP_RINIT_FUNCTION(mbstring)
{
        if (MBSTRG(func_overload)){
                zend_error(E_DEPRECATED, "The mbstring.func_overload directive is deprecated");

                CG(compiler_options) |= ZEND_COMPILE_NO_BUILTIN_STRLEN;
        }

so E_DEPRECATED notice is generated in RINIT stage.
And here is the default error_ reporting value in php.ini
error_ reporting = E_ ALL & ~E_ DEPRECATED & ~E_ STRICT

then the E_DEPRECATED  is already shielded before we call  `error_reporting(E_ALL);` in php code

terrafrost:
Well I'm still not able to reproduce the issue.

My updated Dockerfile:

```
FROM php:7.4

RUN printf "\nmbstring.func_overload=2\nerror_reporting = E_ALL" >> /usr/local/etc/php/php.ini
```

If we go inside the container we can see that everything is setup as you described:

```
$ php -v
PHP 7.4.25 (cli) (built: Oct 22 2021 17:23:33) ( NTS )
Copyright (c) The PHP Group
Zend Engine v3.4.0, Copyright (c) Zend Technologies
$ php -i | grep mbstring.func_overload
mbstring.func_overload => 2 => 2
$ php -i | grep error_reporting
error_reporting => 32767 => 32767
```
Here's my test code:
```
$ php -r "echo ini_get('mbstring.func_overload') . \"\n\";"
2
```

terrafrost:
Maybe I'm missing something?

---
**Sema Reaction:** :question: I have a question

ym2048:
Add `log_errors = On` config item to php.ini, and the run the testscript, the deprecated message will appear.
Here is the changed Dockerfile

```
FROM php:7.4

RUN printf "\nmbstring.func_overload=2\nerror_reporting = E_ALL\nlog_errors = On" >> /usr/local/etc/php/php.ini
```

terrafrost:
I'm able to reproduce it - thanks!

I cherry-pick'd your commit to the 1.0 branch and then merged into 2.0 / 3.0 / master:

https://github.com/phpseclib/phpseclib/commit/c91671f17634fdc98ce7d780e6f4dcb00b51bf71

PR #1710, opened 2021-11-09
Add linter task to phing for checking old PHP versions

### LINKED ISSUE #1709: Suggestion: PHP Linting for Supported Versions
I hate adding cruft to OSS, so take it or leave it, but if it's helpful here's a very simple + performant one-liner you can add to the build process to syntax check all files using a given PHP version. It prints syntax errors to stdout and returns exit code 1 if any syntax errors are found. (one-liner pulled from discussions [here](https://gist.github.com/mathiasverraes/3096500) ) 

Syntax check all files in current directory (recursive):
```bash
find . -type f -name '*.php' -print0 \
  | xargs -0 -n1 -P$(nproc)php5.3 -l -n \
  | (! grep -v "^No syntax errors detected" )
```

If you don't have a certain php version installed, you can use Docker. This bash script can be used in place of the php binary for a given version (though it will be much slower - a few seconds). v5.3 is used as an example:
```bash
#! /usr/bin/env bash
docker run -t --rm -v "$PWD:$PWD" -w $PWD php:5.3-cli php -l -n "$@"
```

Here's example output for the error fixed in PR #1706 :
```
> time find . -type f -name '*.php' -print0  \
  | xargs -0 -n1 -P$(nproc) php5.3 -l -n  \
  | (! grep -v "^No syntax errors detected" )

Parse error: syntax error, unexpected '[' in ./phpseclib/Net/SSH2.php on line 3558
Errors parsing ./phpseclib/Net/SSH2.php
xargs: php: exited with status 255; aborting

real    0m0.030s
user    0m0.072s
sys     0m0.016s

> echo $?
1
```

rsynnest:
Closes #1709


terrafrost:
In .travis.yml codesniffer is setup to run on PHP 5.5:

https://github.com/phpseclib/phpseclib/blob/1.0/.travis.yml#L36

One thing I feel like we could do is to update the .travis.yml on, say, PHP 5.6 or whatever, to lint for every version of PHP. Or at the very least lint for versions of PHP for which there aren't unit tests.

So like even tho the Travis CI "container" is running PHP 5.6, using Docker containers, it could still lint on PHP 4.4, PHP 5.0, etc.

I suppose something else we could do is to have a new build version [i]just[/i] for linting. This build version could be identified by the use of a custom env variable as is done here:

https://github.com/terrafrost/php-amqplib/commit/317f91e3400c2fb63e7b8f4943c154e733e276fe

This .travis.yml might help as well:

https://github.com/phpseclib/benchmarks/blob/bcmath-no-openssl/.travis.yml

It'd also be good to have this on phpseclib 1.0 since that can run on PHP 4.4+. phpseclib 2.0 requires PHP 5.3+ but it's no longer being tested on PHP 5.3 via continuous integration as of https://github.com/phpseclib/phpseclib/commit/0539780a863da2a6fb50ae0a53a1b6a4b9b2b067

idk I just think having this linting done via continuous integration would be really cool!

rsynnest:
CI is definitely the place, I'll look into adding it. I use Gitlab CI so
didn't know to look for a .travis.yml

PR #1730, opened 2022-01-08
SFTP: fix chgrp() for version < 4

PetrP:
$uid and $gid were flipped with [dbfc7622571](https://github.com/phpseclib/phpseclib/commit/dbfc76225714d28d41f832cf424ff6696d14e344#diff-2c14c913d58f0f70ffc8a953c7608e51ac68d69aa85a664375d1a84b7f62d98fR1696)

The test does not test this bug. I don't know how to do it for any ssh on PHPSECLIB_SSH_HOSTNAME which may or may not have some users/groups or permission to change them. If you give me some idea on how to do it, I will be happy to implement it, or remove the test if you see it as unnecessary.
Also seems like all tests are in chain with `@depends` which is strange but I followed it.

### DIFF
--- a/phpseclib/Net/SFTP.php
+++ b/phpseclib/Net/SFTP.php
@@ -1635,7 +1635,7 @@ class SFTP extends SSH2
     public function chgrp($filename, $gid, $recursive = false)
     {
         $attr = $this->version < 4 ?
-            pack('N3', NET_SFTP_ATTR_UIDGID, $gid, -1) :
+            pack('N3', NET_SFTP_ATTR_UIDGID, -1, $gid) :
             Strings::packSSH2('Nss', NET_SFTP_ATTR_OWNERGROUP, '', $gid);
 
         return $this->setstat($filename, $attr, $recursive);
--- a/tests/Functional/Net/SFTPUserStoryTest.php
+++ b/tests/Functional/Net/SFTPUserStoryTest.php
@@ -765,5 +765,22 @@ class Functional_Net_SFTPUserStoryTest extends PhpseclibFunctionalTestCase
         $list_cache_disabled = $sftp->rawlist('.', true);
 
         $this->assertEquals($list_cache_enabled, $list_cache_disabled, 'The files should be the same regardless of stat cache', 0.0, 10, true);
+
+        return $sftp;
+    }
+
+    /**
+     * @depends testRawlistDisabledStatCache
+     */
+    public function testChownChgrp($sftp)
+    {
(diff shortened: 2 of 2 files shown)


terrafrost:
I cherry-picked this into the 1.0 branch and then merged up to the 2.0, 3.0 and master branches:

https://github.com/phpseclib/phpseclib/commit/888867e7b64709e884a2feedd1e2ac38a1051669

> Also seems like all tests are in chain with `@depends` which is strange but I followed it.

Yah idk why it does that. I didn't write the initial unit test suite. It's been like that since there inception:

https://github.com/phpseclib/phpseclib/commit/fe50a6cdacc11d5cffa17f11d60b34f125fd53b0

I mean, it makes sense to make it so that phpseclib doesn't need to log into the SFTP server for each and every test but it seems like you could achieve the same result by making every test (after the testPasswordLogin() test) `@depends testPasswordLogin` instead of the way it currently works.

It does look like there was definitely an intentional effort, tho, to keep the tests linear:

https://github.com/phpseclib/phpseclib/commit/8a43f02ae98ffd302a6658fcddcd102fb62d25ce

Maybe there's a reason for this, maybe not, idk. Either way it doesn't seem like it's causing any issues and, as the saying goes, if it ain't broke, don't fix it.

PetrP:
Thanks

> Maybe there's a reason for this, maybe not, idk. Either way it doesn't seem like it's causing any issues and, as the saying goes, if it ain't broke, don't fix it.

Small issues it causes:
- can't run single test
- because `Functional_Net_SFTPUserStoryTest::testReadableWritable` was failing I had to manually skip it to be able to run my test at the end

Both not big deal, but small annoyances.

Also feedback from someone who tried to run tests, there is no mentioning in [documentation](https://github.com/phpseclib/phpseclib#contributing) about env variables `PHPSECLIB_SSH_*` for tests to not be skipped, I needed to figure it out from code.

terrafrost:
Good points.

If you did a PR to change the behavior I'd prob accept it. It's not a super huge priority for me atm.

I mean, at some point, I would like to revisit the unit tests. I'd like CI on appveyor for Windows and CI on github actions. And I feel like at one point I came across a shim for removed phpunit functions but I'm not able to find it. If I do find it then maybe I could use that as well.

But it's not that high a priority.

Also, confession, I normally just test straight on Travis CI lol. I mean, I'll test locally but I don't use unit tests in my local tests - I just use them on Travis CI. Which does pose some problems because open source projects have to periodically request new credits, which is a bit annoying, but whatever

PR #1746, opened 2022-01-28
Fixed psalm level 6 errors in phpseclib/Net/

hc-jworman:
I fixed all errors Psalm found on error-level 6 except for all the undefined constants.
Psalm Version: 4.9.3

Mostly, it is just fixing type-hints, but there were a few instances of actual bugs.

For instance, `Strings::unpack()` and `Strings::unpackSS2` do not exist. I suspect it was a typo. There was also the use of an undefined variable `$atr`, which I suspect was another typo meant to be `$attr`.

### DIFF
--- a/phpseclib/Crypt/Common/SymmetricKey.php
+++ b/phpseclib/Crypt/Common/SymmetricKey.php
@@ -204,6 +204,13 @@ abstract class SymmetricKey
         self::ENGINE_OPENSSL_GCM => 'OpenSSL (GCM)'
     ];
 
+    /** @var string|null */
+    public $fixed;
+    /** @var string|null */
+    public $invocation_counter;
+    /** @var string|null */
+    public $name;
+
     /**
      * The Encryption Mode
      *
--- a/phpseclib/Crypt/Hash.php
+++ b/phpseclib/Crypt/Hash.php
@@ -69,6 +69,11 @@ class Hash
      */
     const PADDING_SHAKE = 3;
 
+    /** @var int|false */
+    public $etm;
+    /** @var string|false */
+    public $name;
+
     /**
      * Padding Type
      *
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -1204,7 +1204,7 @@ class SSH2
          * Typehint is required due to a bug in Psalm: https://github.com/vimeo/psalm/issues/7508
          * @var \WeakReference<SSH2>|SSH2
          */
-        self::$connections[$this->getResourceId()] = \class_exists('WeakReference')
+        self::$connections[$this->getResourceId()] = class_exists('WeakReference')
             ? \WeakReference::create($this)
             : $this;
 
@@ -1417,7 +1417,7 @@ class SSH2
         if (!$this->send_kex_first) {
             $response = $this->get_binary_packet();
 
-            if (\is_bool($response) || !strlen($response) || ord($response[0]) != NET_SSH2_MSG_KEXINIT) {
+            if (is_bool($response) || !strlen($response) || ord($response[0]) != NET_SSH2_MSG_KEXINIT) {
                 $this->bitmap = 0;
                 throw new \UnexpectedValueException('Expected SSH_MSG_KEXINIT');
             }
@@ -1553,7 +1553,7 @@ class SSH2
             $kexinit_payload_server = $this->get_binary_packet();
 
             if (
-                \is_bool($kexinit_payload_server)
+                is_bool($kexinit_payload_server)
                 || !strlen($kexinit_payload_server)
                 || ord($kexinit_payload_server[0]) != NET_SSH2_MSG_KEXINIT
             ) {
@@ -3306,7 +3306,7 @@ class SSH2
     private function get_binary_packet($skip_channel_filter = false)
     {
         if ($skip_channel_filter) {
-            if (!\is_resource($this->fsock)) {
+            if (!is_resource($this->fsock)) {
                 throw new \InvalidArgumentException('fsock is not a resource.');
             }
             $read = [$this->fsock];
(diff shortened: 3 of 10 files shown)


phpseclib/Crypt/Common/SymmetricKey.php
```diff
@@ -204,6 +204,13 @@ abstract class SymmetricKey
         self::ENGINE_OPENSSL_GCM => 'OpenSSL (GCM)'
     ];
 
+    /** @var string|null */
+    public $fixed;
+    /** @var string|null */
+    public $invocation_counter;
+    /** @var string|null */
+    public $name;
```
terrafrost, line 212: It's adding these because they're used in SSH2.php. SSH2.php _shouldn't_ be using dynamic properties as they're being [deprecated in PHP 8.2](https://wiki.php.net/rfc/deprecate_dynamic_properties) so it's a good find but I think it'd be better to define them in SSH2.php. I'll try to look into this later today or tomorrow.

phpseclib/Crypt/Hash.php
```diff
@@ -69,6 +69,11 @@ class Hash
      */
     const PADDING_SHAKE = 3;
 
+    /** @var int|false */
+    public $etm;
+    /** @var string|false */
+    public $name;
```
terrafrost, line 75: This should be in SSH2.php, as well, as they don't mean anything outside of SSH2.php.

phpseclib/Net/SSH2.php
```diff
@@ -204,7 +204,7 @@ class SSH2
     /**
      * The Socket Object
      *
-     * @var object
+     * @var resource|closed-resource|null
```
terrafrost, line 207: Didn't know closed-resource was a distinctly different thing from resource lol

phpseclib/Net/SSH2.php
```diff
@@ -1411,7 +1417,7 @@ private function connect()
         if (!$this->send_kex_first) {
             $response = $this->get_binary_packet();
 
-            if (!strlen($response) || ord($response[0]) != NET_SSH2_MSG_KEXINIT) {
+            if (\is_bool($response) || !strlen($response) || ord($response[0]) != NET_SSH2_MSG_KEXINIT) {
```
terrafrost, line 1420: I'd say let's remove the `\`. I don't do them for any other function.

I did try to play around with this general idea a while back to see if it'd yield any speedups and it didn't:

https://github.com/terrafrost/phpseclib/commit/437e7f2cf2eb75ce42c3a33a2b29c57098946194

It's possible my testing was inadequate tho but idk lol

phpseclib/Net/SSH2.php
```diff
@@ -1546,7 +1552,11 @@ private function key_exchange($kexinit_payload_server = false)
 
             $kexinit_payload_server = $this->get_binary_packet();
 
-            if (!strlen($kexinit_payload_server) || ord($kexinit_payload_server[0]) != NET_SSH2_MSG_KEXINIT) {
+            if (
+                \is_bool($kexinit_payload_server)
+                || !strlen($kexinit_payload_server)
+                || ord($kexinit_payload_server[0]) != NET_SSH2_MSG_KEXINIT
+            ) {
```
terrafrost, line 1559: I almost wonder if it'd be better to do this:

```
switch (true) {
    case is_bool($kexinit_payload_server):
    case !strlen($kexinit_payload_server):
    case ord($kexinit_payload_server[0]) != NET_SSH2_MSG_KEXINIT:
        // ...
}
```
To each their own but I think that's a little easier to read then `$a || $b || $c || $d`, etc

phpseclib/Net/SSH2.php
```diff
@@ -3290,12 +3300,15 @@ protected function reset_connection($reason)
...
      * @param bool $skip_channel_filter
-     * @return string
+     * @return bool|string
      * @access private
      */
     private function get_binary_packet($skip_channel_filter = false)
     {
         if ($skip_channel_filter) {
+            if (!\is_resource($this->fsock)) {
+                throw new \InvalidArgumentException('fsock is not a resource.');
```
terrafrost, line 3310: Same thing as before - let's remove the `\`. Either all functions should have them or none should imho

terrafrost:
Code changes mostly look good! I left a few in-line comments!

phpseclib/Net/SSH2.php
hc-jworman: closed-resource is a psalm specific type used to catch bugs:
```php
// is a resource
$this->fsock = \fsockopen(...);

// now it is a closed-resource
\fclose($this->fsock);

// Psalm would throw an error here now
fputs($this->fsock, ...);
```

phpseclib/Net/SSH2.php
hc-jworman: It is definitely a speed up, but very (very) slight. I just have my IDE set to automatically add the slash. But following the coding style is more important.

Anyways, if this is worth commenting about, maybe we should implement a checker so it is not up for debate. php-cs-fixer has a rule for this: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/doc/rules/function_notation/native_function_invocation.rst

phpseclib/Net/SSH2.php
hc-jworman: I find switch's very ugly and less readable, so I am only going to do this if you insist.

phpseclib/Net/SSH2.php
hc-jworman: Removed. Here and everywhere else in the project

phpseclib/Net/SSH2.php
terrafrost: It's fine as is!

phpseclib/Net/SSH2.php
terrafrost: I think that's a good idea! Submit a PR? 😁

terrafrost:
I cherry picked both of these commits into the 3.0 branch:

https://github.com/phpseclib/phpseclib/commit/e3b71763ae71752e535f742129043d2f9f3fd54a
https://github.com/phpseclib/phpseclib/commit/0a9fc99dc8b2eb3aa3ce5c37e7752b2e7683bbf9

I then implemented the alternative fixes I proposed for the dynamic properties:

https://github.com/phpseclib/phpseclib/commit/215fd61d1276b63d2a0a39f306b9d8a338211217
https://github.com/phpseclib/phpseclib/commit/3f2a5aa4cdbfd3def85fc2fbc0a0bf26dc255832

I then merged the 3.0 branch into the master branch.

Thanks!

terrafrost:
It'd be cool if psalm could be incorporated into Travis CI cause, I can safely say, without continuous integration, it'll prob fall out of compliance pretty quick

PR #1749, opened 2022-01-31
Fixed risky tests

hc-jworman:
Tests are marked as risky when no assertions are made.
Mostly just added a snapshot-like test; for one of them I just added an `self::assertTrue(true)`.

### DIFF
--- a/tests/Unit/Crypt/AES/TestCase.php
+++ b/tests/Unit/Crypt/AES/TestCase.php
@@ -200,9 +200,7 @@ abstract class Unit_Crypt_AES_TestCase extends PhpseclibTestCase
         $aes->setKey($key);
         $aes->setIV($iv);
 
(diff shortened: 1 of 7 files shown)


tests/Unit/Crypt/AES/TestCase.php
```diff
@@ -200,9 +200,7 @@ public function testContinuousBufferBattery($op, $mode, $test)
         $aes->setKey($key);
         $aes->setIV($iv);
 
-        if (!$this->_checkEngine($aes)) {
```
hc-jworman, line 203: This test was broken; _checkEngine does not return anything

terrafrost:
I cherry-picked this into the 3.0 branch and merged into master:

https://github.com/phpseclib/phpseclib/commit/5916c2bff8c74d5e9df5cf3bcd5d4705ce89d33f

Thanks!

PR #1751, opened 2022-02-03
Remove Dynamic Constants

hc-jworman:
Replaced dynamic constants with pseudo-enums

### DIFF
--- a/phpseclib/Net/SFTP.php
+++ b/phpseclib/Net/SFTP.php
@@ -43,6 +43,7 @@ use phpseclib3\Net\SFTP\OpenFlag;
 use phpseclib3\Net\SFTP\OpenFlag5;
 use phpseclib3\Net\SFTP\PacketType;
 use phpseclib3\Net\SFTP\StatusCode;
+use phpseclib3\Net\Ssh2\MessageType as Ssh2MessageType;
 
 /**
  * Pure-PHP implementations of SFTP.
@@ -380,7 +381,7 @@ class SFTP extends SSH2
 
         $packet = Strings::packSSH2(
             'CsN3',
-            SshMsg::CHANNEL_OPEN,
+            Ssh2MessageType::CHANNEL_OPEN,
             'session',
             self::CHANNEL,
             $this->window_size,
@@ -389,7 +390,7 @@ class SFTP extends SSH2
 
         $this->send_binary_packet($packet);
 
-        $this->channel_status[self::CHANNEL] = SshMsg::CHANNEL_OPEN;
+        $this->channel_status[self::CHANNEL] = Ssh2MessageType::CHANNEL_OPEN;
 
         $response = $this->get_channel_packet(self::CHANNEL, true);
         if ($response === true && $this->isTimeout()) {
@@ -398,7 +399,7 @@ class SFTP extends SSH2
 
         $packet = Strings::packSSH2(
             'CNsbs',
-            SshMsg::CHANNEL_REQUEST,
+            Ssh2MessageType::CHANNEL_REQUEST,
             $this->server_channels[self::CHANNEL],
             'subsystem',
             true,
@@ -406,7 +407,7 @@ class SFTP extends SSH2
         );
         $this->send_binary_packet($packet);
 
-        $this->channel_status[self::CHANNEL] = SshMsg::CHANNEL_REQUEST;
+        $this->channel_status[self::CHANNEL] = Ssh2MessageType::CHANNEL_REQUEST;
 
         $response = $this->get_channel_packet(self::CHANNEL, true);
         if ($response === false) {
@@ -418,7 +419,7 @@ class SFTP extends SSH2
             // is redundant
             $packet = Strings::packSSH2(
                 'CNsCs',
-                SshMsg::CHANNEL_REQUEST,
+                Ssh2MessageType::CHANNEL_REQUEST,
                 $this->server_channels[self::CHANNEL],
                 'exec',
                 1,
@@ -426,7 +427,7 @@ class SFTP extends SSH2
             );
             $this->send_binary_packet($packet);
 
-            $this->channel_status[self::CHANNEL] = SshMsg::CHANNEL_REQUEST;
+            $this->channel_status[self::CHANNEL] = Ssh2MessageType::CHANNEL_REQUEST;
 
             $response = $this->get_channel_packet(self::CHANNEL, true);
             if ($response === false) {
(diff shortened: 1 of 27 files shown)


phpseclib/Net/SFTP.php
```diff
@@ -3025,8 +2857,8 @@ private function parseTime($key, $flags, &$response)
     {
         $attr = [];
         list($attr[$key]) = Strings::unpackSSH2('Q', $response);
-        if ($flags & NET_SFTP_ATTR_SUBSECOND_TIMES) {
-            list($attr[key . '-nseconds']) = Strings::unpackSSH2('N', $response);
```
hc-jworman, line 3029: Check out this typo! `key` is suppose to be `$key`

phpseclib/Net/SshMsgCustom.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace phpseclib3\Net;
+
+/**
+ * @internal
+ */
+abstract class SshMsgCustom
```
terrafrost, line 8: I think it'd be better if this were named Extra, instead of Custom.

phpseclib/Net/SshMsgCustom.php
```diff
@@ -0,0 +1,27 @@
...
+{
+    const KEXDH_INIT = 30;
+    const KEXDH_REPLY = 31;
```
terrafrost, line 11: These were originally in `$this->message_numbers`:

https://github.com/phpseclib/phpseclib/blob/cc65aa3b0e9efd982688f921028cbdb158c2905e/phpseclib/Net/SSH2.php#L1211-L1241

Why were they moved out? Like everything else in that array was put in `phpseclib/Net/SshMsg.php` but not these?

phpseclib/Net/SFTP.php
terrafrost: Nice find!

phpseclib/Net/SFTP/OpenFlag5.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace phpseclib3\Net\SFTP;
+
+/**
+ * @internal
+ */
+abstract class OpenFlag5
```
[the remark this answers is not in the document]
hc-jworman: All comments added back

phpseclib/Net/SshMsgCustom.php
hc-jworman: Done

phpseclib/Net/SshMsgCustom.php
hc-jworman: The spec does not give this 30-49 a specific value, so I thought it should go in extra with the other 30-49 values.

To compensate for this, I had to update lines like this: `$this->updateLogHistory('SSH_MSG_KEXDH_INIT', 'SSH_MSG_KEX_ECDH_INIT');` to be `$this->updateLogHistory('UNKNOWN (30)', 'SSH_MSG_KEX_ECDH_INIT');`

phpseclib/Net/SshDisconnect.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace phpseclib3\Net;
+
+use phpseclib3\Common\ConstantUtilityTrait;
+
+/**
+ * @internal
+ */
+abstract class SshDisconnect
```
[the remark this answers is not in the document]
hc-jworman: Done

phpseclib/Net/Ssh2/MessageType.php
```diff
@@ -0,0 +1,49 @@
+<?php
+
+namespace phpseclib3\Net\Ssh2;
```
terrafrost, line 3: Could you rename it to SSH2 instead of Ssh2?

Both SSH2 and SFTP are all caps as is the newly added SFTP directory. By that same reasoning it stands to reason that it ought to be SSH2 instead of Ssh2.

Thanks!

hc-jworman: Done

phpseclib/Net/SshMsgCustom.php
terrafrost: That's kinda a bug in the SSH specs. It's _referenced_ in the original SSH2 RFCs (in particular, [RFC4253](https://datatracker.ietf.org/doc/html/rfc4253)) but it's not actually defined anywhere.

I asked about where it was defined back in 2007, when I first started working on phpseclib, at https://groups.google.com/g/comp.security.ssh/c/f5jNVb0fMe4/m/U93doB0Sfn4J, and was told:

> They don't. It accidentally got left out, I think that when it got
(correctly) removed from [SSH-NUMBERS], it was also mistakenly removed
from [SSH-TRANSPORT] as well.
>
> Irritatingly, the erratum for this also seems to have an error, in that
it lists SSH_MSG_KEXDH_REPLY as being 3 when it's obviously 31. Time
for an erratum for the erratum, I suppose.

terrafrost: I went ahead and made this change myself!:

https://github.com/phpseclib/phpseclib/commit/b3f14dee377f107ac3bdf7927312148a492f0d85

terrafrost:
I merged this into master but not into the 3.0 branch, mainly because some of the constant changes are BC breaks.

Like doing `$sftp->rawlist()`. That returns an array with a bunch of subarrays that contain type as one of the keys. type is set to an integer but the way I mention you know what that actually corresponds to in the docs (https://phpseclib.com/docs/sftp#directory-listing) is to reference the constant. Since that constant has been replaced that method no longer works.

But that's why master exists too - sometimes being bound by backwards compatibility can be rather debilitating.

hc-jworman:
We can still make it BC without too much trouble, just need to redefine the old global constants in terms of the new class constants.

terrafrost:
> We can still make it BC without too much trouble, just need to redefine the old global constants in terms of the new class constants.

Yah we could but tbh idk that I really care lol. I'll backport the bug fix you found but it's kinda a hassle cherry-picking stuff to the 3.0 branch too lol. I mean, if you want to do it, feel free, but otherwise idk that I really care that much lol.

Like for new features that improve the user experience that don't break BC they prob ought to be backported. Bug fixes obviously ought to be. But most of this PR is an architectural improvement. Outside of the one line bug you found it's not a bug fix nor is it a new feature that improves the user experience. It's def a change that should be made (and indeed one that was on my radar, albeit with no ETA lol) but yah.

PR #1754, opened 2022-02-04
Whitespace php-cs-fixer rules

hc-jworman:
More rules to make the styling more consistent


phpseclib/Math/BigInteger/Engines/PHP.php
```diff
@@ -578,7 +578,7 @@ protected function divideHelper(PHP $y)
             $lhs =  new static();
             $rhs =  new static();
         }
-	if (static::class != get_class($temp)) {
```
terrafrost, line 581: Should be eight spaces - not four.

phpseclib/Net/SSH2.php
```diff
@@ -4637,7 +4637,7 @@ public static function getSupportedEncryptionAlgorithms()
             'twofish192-cbc', // OPTIONAL          Twofish with a 192-bit key
             'twofish256-cbc',
             'twofish-cbc',    // OPTIONAL          alias for "twofish256-cbc"
-                              //                   (this is being retained for historical reasons)
```
terrafrost, line 4640: Another place I think we should do `// @codingStandardsIgnoreStart` and `// @codingStandardsIgnoreEnd`

phpseclib/System/SSH/Agent.php
```diff
@@ -192,7 +192,7 @@ public function requestIdentities()
                 case 'ecdsa-sha2-nistp256':
                 case 'ecdsa-sha2-nistp384':
                 case 'ecdsa-sha2-nistp521':
-		    $key = PublicKeyLoader::load($key_type . ' ' . base64_encode($key_blob));
```
terrafrost, line 195: Eight more spaces should be added here.

terrafrost (commented):
For the most part I like it but, in DES.php, we went from this:

```php
        $block = ($shuffleip[ $r        & 0xFF] & "\x80\x80\x80\x80\x80\x80\x80\x80") |
                 ($shuffleip[($r >>  8) & 0xFF] & "\x40\x40\x40\x40\x40\x40\x40\x40") |
                 ($shuffleip[($r >> 16) & 0xFF] & "\x20\x20\x20\x20\x20\x20\x20\x20") |
                 ($shuffleip[($r >> 24) & 0xFF] & "\x10\x10\x10\x10\x10\x10\x10\x10") |
                 ($shuffleip[ $l        & 0xFF] & "\x08\x08\x08\x08\x08\x08\x08\x08") |
                 ($shuffleip[($l >>  8) & 0xFF] & "\x04\x04\x04\x04\x04\x04\x04\x04") |
                 ($shuffleip[($l >> 16) & 0xFF] & "\x02\x02\x02\x02\x02\x02\x02\x02") |
                 ($shuffleip[($l >> 24) & 0xFF] & "\x01\x01\x01\x01\x01\x01\x01\x01");
```

To this:

```php
        $block = ($shuffleip[$r        & 0xFF] & "\x80\x80\x80\x80\x80\x80\x80\x80") |
                 ($shuffleip[($r >>  8) & 0xFF] & "\x40\x40\x40\x40\x40\x40\x40\x40") |
                 ($shuffleip[($r >> 16) & 0xFF] & "\x20\x20\x20\x20\x20\x20\x20\x20") |
                 ($shuffleip[($r >> 24) & 0xFF] & "\x10\x10\x10\x10\x10\x10\x10\x10") |
                 ($shuffleip[$l        & 0xFF] & "\x08\x08\x08\x08\x08\x08\x08\x08") |
                 ($shuffleip[($l >>  8) & 0xFF] & "\x04\x04\x04\x04\x04\x04\x04\x04") |
                 ($shuffleip[($l >> 16) & 0xFF] & "\x02\x02\x02\x02\x02\x02\x02\x02") |
                 ($shuffleip[($l >> 24) & 0xFF] & "\x01\x01\x01\x01\x01\x01\x01\x01");
```
Like before everything lined up and after they don't. If we don't care about stuff lining up then we ought to replace stuff like `$l        & 0xFF` with `$l & 0xFF`.

But I think the best solution is to wrap `// @codingStandardsIgnoreStart` and `// @codingStandardsIgnoreEnd` around these pieces of code. It's already done for Twofish.php.

There's a little bit of this going on in Math/BigInteger/Engines/Engine.php, Blowfish.php, Prime.php, Salsa20.php and more in Twofish.php as well.

For a lot of the stuff in ASN1/Maps/* we're going from this:

```php
            'authorityCertIssuer'       => [
                                             'constant' => 1,
                                             'optional' => true,
                                             'implicit' => true
                                           ] + GeneralNames::MAP,
            'authorityCertSerialNumber' => [
                                             'constant' => 2,
                                             'optional' => true,
                                             'implicit' => true
                                           ] + CertificateSerialNumber::MAP
```

To this:

```php
            'authorityCertIssuer'       => [
                'constant' => 1,
                'optional' => true,
                'implicit' => true
            ] + GeneralNames::MAP,
            'authorityCertSerialNumber' => [
                'constant' => 2,
                'optional' => true,
                'implicit' => true
            ] + CertificateSerialNumber::MAP
```
I'm okay with this but can we remove the white space between the key and the `=> [`? eg. can `'authorityCertIssuer'       => [` be turned into `'authorityCertIssuer' => [`?

terrafrost:
For some context, incidentally, phpseclib 1.0 and 2.0 do CS checks in PHP 5.5 via PHP_CodeSniffer:

https://github.com/phpseclib/phpseclib/blob/1.0/.travis.yml#L38
https://github.com/phpseclib/phpseclib/blob/1.0/build/build.xml

phpseclib 3.0, however, dropped PHP 5.5 support and I guess I didn't move the code sniffing to a newer version of PHP!

phpseclib/Net/SSH2.php
hc-jworman: Not needed, I can manually align

phpseclib/System/SSH/Agent.php
hc-jworman: Fixed

terrafrost:
Sorry for the delay - I've had a busy week!

Anyway,

> @codingStandardsIgnoreStart only applies to php_codesniffer, not php-cs-fixer.

Maybe we can use php_codesniffer then? I _really_ dislike the idea of hard and fast rules that don't take the specifics into consideration. Like even PSR-12 allows for flexibility with statements like "_There MUST NOT be a hard limit on line length_".

More generally, I feel like some of the latest php-cs-fixer changes are decidedly making the code _less_ readable. eg.

```php
                $t = $sbox1[($b1 >> 24) & 0x3F] ^ $sbox2[($b2 >> 24) & 0x3F]
                    ^ $sbox3[($b1 >> 16) & 0x3F] ^ $sbox4[($b2 >> 16) & 0x3F]
                    ^ $sbox5[($b1 >> 8) & 0x3F] ^ $sbox6[($b2 >> 8) & 0x3F]
                    ^ $sbox7[$b1 & 0x3F] ^ $sbox8[$b2 & 0x3F]
                    ^ $l;
```
vs.
```php
                $t = $sbox1[($b1 >> 24) & 0x3F] ^ $sbox2[($b2 >> 24) & 0x3F] ^
                     $sbox3[($b1 >> 16) & 0x3F] ^ $sbox4[($b2 >> 16) & 0x3F] ^
                     $sbox5[($b1 >>  8) & 0x3F] ^ $sbox6[($b2 >>  8) & 0x3F] ^
                     $sbox7[ $b1        & 0x3F] ^ $sbox8[ $b2        & 0x3F] ^ $l;
```
I don't consider the php-cs-fixer changes to be an improvement. Sure, that's gonna be subjective for any arbitrary segment of code, but between coding standards and programs, people have opinions programs don't.

Also, the newest commit moves operators on multi line lines to the start of each new line as opposed to the end. eg. before we had this (from Crypt/DH/Formats/Keys/PKCS1.php):

```php
        return "-----BEGIN DH PARAMETERS-----\r\n" .
               chunk_split(base64_encode($params), 64) .
               "-----END DH PARAMETERS-----\r\n";
```
...now we have this:

```php
        return "-----BEGIN DH PARAMETERS-----\r\n"
               . chunk_split(base64_encode($params), 64)
               . "-----END DH PARAMETERS-----\r\n";
```
I'm not sure that's an improvement. Like before, it was pretty obvious, at a glace, that the first block of `-`'s before the BEGIN and END were the same length - now it's not so clear.

I mean, in some cases, doing the operator before makes more sense then doing it after, but it's not really a hard and fast rule.

As a general rule, unless [PSR-2](https://www.php-fig.org/psr/psr-2/) or [PSR-12](https://www.php-fig.org/psr/psr-12/) (or some yet to be written PR) explicitly say to do something (eg. tabs vs spaces) I'd just assume let the person making the commit use their own discretion. php-cs-fixer takes a lot of this autonomy away.

The closest I could find w.r.t. the operators in the PSR rules is this (from PSR-12):

> Boolean operators between conditions MUST always be at the beginning or at the end of the line, not a mix of both.

So it looks like even the PSR standards leave this decision up to the implementer.

terrafrost:
Like I guess to kinda sum up my philosophy...  the PSR standards don't say anything about which of these two you ought to do:

```php
switch (true) {
    case $a:
    case $b:
    case $c:
        // do something
}
```
vs.

```php
if ($a || $b || $c) {
    // do something
}
```
Consequently, it's really in the eye of the beholder which one ought to be employed. I prefer the former, you said, elsewhere, that you prefer the latter. Which is fine.

If you were to implement a new feature that used the latter as opposed to the former I'd probably let it be merged as is and wouldn't make any changes to it. I mean, by merging it in I assume all responsibility for future maintenance of it and fixing any bugs that might exist in it and what not but I can work with either, in spite of my preference for the former.

Like as a general rule, when the CS standards say nothing I'll defer to the judgement of the person who made the original change. I mean, that's not a hard and fast rule, either. I may, at my discretion (and I suppose depending on my mood lol) change it or I may not, altho I do believe I, historically, generally let the judgement of the PR author stand. I suppose I could do a survey of all the comments I've ever made to add some "oomph" behind my supposition but that seems like overkill lol.

hc-jworman:
Maybe close this, in favor of https://github.com/phpseclib/phpseclib/pull/1760

PR #1760, opened 2022-02-17
php_codesniffer added to build

hc-jworman:
Added phpcs to the build
Ran phpbfs auto-fixer, excluded the things that couldn't be auto-fixed


terrafrost:
I cherry picked this into the 3.0 branch and merged into master.

I applied the `// @codingStandardsIgnoreStart` thing to ChaCha20.php with a later commit:

https://github.com/phpseclib/phpseclib/commit/f2e9dd993d9131e570f54ba9f6c38a39db11aaba

I'm a little surprised it's replacing stuff like `(int)($x / $y)` with `(int) ($x / $y)` but whatever - I can live with it.

I also do have a slight deference to `$var.= 'whatever';` over `$var .= 'whatever';`. I think the following more clearly illustrates why this might be desirable:

```php
        $key = "PuTTY-User-Key-File-$version: $type\r\n";
        $key.= "Encryption: $encryption\r\n";
        $key.= "Comment: $comment\r\n";
```
But I can live with that too. The issues this is identifying and can identify in the future more than make up for a few quirks that I can live with!

PR #1764, opened 2022-02-23
Remove some psr12 phpcs rule exclusions

hc-jworman:
PSR4 for tests, single file per class, and fall-through comments in switches


terrafrost:
Something else for you to work on, if you're down for it...

#1770 is about deprecation warnings on PHP 8.1. I would have thought that PHPUnit would treat those as unit tests that error'd out but, clearly, it's not, and it's not clear to me why.

PR #1765, opened 2022-02-24
actions/stale github action

jack-worman:
I noticed there are many stale issues and pull request. This github action would help manage them.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,20 @@
+name: Stale issue handler
+on:
+    schedule:
+        -   cron: 0 0 * * *
+
+permissions:
+    issues: write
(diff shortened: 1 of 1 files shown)


terrafrost:
idk how I feel about this. Periodically I look back at old tickets and manually close those that either nothing can be done out (more info is needed and it's been x months since the OP last posted so I'm unlikely to get more info) or that have actually been fixed. But it's also not uncommon that I come across tickets that haven't been addressed as well.

I'll mull on this some!

hc-jworman:
Let's increase the "days-before-stale" to 180 days (6 months) and the "days-before-close" to 30 days (1 month).

And there is always the option to add the "keep" label to an issue or pull request, to have it stay around as long as you want.

PR #1785, opened 2022-04-22
Remove_@access

hc-jworman:
Removed @access doc-block tags because they are redundant and sometimes inaccurate.
Also, removed @package and @category


terrafrost:
phpseclib/System/SSH/Agent/Identity.php is kinda weird. It's not really intended to be used by end users, so it shouldn't exactly be treated as public, but there's no PHP keyword that really captures what it ought to be. Like for that one I did `@access internal`, which is suboptimal but idk lol. I'm okay with this being removed however.

I also note that the visibility attribute for stuff like `const MODE_CFB = 3;` in SymmetricKey.php have been removed. PHP introduce support for class constant visibility until [PHP 7.1.0](https://wiki.php.net/rfc/class_const_visibility). phpseclib 3.0 supports versions of PHP as low as 5.6. I'll def change that for the master branch (which will eventually become the 4.0 branch) but I have not yet made that change. Like right now I get that everything is implicitly public but, in this case, I think showing how the visibility would be if phpseclib supported PHP 7.1.0+ is useful.

If it doesn't interfere with static analysis or IDE's like PHP Storm or whatever I'm kinda inclined to keep the visibility stuff for class constants.

hc-jworman:
"@access internal" may not exist, but "@internal" does. It is supported by static analysis tools like PhpStorm and Psalm: https://psalm.dev/docs/annotating_code/supported_annotations/#psalm-internal

I will happily bump master's PHP version to 7.1 and add visibility to every constant. Just give me the go ahead.

terrafrost:
> `@access internal` may not exist, but `@internal` does. It is supported by static analysis tools like PhpStorm and Psalm: https://psalm.dev/docs/annotating_code/supported_annotations/#psalm-internal

Yah - let's do that.

Thanks!

> I will happily bump master's PHP version to 7.1 and add visibility to every constant. Just give me the go ahead.

Two concerns I have with that:

1. I'd just assume cherry-pick this to the 3.0 branch and under no circumstance will I introduce a BC breaking change in 3.0 by upping the minimally required version of PHP. Like as is, my above comments not withstanding, there should be no reason this can't be cherry-picked into the 3.0 branch. If you, in this branch, you bump the minimally required PHP version to 7.1, I can no (easily) longer cherry pick.

    I mean, 3.0 is still an actively supported branch and will be for the foreseeable future. Most PR's are to the master branch and if I'm not prepared to backport bug fixes and what not from the master branch to the 1.0 / 2.0 / 3.0 branches than those branches need to be EOL'd and I'm not prepared to do that.

    Now this PR isn't for a bug fix, per say, but still...  as written it shouldn't be too difficult to cherry-pick this to the 3.0 branch. That changes if you bump master's PHP version to 7.1.

2. I'd rather bump the minimum supported PHP version for the master branch iteratively after comprehensively supporting all the new / relevant features that each successive PHP version offers. Like PHP 7.0 introduced strict typing. So I'd just assume not bump the minimum required PHP version of phpseclib to 7.0 until the master branch uses strict typing. For PHP 7.1, in addition to class constant visibility there's also:

    - [symmetric array destructuring](https://wiki.php.net/rfc/short_list_syntax) (ie. instead of `list($a, $b) = ...;` it'll then be `[$a, $b] = ...;`)
    - [negative string offsets](https://wiki.php.net/rfc/negative-string-offsets) (ie. `substr($parts[$i - 1], -1)` becomes `$parts[$i - 1][-1]` and `$line[strlen($line) - 1]` becomes `$line[-1]`, etc)

    If composer.json is updated to bump the minimally required PHP version to 7.1 before all that is done then I'm more likely to forget that it's something I even _want_ to do.

terrafrost:
Re-reading my post... it came off harsher than I intended it too and I apologize for that!

hc-jworman:
I didn't think it was harsh... :)

So after this gets merged, you would be okay with me implementing 7.0 features in master?

terrafrost:
> So after this gets merged, you would be okay with me implementing 7.0 features in master?

Yah that'd be awesome! I guess for the time being you could do features up to 8.0 but let's start off with 7.0, then 7.1 and 7.2, until we get to 8.0.

hc-jworman:
Added `@internal` to `phpseclib3\System\SSH\Agent\Identity`

terrafrost:
I squashed these into one commit and cherry-picked it into the 3.0 branch:

https://github.com/phpseclib/phpseclib/commit/8855351cbb1ee8c3793f80f7c87aac5c2bf145eb

PR #1801, opened 2022-06-10
get_sftp_packet: Do not bail on receiving spurious (empty) packets.

datacentergl:
Previously the function would return `false' on such a case, causing
the caller to abort on an UnexpectedValueException; but then the pre-
viously unreceived packet would arrive at further get_sftp_packet calls
which of course do not expect it, aborting again on a UnexpectedValue-
Exception and so on (basically the remaining of the session would be
broken).

The typical error messages in such a broken session look like this:
 1  Error:Expected NET_SFTP_HANDLE or NET_SFTP_STATUS. Got packet type:
 2  Error:Expected NET_SFTP_DATA or NET_SFTP_STATUS. Got packet type: 102
(etc.)

### DIFF
--- a/phpseclib/Net/SFTP.php
+++ b/phpseclib/Net/SFTP.php
@@ -3095,7 +3095,7 @@ class SFTP extends SSH2
                 }
                 $this->packet_type = false;
                 $this->packet_buffer = '';
-                return false;
+                return ($this->channel_close ? false : $this->get_sftp_packet($request_id));
             }
             $this->packet_buffer .= $temp;
         }
@@ -3117,9 +3117,12 @@ class SFTP extends SSH2
         while ($tempLength > 0) {
             $temp = $this->get_channel_packet(self::CHANNEL, true);
             if (is_bool($temp)) {
+                if ($this->channel_status[self::CHANNEL] === NET_SSH2_MSG_CHANNEL_CLOSE) {
+                    $this->channel_close = true;
+                }
                 $this->packet_type = false;
                 $this->packet_buffer = '';
-                return false;
+                return ($this->channel_close ? false : $this->get_sftp_packet($request_id));
             }
             $this->packet_buffer .= $temp;
             $tempLength -= strlen($temp);


terrafrost:
I do not agree with this change.

From Net/SSH2.php:

```php
     * Returns the data as a string. bool(true) is returned if:
     *
     * - the server closes the channel
     * - if the connection times out
     * - if the channel status is CHANNEL_OPEN and the response was CHANNEL_OPEN_CONFIRMATION
     * - if the channel status is CHANNEL_REQUEST and the response was CHANNEL_SUCCESS
     *
     * bool(false) is returned if:
     *
     * - if the channel status is CHANNEL_REQUEST and the response was CHANNEL_FAILURE
```
Your code checks to see if the server closed the channel but it doesn't check to see if the channel timed out. Unless I'm mistaken your code would just infinitely loop in the event of a timeout. This can be seen by doing `$ssh->get('/dev/random');`. Without your changes this times out after 10s. With your changes it just hangs there into perpetuity.

As for the other two....  `get_sftp_packet()` isn't called unless the channel status is CHANNEL_DATA so it shouldn't be possible for those to have any impact.

If there's a specific problem you're having that motivated this change could you provide the full SSH2 logs? eg. do `define('NET_SSH2_LOGGING', 3)` and then send me what you got? You could either post it here (well in a pastebin.com link and then post it here) or email it to terrafrost@php.net.

That said, you do have one addition that I think is good:

```
                if ($this->channel_status[self::CHANNEL] === NET_SSH2_MSG_CHANNEL_CLOSE) {
                    $this->channel_close = true;
                }
```
You've added that to the `while ($tempLength > 0) {` loop. I'll try to add that tomorrow!

terrafrost:
> That said, you do have one addition that I think is good:
> 
> ```
>                 if ($this->channel_status[self::CHANNEL] === NET_SSH2_MSG_CHANNEL_CLOSE) {
>                     $this->channel_close = true;
>                 }
> ```
> 
> You've added that to the `while ($tempLength > 0) {` loop. I'll try to add that tomorrow!

I added it:

https://github.com/phpseclib/phpseclib/commit/85205bf6d5ea62249ed31585e31c39ed16747714

Thanks!

terrafrost:
Closing due to inactivity.

Issue #1805, opened 2022-06-20
Static call to non-static method (results in an error starting from PHP 8.0)


vadym-ovechkin:
Hello, during preparation for migration of my project to PHP 8.1 I ran a static analyzer which have detected one problem in the package.

It complains about this line: https://github.com/phpseclib/phpseclib/blob/545b43cd700c986b0b11caa0fdf8b0d0ac18f6c6/phpseclib/File/X509.php#L2643

The `dnsName` method is not static but called as a static one. This results in an error starting from PHP 8.0 (see https://www.php.net/manual/en/migration80.incompatible.php)

I guess this helper method can simply be marked as static, and this will solve the problem

jack-worman:
Would you provide a PR?

jack-worman:
Closed by #1806

PR #1806, opened 2022-06-20
Prevent static call to non-static method error

vadym-ovechkin:
Starting from PHP 8.0 the following line will result in an error:
```
$altName = array_map(['\phpseclib3\File\X509', 'dnsName'], $subject->domains); 
```
The dnsName method is not static but called as a static one. This results in an error starting from PHP 8.0 (see https://www.php.net/manual/en/migration80.incompatible.php)

See issue #1805

### DIFF
--- a/phpseclib/File/X509.php
+++ b/phpseclib/File/X509.php
@@ -3714,7 +3714,7 @@ class X509
      * @param string $domain
      * @return array
      */
-    private function dnsName($domain)
+    private static function dnsName($domain)
     {
         return ['dNSName' => $domain];
     }


jack-worman (approved):

terrafrost:
This'll need to be cherry-picked to the 3.0 branch as well. As is no one should be using the master branch as it's API can drastically change from commit to commit but the 3.0 branch otoh

jack-worman:
I was about to, but work emergency came up :)

terrafrost:
Oh lol - no worries! I went ahead and did it:

https://github.com/phpseclib/phpseclib/commit/7de44b745d61765c9ccf9155d8ef4508c840fc37

Thanks!

PR #1815, opened 2022-07-07
Upgrading to Php 7.1

jack-worman:
* Removed mcrypt
* php-cs-fixer upgrade rules
    * added class constant visibility
    * added a few missing `declare(strict_types=1)`
* Added test to ensure certain error type are not baselined
* Using negative string offsets
* `$password` parameter typehints changed to `?string $password = null`


terrafrost:
I'll try to take as time permits! I'm pretty busy through the weekend so we'll see what we see.

I'd also like to see the following new 7.1 PHP features implemented:

- class constant visibility (eg. public `const WHATEVER = ...;` instead of just `const WHATEVER = ...;`)
- negative string offsets (`substr($parts[$i - 1], -1)` becomes `$parts[$i - 1][-1]` and `$line[strlen($line) - 1]` becomes `$line[-1]`, etc)

And also some of the functions that currently return false or arrays made to return null or arrays (or whatever) instead to facilitate simplified type hinting.

terrafrost (commented):
I like the fact that you're using the heredoc syntax for the closures!

That said, I feel like it would have been better and less overwhelming a code review if this had been split up into multiple parts. eg. separate PR's for:

- adding public visibility to class methods
- removing mcrypt
- removing unused variables
- adding primitive type hinting
- making use of negative string offsets
- adding declare(strict_types=1) to files that did not already have it
- replacing list() with []

That way some individual changes could be backported to 3.0.

Regardless, thanks for these changes!

PR #1818, opened 2022-07-12
Fix return type comment on Crypt/RSA::createKey()

anthonyryan1:
Function returns a Crypt/RSA/PrivateKey and updating
the PHPDoc for it will help with static analysis.

### DIFF
--- a/phpseclib/Crypt/RSA.php
+++ b/phpseclib/Crypt/RSA.php
@@ -286,7 +286,7 @@ abstract class RSA extends AsymmetricKey
      *
      * The public key can be extracted from the private key
      *
-     * @return RSA
+     * @return RSA\PrivateKey
      */
     public static function createKey(int $bits = 2048)
     {


phpseclib/Crypt/RSA.php
```diff
@@ -286,7 +286,7 @@ public static function setOpenSSLConfigPath(string $val): void
      *
      * The public key can be extracted from the private key
      *
-     * @return RSA
+     * @return RSA\PrivateKey
      */
     public static function createKey(int $bits = 2048)
```
jack-worman, line 291: Add the return type directly:
```suggestion
    public static function createKey(int $bits = 2048): PrivateKey
```

phpseclib/Crypt/RSA.php
```diff
@@ -286,7 +286,7 @@ public static function setOpenSSLConfigPath(string $val): void
      *
      * The public key can be extracted from the private key
      *
-     * @return RSA
+     * @return RSA\PrivateKey
```
jack-worman, line 289: Psalm CI is now failing, I believe we should just baseline the new errors.

jack-worman (changes requested):

PR #1822, opened 2022-07-19
Add PHP 8.2 support

CallMeJon90:
I tried running tests on docker with php8.2 and there were some problems that I was able to solve.
These changes work on php8.1 but I have not tried on earlier versions.


phpseclib/Math/PrimeField.php
```diff
@@ -41,6 +41,11 @@ class PrimeField extends FiniteField
      */
     protected $instanceID;
 
+    /**
+     * @var BigInteger
+     */
+    protected $modulo;
```
terrafrost, line 47: I think we can just delete `$this->modulo = $modulo;` from the constructor. Doesn't appear to be used anywhere else.

terrafrost (changes requested):
I tried to add 8.2 to .github/workflows/ci.yml but got an error. This is something that'll need to be figured out

phpseclib/Math/BigInteger/Engines/PHP/Reductions/EvalBarrett.php
```diff
@@ -98,7 +98,7 @@ protected static function generateCustomReduction(PHP $m, string $class): callab
                 $rhs = new ' . $class . '();
                 $lhs->value = $n;
                 $rhs->value = [' .
-                implode(',', array_map('self::float2string', $m)) . '];
+                implode(',', array_map(self::class . '::float2string', $m)) . '];
```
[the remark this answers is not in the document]
jack-worman: https://php.watch/versions/8.2/partially-supported-callable-deprecation

CallMeJon90:
Thank you for your responses.
Now I have an urgency to fix, as soon as possible I will continue to work on it.

.github/workflows/ci.yml
```diff
@@ -80,4 +80,4 @@ jobs:
             fail-fast: false
             matrix:
                 os: [ubuntu-latest, windows-latest, macos-latest]
-                php-version: ['8.1']
+                php-version: ['8.2']
```
jack-worman, line 83: Should still be testing on 8.1.
```suggestion
                php-version: ['8.1', '8.2']
```

jack-worman: And 8.2 should be marked as experimental

CallMeJon90:
I don't see the error in getopt.php:38
I need to understand why

Slamdunk:
Composer SAT resolver installed the only set where PHP 8.2 is allowed, which is a really old one where all packaged didn't properly declared the php requirement in their `composer.json` files.

Deps on PHP 8.2 build need to be installed with `--ignore-platform-req=php` flag I think.

Slamdunk:
I suggest to add `--ignore-platform-req=php` on both builds until all CI is green, and only then edit the `ci.yml` to have `--ignore-platform-req=php` on PHP 8.2 only.

Github Actions YAML files may be very time consuming for newcomers: let's focus on the code first.

phpseclib/Crypt/Common/SymmetricKey.php
```diff
@@ -263,6 +268,85 @@ abstract class SymmetricKey
...
+     * and the time it would be needed for it's
+     * initialization [by mcrypt_generic_init()]
+     * which, typically, depends on the complexity
+     * on its internaly Key-expanding algorithm.
+     *
+     * @see self::encrypt()
+     * @var int
+     */
+    protected $cfb_init_len = 600;
+
```
jack-worman, line 349: Merge error? These properties along with mcrypt were just removed.

phpseclib/Crypt/DH/PublicKey.php
```diff
@@ -25,6 +26,13 @@ class PublicKey extends DH
 {
     use Common\Traits\Fingerprint;
 
+    /**
+     * Public Key
+     *
+     * @var BigInteger
+     */
+    protected $publicKey;
```
jack-worman, line 34: Looks like this should go on `DH` instead

phpseclib/Crypt/DH/PublicKey.php
```diff
@@ -25,6 +26,13 @@ class PublicKey extends DH
 {
     use Common\Traits\Fingerprint;
 
+    /**
+     * Public Key
```
jack-worman, line 30: No useless comments please

phpseclib/Crypt/EC/BaseCurves/KoblitzPrime.php
```diff
@@ -40,6 +40,15 @@
  */
 class KoblitzPrime extends Prime
 {
+    /**
+     * @var array<array<string,BigInteger>>
```
jack-worman, line 44:
```suggestion
     * @var list<array{a: BigInteger, b: BigInteger}>
```

terrafrost (commented):
It's looking better!

The mcrypt stuff is weird but that's already been commented on.

Thanks for your work on this!

phpseclib/Crypt/Common/SymmetricKey.php
```diff
@@ -211,6 +211,11 @@ abstract class SymmetricKey
      */
     protected $key = false;
 
+    /**
+     * @var null|string
+     */
+    private $hKey = null;
```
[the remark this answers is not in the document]
Slamdunk: > I don't believe the `= null` is necessary

If `null` type is necessary for the variable, yes the initialization is necessary too: once typed attributes will land here, `private ?string $hKey;` cannot be checked without being initialized first, as it is currently used this way.

But as far as I can see, in the lazy-loading part an `isset()` check can be added first, so we can completely drop the `null` type requirement.

PR #1835, opened 2022-09-01
Throw \phpseclib3\Exception\ExceptionInterface exceptions

jack-worman:


terrafrost:
I'm gonna try to look at this as time permits. I'm vacationing in Spain right now (traveled here from Austin, TX, USA). I'll be in Europe the rest of this week and next.

If I can't get to it on this trip then when I get back.

Thanks!

phpseclib/File/ASN1.php
```diff
@@ -23,8 +23,9 @@
 
 namespace phpseclib3\File;
 
-use DateTime;
```
terrafrost, line 26: I opened up https://github.com/phpseclib/phpseclib/blob/master/phpseclib/File/ASN1.php and did Ctrl + F for Base64 and it doesn't look like it's actually used in that file?

phpseclib/Crypt/DH.php
```diff
@@ -84,7 +85,7 @@ public static function createParameters(...$args): Parameters
         $params = new Parameters();
         if (count($args) == 2 && $args[0] instanceof BigInteger && $args[1] instanceof BigInteger) {
             //if (!$args[0]->isPrime()) {
-            //    throw new \InvalidArgumentException('The first parameter should be a prime number');
+            //    throw new \phpseclib3\Exception\InvalidArgumentException('The first parameter should be a prime number');
```
terrafrost, line 88: Would prob be better to just remove the \ instead of adding the full namespacing.

terrafrost (commented):
Apologies for the delayed response! My vacay ended this past weekend and I've been trying to play catchup at work for most of this week.

Anyway, the code changes look great! There are two semi-nit-picky comments I made but those the only "issues" that I found!

terrafrost:
Merged. Thanks!

Issue #1847, opened 2022-09-28
delete of non-existent folder returns true when it used to return false


phil-davis:
This commit changed the behavior:
https://github.com/phpseclib/phpseclib/commit/1dfd1b9cd4cc0f8566ee12a18f8abb9657eebb18

That was "a good thing" for issue #1791 when the folder exists but there is difficulty getting the folder contents.

I noticed the behavior change in unit tests: https://github.com/owncloud/core/pull/40323#issuecomment-1254719428

The test does:
```
	public function testDirectories($directory) {
		$this->assertFalse($this->instance->file_exists('/' . $directory));

		$this->assertTrue($this->instance->mkdir('/' . $directory));

		$this->assertTrue($this->instance->file_exists('/' . $directory));
		$this->assertTrue($this->instance->is_dir('/' . $directory));
		$this->assertFalse($this->instance->is_file('/' . $directory));
		$this->assertEquals('dir', $this->instance->filetype('/' . $directory));
		$this->assertEquals(0, $this->instance->filesize('/' . $directory));
		$this->assertTrue($this->instance->isReadable('/' . $directory));
		$this->assertTrue($this->instance->isUpdatable('/' . $directory));

		$dh = $this->instance->opendir('/');
		$content = [];
		while ($file = \readdir($dh)) {
			if ($file != '.' and $file != '..') {
				$content[] = $file;
			}
		}
		$this->assertEquals([$directory], $content);

		$this->assertFalse($this->instance->mkdir('/' . $directory)); //can't create existing folders
		$this->assertTrue($this->instance->rmdir('/' . $directory));

		$this->wait();
		$this->assertFalse($this->instance->file_exists('/' . $directory));

		$this->assertFalse($this->instance->rmdir('/' . $directory)); //can't remove non existing folders

		$dh = $this->instance->opendir('/');
		$content = [];
		while ($file = \readdir($dh)) {
			if ($file != '.' and $file != '..') {
				$content[] = $file;
			}
		}
		$this->assertEquals([], $content);
	}
```

and the existing test fails at:
```
		$this->assertFalse($this->instance->rmdir('/' . $directory)); //can't remove non existing folders
```

The `rmdir` in that class does:
```
			$result = $this->getConnection()->delete($this->absPath($path), true);
```
and returns the result.

With the fix in phpseclib 3.0.15 I now get back `true` for an attempt to delete a folder that does not even exist.

[...]

PR #1848, opened 2022-09-28
Return false when attempting to delete a non-existent directory

phil-davis:
1) Add test assertion that delete on a non-existent directory returns false
2) Adjust testDeleteEmptyDir so that Psalm understands what `$sftp` is
3) Adjust real code so that SFTP `readList` can return a status (if it knows one) rather than always returning `false` when something unexpected happens. Enhance `delete_recursive` so that it detects `StatusCode::NO_SUCH_FILE` and explicitly returns `false` in that case. That will put back the previous behavior - a delete call on a non-existent folder returned `false`

Fixes issue #1847

### DIFF
--- a/phpseclib/Net/SFTP.php
+++ b/phpseclib/Net/SFTP.php
@@ -816,6 +816,12 @@ class SFTP extends SSH2
     {
         $files = $this->readlist($dir, false);
 
+        // If we get an int back, then that is an "unexpected" status.
+        // We do not have a file list, so return false.
+        if (is_int($files)) {
+            return false;
+        }
+
         if (!$recursive || $files === false) {
             return $files;
         }
@@ -846,6 +852,13 @@ class SFTP extends SSH2
     public function rawlist(string $dir = '.', bool $recursive = false)
     {
         $files = $this->readlist($dir, true);
+
+        // If we get an int back, then that is an "unexpected" status.
+        // We do not have a file list, so return false.
+        if (is_int($files)) {
+            return false;
+        }
+
         if (!$recursive || $files === false) {
             return $files;
         }
@@ -882,7 +895,7 @@ class SFTP extends SSH2
     /**
      * Reads a list, be it detailed or not, of files in the given directory
      *
-     * @return array|false
+     * @return array|int|false array of files, integer status (if known) or false if something else is wrong
      * @throws UnexpectedValueException on receipt of unexpected packets
      */
     private function readlist(string $dir, bool $raw = true)
@@ -909,8 +922,9 @@ class SFTP extends SSH2
                 break;
             case SFTPPacketType::STATUS:
                 // presumably SSH_FX_NO_SUCH_FILE or SSH_FX_PERMISSION_DENIED
-                $this->logError($response);
-                return false;
+                [$status] = Strings::unpackSSH2('N', $response);
+                $this->logError($response, $status);
+                return $status;
             default:
                 throw new UnexpectedValueException('Expected PacketType::HANDLE or PacketType::STATUS. '
                                                   . 'Got packet type: ' . $this->packet_type);
@@ -963,7 +977,7 @@ class SFTP extends SSH2
                     [$status] = Strings::unpackSSH2('N', $response);
                     if ($status != StatusCode::EOF) {
                         $this->logError($response, $status);
-                        return false;
+                        return $status;
                     }
                     break 2;
                 default:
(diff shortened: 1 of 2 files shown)


phil-davis:
How do I get CI to run?
I would like to see this test assertion fail, and then push some code to make it pass.

phil-davis:
Note: Psalm complained about the test code addition that I did in the first commit:
```
ERROR: MixedMethodCall - ../tests/Functional/Net/SFTPUserStoryTest.php:642:20 - Cannot determine the type of $sftp when calling method delete (see https://psalm.dev/015)
            $sftp->delete(self::$scratchDir),

  The type of delete is sourced from here - ../tests/Functional/Net/SFTPUserStoryTest.php:621:40
    public function testDeleteEmptyDir($sftp)
```

I do not understand why it started complaining. I think that it should have already been complaining about the same thing in the existing almost-identical code a few lines above that point.

Anyway, after adding an `assert()` and PHPdoc for `scratchDir`, Psalm is happy.

jack-worman:
You can run the quality tools by yourself by installing them via "composer global". If you then put you global composer bin directory in your $PATH, you can use the convenient scripts defined in composer.json.

> I do not understand why it started complaining

That error was pre-existing it has just been baselined for now. We are slowly chipping away at those.

phil-davis:
> You can run the quality tools by yourself by installing them via "composer global". If you then put you global composer bin directory in your $PATH, you can use the convenient scripts defined in composer.json.

It looked like I would have to mess about a bit to set up the SFTP server to run SFTP unit tests.

I did run the `psalm` and code-style tests easily locally.

phil-davis:
I need to run Psalm again locally and fix the issue that I introduced...

tests/Functional/Net/SFTPUserStoryTest.php
```diff
@@ -620,6 +623,7 @@ public function testRmDirScratchNonexistent($sftp)
      */
     public function testDeleteEmptyDir($sftp)
     {
+        assert($sftp instanceof SFTP);
```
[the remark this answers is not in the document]
phil-davis: Fixed

phil-davis:
> I need to run Psalm again locally and fix the issue that I introduced...

Done - type declarations are sorted out.

Please run CI and review.

phil-davis:
Ready for review and comments please.

phpseclib/Net/SFTP.php
```diff
@@ -2310,9 +2324,14 @@ private function delete_recursive(string $path, int &$i): bool
...
-        // permissions didn't allow reading
-        if (empty($entries)) {
+        // The folder does not exist at all, so we cannot delete it.
+        if ($entries === StatusCode::NO_SUCH_FILE) {
+            return false;
+        }
+
+        // Normally $entries would have at least . and .. but it might not if the directories
+        // permissions didn't allow reading. If this happens then default to an empty list of files.
+        if (($entries === false) || is_int($entries)) {
```
terrafrost, line 2334: On line 1565 it's `$entries === false || is_int($entries)` but on this one it's `($entries === false) || is_int($entries)`. They prob ought to be consistent.

Anyway, let me backport this to 1.0 / 2.0 / 3.0 and I can change this while I'm at it.

Not sure if I'll be able to do this today (got some stuff going on in < 1h) but I'll see what I can do!

phil-davis: I made line 1565 consistent. Personally I like to have "more" brackets, so that there is no doubt about order of evaluation of stuff. But I don't know what the project preference/policy is.

phil-davis: Let me know if I should make PRs to the old branches.

terrafrost: > I made line 1565 consistent. Personally I like to have "more" brackets, so that there is no doubt about order of evaluation of stuff. But I don't know what the project preference/policy is.

It'd be more consistent with phpseclib's current coding standards to do `$entries === false || is_int($entries)`. A few examples:

https://github.com/phpseclib/phpseclib/blob/master/phpseclib/Crypt/Common/SymmetricKey.php#L738
https://github.com/phpseclib/phpseclib/blob/master/phpseclib/Crypt/Common/SymmetricKey.php#L811
etc

If we were going to go with more parenthesis then ideally phpseclib would be updated in all the other places where it doesn't do that currently, as well, which would be quite a bit more work.

> Let me know if I should make PRs to the old branches.

Yah - that'd be awesome - thanks!!

phil-davis:
Backport to 3.0 in PR #1849 
I will make the others in a few hours - going out now.

terrafrost:
I squashed all the commits for #1848 into a single commit and then cherry-picked that to the 1.0 branch and merged that into 2.0 / 3.0 / master.

Here's the commit:

https://github.com/phpseclib/phpseclib/commit/24845ccbcc3039a292bfd1ad3d52dd5d8dc6b304

Thanks!

PR #1850, opened 2022-10-03
[3.0] Return false when attempting to delete a non-existent directory

phil-davis:
Backport #1848 

Note: I did not backport the commit for "Declare return type of readlst" - I don't think that will work on PHP 5.6?

### DIFF
--- a/phpseclib/Net/SFTP.php
+++ b/phpseclib/Net/SFTP.php
@@ -1003,6 +1003,12 @@ class SFTP extends SSH2
     {
         $files = $this->readlist($dir, false);
 
+        // If we get an int back, then that is an "unexpected" status.
+        // We do not have a file list, so return false.
+        if (is_int($files)) {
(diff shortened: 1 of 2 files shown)


phil-davis:
@jack-worman @terrafrost I think that I got the right things to backport, and the right language features for the older PHP versions. Please review.

When you are OK with this, then I can backport to 2.0 and 1.0 (I don't want to start on those if there is something still to adjust here)

phil-davis:
Hmmm - there is red CI. I guess there is a bunch of stuff that needs to be done a different way for older PHP versions...

I may not get to look at this for a while, and I am not already familiar with the past code structures etc. It is not just a matter of "git cherry-pick". So feel free to do the backports yourself. I thought that I could be helpful, but probably I will cause more delay due to back-and-forth comments as I learn how things should be done in each older version.

PR #1864, opened 2022-11-16
added error handling in case the server doesn't send any content

ppaulis:
Signed-off-by: Pascal Paulis <ppaulis@gmail.com>

### DIFF
--- a/phpseclib/Net/SCP.php
+++ b/phpseclib/Net/SCP.php
@@ -246,8 +246,12 @@ class SCP
 
         $content = '';
         while ($size < $info['size']) {
-            if (!$data = $this->_receive()) {
-                throw new \Exception('No data received from server');
+            $data = $this->_receive();
+
+            // Terminate the loop in case the server repeatedly sends an empty response
+            if (!$data) {
+                user_error('No data received from server', E_USER_NOTICE);
+                return false;
             }
 
             // SCP usually seems to split stuff out into 16k chunks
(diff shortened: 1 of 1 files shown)


phpseclib/Net/SCP.php
```diff
@@ -246,7 +246,10 @@ function get($remote_file, $local_file = false)
 
         $content = '';
         while ($size < $info['size']) {
-            $data = $this->_receive();
+            if (!$data = $this->_receive()) {
+                throw new \Exception('No data received from server');
```
terrafrost, line 250: phpseclib 2.0 didn't throw any exceptions (outside of bootstrap.php, which is a little different). The phpseclib 2.0 way of doing this is to do `user_error`.

phpseclib 3.0 throws exceptions but it doesn't throw generic exceptions - the closest thing to a generic exception that it throws is prob [RuntimeException](https://www.php.net/manual/en/class.runtimeexception.php).

Also, from a CS perspective, I'd prefer it if you did the `if` statement more like this instead:

```php
$data = $this->_receive();
if (!$data) {
    // do stuff
}
```

terrafrost:
I squashed your two commits into one and then cherry-picked that into the 1.0 branch, which I merged into 2.0 / 3.0 / master:

https://github.com/phpseclib/phpseclib/commit/68c3eef88f14f1b4002270cdf90aedf764e21577

Thanks!

PR #1869, opened 2022-11-24
Fix PHP 8.2 dynamic property warning in ASN1.php

come-nc:

PR #1870, opened 2022-12-03
Creation of dynamic property is deprecated

luigifab:
Without this change, with PHP 8.2 I have: `Deprecated functionality: Creation of dynamic property phpseclib3\Crypt\Blowfish::$mcrypt_polyfill_init is deprecated  in .../mcryptcompat/mcrypt.php on line 680` (with latest [mcryptcompat](https://github.com/phpseclib/mcrypt_compat) and OpenMage)

### DIFF
--- a/phpseclib/Crypt/Common/BlockCipher.php
+++ b/phpseclib/Crypt/Common/BlockCipher.php
@@ -23,4 +23,5 @@ namespace phpseclib3\Crypt\Common;
  */
 abstract class BlockCipher extends SymmetricKey
 {
+    public $mcrypt_polyfill_init;
 }


terrafrost:
I'm not putting a variable exclusively used in another library (mcrypt_compat) into phpseclib core. The fix for this needs to live in mcrypt_compat.

terrafrost:
I added a new method - `continuousBufferEnabled()`:

https://github.com/phpseclib/phpseclib/commit/16df0023256832643d75fe197bc93d80f82e84f6

It'll be that method that I'll call, instead of `$mcrypt_polyfill_init`, in mcrypt_compat. I'll do that after I [release a new version of phpseclib in a few days](https://github.com/phpseclib/phpseclib/pull/1869#issuecomment-1337604700).

terrafrost:
New versions of mcrypt_compat have been released that should fix this!

I wound up not using continuousBufferEnabled() as that apparently doesn't work on ECB mode so I wound up just using reflection. Had I known I'd be going that route I could have done a release sooner but whatever. It's done now!

PR #1876, opened 2023-01-04
fix "Creating default object from empty value" error

Kefkef123:
This seems like a typo that causes a "Creating default object from empty value" warning on my side, please let me know if anything is wrong


terrafrost:
I cherry-picked this into the 3.0 branch and then merged that into master:

https://github.com/phpseclib/phpseclib/commit/3ad3693d31d6685d590ae67da0367b3c0976407e

Thanks!

PR #1878, opened 2023-02-02
Add test for inheritance

kylekatarnls:
Test for #1879


terrafrost:
Closing since, per the comments in #1879, this is not the direction I ultimately went.

PR #1881, opened 2023-02-06
Added function to retrieve a certificate's serial number

maehdros:
Added a small accessor to retrieve a certificate Serial Number

### DIFF
--- a/phpseclib/File/X509.php
+++ b/phpseclib/File/X509.php
@@ -2949,6 +2949,16 @@ class X509
     }
 
     /**
+     * Get the Serial Number
+     *
+     * @return BigInteger
+     */
+    public function getSerialNumber():BigInteger
+    {
+        return $this->serialNumber;
+    }
+
+    /**
      * Turns the certificate into a certificate authority
      */
     public function makeCA(): void
--- a/phpseclib/File/X509.php
+++ b/phpseclib/File/X509.php
@@ -2951,11 +2951,16 @@ class X509
     /**
      * Get the Serial Number
      *
-     * @return BigInteger
+     * @return BigInteger|null
      */
-    public function getSerialNumber():BigInteger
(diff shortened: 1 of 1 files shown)


terrafrost:
What's the use case for this? Like other than being more succinct why is this better than simply doing `(new X509())->load('...')['tbsCertificate']['serialNumber']`?

As currently designed, array keys are the primary way you're supposed to access stuff. The main exceptions to this are `getDN()`, `getIssuerDN()`, `getDNProp()`, `getIssuerDNProp()` and `getPublicKey()`.

In the case of the DN stuff I feel like it makes sense because DNs can come in a myriad of formats. You might want the phpseclib internal representation (rdnSequence[0][0]['type'] == 'id-at-countryName'), an OpenSSL style interpretation (eg. C, ST, O, OU, etc), a string, etc. And how you want to query the various parts of the DN might vary as well.

In the case of getPublicKey(), calling that returns you an actual instance of the public key, that you can use, without any further handling, to do signature verification. The alternative is to do `$cert['tbsCertificate']['subjectPublicKeyInfo']['subjectPublicKey']` but then that'd return a string vs an object that you can actually use.

But in this case there's no real difference between accessing it through array keys vs accessing it through the new function you're added. Consequently, I'm disinclined to add this. Like if we go this route do we start adding methods for stuff like `getStartDate()` and `getEndDate()`?

Also, this only works on X509 certs when the X509 object can also hold CSR's and CRL's (phpseclib 4.0 will change this). `getPublicKey()` can be used with all of them. `getSubjectDN` can be used with X509 certs and CSR's and `getIssuerDN` an be used with X509 certs and CRL's.

Maybe in the 4.0 branch, after I've split X509 out into CRL.php and CSR.php and SPKAC.php I might be more open to adding this but, for the time being, I'm not.

PR #1888, opened 2023-02-28
Better support for multiple interactive channels & expose shell functions

rposky:
I set out with the simple intent to make the initShell() function public and to add isShellOpen(), with the goal of enabling integration patterns in which callers can better govern when shell initialization occurs and/or detect server-side channel closure. Presently, the shell is implicitly opened upon read() / write(), and there is ambiguous visibility as to whether the channel remains open past the return of those functions, such that the caller cannot be sure whether a future call to read() / write() implies a new interactive shell environment.

In pursuing that goal, it became evident that other channel interactions affect internal state tracking for the interactive shell to the extent that further work was prompted to separate these functions, and to forestall a confusing experience with the added isShellOpen() for callers combining usage with other channels. Specifically, it is the case now that calling exec() or closing an interactive channel, i.e. through reset(), disablePTY(), and/or stopSubsystem(), internally invalidates the interactive shell, as does the handling of a server CHANNEL_CLOSE message. These behaviors indicate an expectation of at most a single open interactive channel at any given time, and hence the expansion of the goal into bettering support for multiple interactive channels (and it seemed to me that lower, protocol-level support for multiple channels was actually pretty good).

If multiple interactive channels can be simultaneously open, it follows that callers need be able to select among them when calling read() / write(). This is achieved through an added $channel argument to those functions, along with reset(), however with the current channel selection behavior intact. A novel error is introduced for callers specifying an un-open non-shell channel for read() / write(), but shell auto-initialization remains. This could perhaps be made more user-friendly in the future by auto-initializing the exec pty and/or a subsystem as well.

To determine the open state of an interactive channel, a check on the respective channel status has been introduced (expecting status to be DATA when open), and which allows for removal of the channel-specific state tracking flags $in_request_pty_exec and $in_subsystem. Tying the channel status to channel open-ness allows for prompt feedback upon channel closure and without the need to maintain channel-specific flags from unrelated paths. A bug is also corrected in read() in which the exec pty is marked closed upon timeout and is no longer closable via disablePTY(), despite the channel may actually remain open. A downside to this channel status approach is that buffered channel data remaining past a return from read() is no longer accessible on a closed channel. This is left unresolved, and the impact is presumed to be low given that the channel has closed unexpectedly in this case. We may however need to reset channel buffers upon initialization...

Another consequential change is in the removal of potential disconnect() during channel closure via close_channel(). The disconnect() relied on timeout mechanics to determine that a server had not responded to channel closure, but it did not account for the timeout state when entering the routine, such that a disconnect() was guaranteed when calling i.e. disablePTY() after a read() which had timed out, as well as when calling non-pty exec() which timed out. I considered resetting the timeout state during close_channel(), but this risks loss of visibility into the timeout state of earlier requests, and propagates use of a feature with stated scope over just exec() and read(), and which feature has also not been well-integrated into other public methods which prompt protocol and network activity. More importantly however, from the standpoint of channel closure, it seems sane enough to me to ensure we ourselves never write more data to a closed channel, and perhaps even that a channel number is not re-used unless a server has acknowledged closure, but that we don't need to go scorched Earth and disconnect() for a failure to acknowledge (or for a failure of network to deliver / client to detect).

I think that summarizes well-enough the changes in this proposal, but please let me know where I can clarify further, as necessary. Your feedback is much appreciated, and thank you for your consideration.

### DIFF
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -722,16 +722,6 @@ class SSH2
     private bool $request_pty = false;
 
     /**
-     * Flag set while exec() is running when using enablePTY()
-     */
-    private bool $in_request_pty_exec = false;
-
-    /**
-     * Flag set after startSubsystem() is called
-     */
-    private bool $in_subsystem = false;
-
-    /**
      * Contents of stdError
      */
     private string $stdErrorLog;
@@ -2423,7 +2413,7 @@ class SSH2
             return false;
         }
 
-        if ($this->in_request_pty_exec) {
+        if ($this->isPTYOpen()) {
             throw new RuntimeException('If you want to run multiple exec()\'s you will need to disable (and re-enable if appropriate) a PTY for each one.');
         }
 
@@ -2473,8 +2463,6 @@ class SSH2
                 $this->disconnect_helper(DisconnectReason::BY_APPLICATION);
                 throw new RuntimeException('Unable to request pseudo-terminal');
             }
-
-            $this->in_request_pty_exec = true;
         }
 
         // sending a pty-req SSH_MSG_CHANNEL_REQUEST message is unnecessary and, in fact, in most cases, slows things
@@ -2504,7 +2492,7 @@ class SSH2
 
         $this->channel_status[self::CHANNEL_EXEC] = MessageType::CHANNEL_DATA;
 
-        if ($this->in_request_pty_exec) {
+        if ($this->request_pty === true) {
             return true;
         }
 
@@ -2532,15 +2520,25 @@ class SSH2
     /**
      * Creates an interactive shell
      *
+     * Returns bool(true) if the shell was opened.
+     * Returns bool(false) if the shell was already open.
+     *
+     * @return bool
+     * @throws InsufficientSetupException if not authenticated
      * @throws UnexpectedValueException on receipt of unexpected packets
      * @throws RuntimeException on other errors
-     *@see self::read()
+     * @see self::IsShellOpen()
+     * @see self::read()
      * @see self::write()
      */
-    private function initShell(): bool
+    public function openShell(): bool
     {
-        if ($this->in_request_pty_exec === true) {
-            return true;
+        if ($this->isShellOpen()) {
+            return false;
+        }
+
+        if (!$this->isAuthenticated()) {
+            throw new InsufficientSetupException('Operation disallowed prior to login()');
         }
 
         $this->window_size_server_to_client[self::CHANNEL_SHELL] = $this->window_size;
@@ -2606,21 +2604,34 @@ class SSH2
     }
 
     /**
-     * Return the channel to be used with read() / write()
-     *
+     * Return the channel to be used with read(), write(), and reset(), if none were specified
+     * @deprecated for lack of transparency in intended channel target, to be potentially replaced
+     *             with method which guarantees open-ness of all yielded channels and throws
+     *             error for multiple open channels
      * @see self::read()
      * @see self::write()
      */
     private function get_interactive_channel(): int
     {
         switch (true) {
-            case $this->in_subsystem:
+            case $this->is_channel_status_data(self::CHANNEL_SUBSYSTEM):
                 return self::CHANNEL_SUBSYSTEM;
-            case $this->in_request_pty_exec:
+            case $this->is_channel_status_data(self::CHANNEL_EXEC):
                 return self::CHANNEL_EXEC;
             default:
                 return self::CHANNEL_SHELL;
         }
+    }
+
+    /**
+     * Indicates the DATA status on the given channel
+     *
+     * @param int $channel The channel number to evaluate
+     * @return bool
+     */
+    private function is_channel_status_data(int $channel): bool
+    {
+        return isset($this->channel_status[$channel]) && $this->channel_status[$channel] == MessageType::CHANNEL_DATA;
     }
 
     /**
@@ -2677,24 +2688,37 @@ class SSH2
      * Returns when there's a match for $expect, which can take the form of a string literal or,
      * if $mode == self::READ_REGEX, a regular expression.
      *
+     * If not specifying a channel, an open interactive channel will be selected, or, if there are
+     * no open channels, an interactive shell will be created. If there are multiple open
+     * interactive channels, a legacy behavior will apply in which channel selection prioritizes
+     * an active subsystem, the exec pty, and, lastly, the shell. If using multiple interactive
+     * channels, callers are discouraged from relying on this legacy behavior and should specify
+     * the intended channel.
+     *
+     * @param string $expect
+     * @param int $mode One of the self::READ_* constants
+     * @param int|null $channel One of the self::CHANNEL_* constants
      * @return string|bool|null
      * @throws RuntimeException on connection error
+     * @throws InsufficientSetupException on unexpected channel status, possibly due to closure
      * @see self::write()
      */
-    public function read(string $expect = '', int $mode = self::READ_SIMPLE)
+    public function read(string $expect = '', int $mode = self::READ_SIMPLE, int $channel = null)
     {
         $this->curTimeout = $this->timeout;
         $this->is_timeout = false;
 
-        if (!$this->isAuthenticated()) {
-            throw new InsufficientSetupException('Operation disallowed prior to login()');
+        if ($channel === null) {
+            $channel = $this->get_interactive_channel();
         }
 
-        if (!($this->bitmap & self::MASK_SHELL) && !$this->initShell()) {
-            throw new RuntimeException('Unable to initiate an interactive shell session');
+        if (!$this->isInteractiveChannelOpen($channel)) {
+            if ($channel != self::CHANNEL_SHELL) {
+                throw new InsufficientSetupException('Data is not available on channel');
+            } elseif (!$this->openShell()) {
+                throw new RuntimeException('Unable to initiate an interactive shell session');
+            }
         }
-
-        $channel = $this->get_interactive_channel();
 
         if ($mode == self::READ_NEXT) {
             return $this->get_channel_packet($channel);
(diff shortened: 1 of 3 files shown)


terrafrost:
What was the motivation for removing the `$this->isAuthenticated()` check from `write()` and `read()`?

Also, I think it'd prob be worthwhile to throw an exception for servers that don't support multiple channels. Quoting the comments of tests/Functional/Net/SSH2Test.php's `testChannelDataAfterOpen()` method:

```
        // Ubuntu's OpenSSH from 5.8 to 6.9 didn't work with multiple channels. see
        // https://bugs.launchpad.net/ubuntu/+source/openssh/+bug/1334916 for more info.
        // https://lists.ubuntu.com/archives/oneiric-changes/2011-July/005772.html discusses
        // when consolekit was incorporated.
        // https://marc.info/?l=openssh-unix-dev&m=163409903417589&w=2 discusses some of the
        // issues with how Ubuntu incorporated consolekit
```
You can open up that file / method to see how I do the server identification string check.

That said, as they currently existed the changes do not appear to introduce any presumably meaningful BC breaks so I'd be okay with these being in the 3.0 branch. I mean, you do say "_A downside to this channel status approach is that buffered channel data remaining past a return from read() is no longer accessible on a closed channel. This is left unresolved, and the impact is presumed to be low given that the channel has closed unexpectedly in this case_" so I guess that'd be a technical BC break but idk if it'd be a meaningful one. To see how impactful an issue that is, a release of 3.0 could be done wherein this feature is undocumented. If phpseclib 3.0 gets lots of bug reports as a result of the feature I just revert it and then do a new release without the feature. It's a non-zero risk but I think it's a pretty small risk all the same. And, tbh, even fixes to legit bonafide bugs could be considered BC breaks. https://xkcd.com/1172/ kinda touches on this.

Anyway, pursuant to that, could you maybe backport them to that branch, create a PR and then merge that PR into the master branch, resolving the conflicts in the process, and create a PR on that, as well?

Two final comments:

1. I'm planning on releasing a new version of phpseclib 3.0.x this weekend so even if you create the PR for 3.0 before I do the new release I don't want to include this feature in this weekends release.

2. I am planning, at some point, to significantly expand phpseclib's multi channel support at some point. Like I'd make this a more expansive change then even your change. Like altho your change lets you have an interactive shell open whilst simultaneously having an `$ssh->exec()` instance with a PTY it doesn't let you (for example) have multiple interactive shells open. So here's an example of the API I've been thinking of employing:

   ```php
   $ssh = new SSH2();
   $ssh->login('...');

   $sh01 = $ssh->shell();
   $sh01->read('[prompt]');
   $sh01->write('...');

   $sh02 = $ssh->shell();
   $sh02->read('[prompt]');
   $sh02->write('...');
   ```

   These changes would also extend to SFTP. eg.
   
   ```php
   $remoteW = $sftp->open('filename.ext', 'w');
   $remoteR = $sftp->open('otherfile.ext', 'r');
   while (!feof($remoteR)) {
      fputs($remoteW, fread($remoteR, 1024);
   }
   ```
   
   In fact, one thing I'm kinda thinking this approach could enable is an experimental "super fast" single file upload. So like normally you send 1024 SSH_FXP_WRITE packets and then read that many SSH_FXP_STATUS packets back. But what if you didn't have to do that? What if you opened up a bunch of channels, sent 1024 SSH_FXP_WRITE packets for different offsets and then closed the channel rather than reading the SSH_FXP_STATUS packets? idk if that'd actually work but it's something that can't really be done until _full_ multi channel support has been implemented.
   
   That said, idk if this would go into phpseclib 4.0 or not. Right now, my main focus for phpseclib 4.0 has been to redo X509 / ASN1. Like _completely_ redo it, using lessons learned.

rposky:
Ah, good catch of removal on `$this->isAuthenticated()` from `write()` and `read()`. Initially I had presumed to relocate to `openShell()` and maintain current behavior, but I failed to revisit as scope of this issue expanded (as did my understanding of the code) to `exec pty` and `subsystem` channels. I will restore the check as it was, while also maintaining in`openShell()`.

I hesitate however to embed a server operating system check and disable the multiple channels feature for certain versions of Ubuntu. Following the [link](https://bugs.launchpad.net/ubuntu/+source/openssh/+bug/1334916) in the referenced unit test `testChannelDataAfterOpen()`, there is even a workaround cited, seemingly resolving the issue "if sshd is compiled without consolekit.patch". Even were it not the case that a workaround exists, I'd say let your consumers bear the burden of the specific client and server environments they operate in (they ultimately do anyhow), and how they integrate to the library as a result.

Thank you for your review and willingness to incorporate. I'll certainly put together the requested PRs, once we reach agreement on the proposed changes. I like the new interface idea. I did a little experimenting with creating a separate object of an SSH channel even, as there seemed to be enough complexity in the protocol to warrant a good encapsulation, but quickly ran into refactoring of `get_channel_packet()` and `get_binary_packet()` and backed away. I would probably have to first unroll all of the recursion in those functions to make good sense of current behavior, no easy task itself.

Point taken about that multi-channel interactions are still very limited, i.e. in no capability for multiple shells nor subsystems, etc. I see this issue as an incremental step in that direction, and that the added specification of a channel id to `read()`, `write()`, and `reset()` are useful to that end. However, the added reliance on `self::CHANNEL_*` constants would seem to be opposed to that direction, given that I imagine channel ids could not be fixed or even necessarily known ahead-of-time with true multiple channel support.

So considering that, perhaps it would be be best to lay down some groundwork now such that callers are not further integrating to those constants? An added interface `getChannelId()` could work, with current behavior of yielding the last channel opened, regardless of current channel status. I'd like to hear your thoughts on that.

rposky:
Oops, I told you wrong regarding the removal of `$this->isAuthenticated()` from `read() ` and `write()`. It actually has been mostly maintained in the form of `isInteractiveChannelOpen()`, which includes a check for `isConnected()` therein, but perhaps that should be querying `isAuthenticated()` instead. I'll go ahead with that change as well as the `getChannelId()` idea thrown out there while awaiting further feedback.

terrafrost:
So I'm currently doing SXSW Interactive and will try to review this in the next few days.

Thanks!

terrafrost:
But yah, pursuant to my 3.0 thoughts, maybe you could backport this to 3.0 and merge that into master and create a PR for the 3.0 and master branches? And maybe for the 3.0 branch squash it to a single commit so that if it is something I ultimately do decide then I can do so more easily?

Thanks!

rposky:
Thanks again for the feedback. I'll go ahead with the requested backport and pull requests.

> idk that I agree with this. Like right now, people running these problem versions of OpenSSH will experience a disconnect and won't be sure why they've experienced it and will post here asking about what happened.

> At minimum, what would be good is to make it so that if no response is received from the server when a new channel is requested (and if the server identification string matches) then instead of throwing the current error throw an error that says that their version of OpenSSH is bad and maybe include a link to https://bugs.launchpad.net/ubuntu/+source/openssh/+bug/1334916

While I don't mind the suggested error pivot as much, it still strikes me as inappropriate to embed in the utility. It's destined to become vestigial as systems and clients naturally evolve, and it requires work to discard, but in the meantime it will clutter.

Given that the issue does arise with some frequency, perhaps an FAQ would be helpful?

> With this approach I'm not sure getChannelId() is needed.

No, perhaps not, but I think it does grant more flexibility to future modifications than does having users directly integrate to the CHANNEL_* constants. And perhaps there is a future state where your Channel object idea is introduced and integrated to the existing APIs where it remains useful.

terrafrost:
> it still strikes me as inappropriate to embed in the utility. It's destined to become vestigial as systems and clients naturally evolve, and it requires work to discard, but in the meantime it will clutter.
>
> Given that the issue does arise with some frequency, perhaps an FAQ would be helpful?

I still want it in there and I'll add it myself if you won't.

I mean, using the vestigial argument, phpseclib should prob remove support for RSA keys as Ed25519 is the current best practices approach to public key auth in SSH2. TripleDES, DES, Blowfish, RC4, etc, ought to all be removed, with the only options being ChaCha20 and AES.

Using that argument, the `bad_algorithm_candidate` method ought to be removed from SSH2 and PuTTY prob ought to remove their [Bugs and More Bugs panel](https://the.earth.li/~sgtatham/putty/0.78/htmldoc/Chapter4.html#config-ssh-bugs).

> No, perhaps not, but I think it does grant more flexibility to future modifications than does having users directly integrate to the CHANNEL_* constants. And perhaps there is a future state where your Channel object idea is introduced and integrated to the existing APIs where it remains useful.

Well, as long as it isn't a BC break in the 3.0 branch, I'm good with it being in that branch! If it does BC break stuff than it should only live in the master branch.

rposky:
> PuTTY prob ought to remove their [Bugs and More Bugs panel](https://the.earth.li/~sgtatham/putty/0.78/htmldoc/Chapter4.html#config-ssh-bugs).

Ah yeah, an interesting parallel. Their configurable approach to smoothing over these issues is crucial, I think.

terrafrost:
I merged the other two PRs so imma close this one.

As for the exception I mentioned that I'll add in... to facilitate that I'm thinking I'll add a new method - `openChannel($channel)` - that'll do this:

```php
        $this->window_size_server_to_client[self::CHANNEL_SHELL] = $this->window_size;
        $packet_size = 0x4000;

        $packet = Strings::packSSH2(
            'CsN3',
            NET_SSH2_MSG_CHANNEL_OPEN,
            'session',
            self::CHANNEL_SHELL,
            $this->window_size_server_to_client[self::CHANNEL_SHELL],
            $packet_size
        );

        $this->send_binary_packet($packet);

        $this->channel_status[self::CHANNEL_SHELL] = NET_SSH2_MSG_CHANNEL_OPEN;
```
(replacing `self::CHANNEL_SHELL` with `$channel`)

I'll use this new function in `startSubsystem()`, `ping()`, `exec()`, `initShell()` and in SFTP.php and I'll add a counter to the function so that if the number of channels exceeds 1 and the server identification string matches that the exception I proposed will be thrown.

I'll need to install the problem version of OpenSSH to test it out but I'll try to do that at some point. If nothing else this also will de-duplicate some code, which is always a plus!

rposky:
Right on. I too welcome the code de-duplication. Feel free to @ me for a code review.

PR #1894, opened 2023-03-09
ASN1: 3.0 decodeBER() was optimized to remove duplicate work.

shetland:
PHP is already slow enough, why make it worse?

With this change, decodeBER() now runs twice as fast.

### DIFF
--- a/phpseclib/File/ASN1.php
+++ b/phpseclib/File/ASN1.php
@@ -206,7 +206,7 @@ abstract class ASN1
             return null;
         }
 
-        return [self::decode_ber($encoded)];
+        return [$decoded];
     }
 
     /**


shetland:
I noticed the same failure before I submitted this, but that is unrelated to the change being made here, and was already present in the master branch.

Happy to fix it as instructed by the tool if you'd like.

terrafrost:
So I'm currently doing SXSW Interactive and will try to review this in the next few days.

Thanks!

terrafrost:
Nice find! I cherry-picked your commit from the master branch and put into the 3.0 branch, which I then merged into master:

https://github.com/phpseclib/phpseclib/commit/abbc1ab7c794b6cbb7ddb2d94dfe85c02634f020

Thanks!

PR #1902, opened 2023-03-28
Fix SSH2::login phpdoc

### LINKED ISSUE #1627: PHPDoc / Type hints are wrong
I have found multiple instances where the documented type hints in the library are wrong or inaccurate.

* `\phpseclib3\Net\SSH2::exec()` has `$callback` which has a type hint of `Callback` (notice the captial 'C'), which lets readers to believe there is a class `\phpseclib3\Net\Callback`.

* `\phpseclib3\Net\SSH2::login()`  has `@param string[] ...$args`, but going from the documentation you are expecting instances of `\phpseclib3\Crypt\RSA\PrivateKey` there, too.

Those are just two examples that I came across recently. Version 2 was even trickier to get around :D 

Do you have considered using tools like `phpstan` to avoid such issues?
PHPStan is pretty simple to include and helps avoid A LOT! of those issues.

I think you could rectify some of the documentation issues in a bug fix release, since your code might work, but the type hints are not correct. Technically it is not a breaking change that way.

VincentLanglet:
Closes #1627

cc @terrafrost

### DIFF
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -2117,7 +2117,7 @@ class SSH2
      * The $password parameter can be a plaintext password, a \phpseclib3\Crypt\RSA|EC|DSA object, a \phpseclib3\System\SSH\Agent object or an array
      *
      * @param string $username
-     * @param string|AsymmetricKey|array[]|Agent|null ...$args
+     * @param string|PrivateKey|array[]|Agent|null ...$args
      * @return bool
      * @see self::_login()
      */
@@ -2142,7 +2142,7 @@ class SSH2
      * Login Helper
      *
      * @param string $username
-     * @param string ...$args
+     * @param string|PrivateKey|array[]|Agent|null ...$args
      * @return bool
      * @see self::_login_helper()
      */


VincentLanglet:
Thanks a lot @terrafrost, do you mean doing a patch-release ? (Or is there any patch release planned soon ?)

VincentLanglet:
Thanks a lot @terrafrost, do you mind doing a patch-release ? (Or is there any patch release planned soon ?)

terrafrost:
@VincentLanglet - I just released new versions of 2.0 and 3.0.

Thanks!

VincentLanglet:
Hi @MofX, either the phpdoc must be updated to `PrivateKey|AsymmetricKey` either the doc of this lib need to be updated.

I'm not sure if all the `AsymmetricKey` is really meant to be supported (or only some of them)

MofX:
I don't know that either. I am not really familiar with the library only using it for a tiny project, that happens to use psalm for static type checking and I had to update phpseclib and that broke this check.
I'll leave fixing this up to the library experts :)

PR #1904, opened 2023-03-31
(3.0 branch) SSH/SFTP: create new openChannel() method to eliminate dupe code

terrafrost:
@rposky - since you offered to code review in #1888 I'm @'ing you!


rposky:
Thanks! I reviewed #1905 which applies here as well.

Issue #1440, opened 2019-12-11
setTimeout doesn't work with the sleep command


neskhodovskiy:
Hello, thank you for the wonderful library and all the work you do with phpseclib. 

I am curious about the following case.

When I'm setting timeout for some command like 'ping' it works as expected.

When I'm setting timeout and call 'sleep', command waits the exact number of seconds passed to sleep, instead of respecting the timeout set to SSH2.

The code:

```php
<?php
use phpseclib\Crypt\RSA;
use phpseclib\Net\SSH2;

$key = new RSA();
$key->loadKey(file_get_contents('private-key.txt'));

$ssh = new SSH2('127.0.0.1');
if (!$ssh->login('username', $key)) {
    exit('Login Failed');
}

$ssh->setTimeout(3);

$start = microtime(true);
$ssh->exec('ping google.com');
$elapsed = microtime(true) - $start;

if ($ssh->isTimeout()) {
   echo sprintf('Command timed out in %01.2f seconds', $elapsed);
} else {
   echo sprintf('Command completed in %01.2f seconds', $elapsed);
}
```

The code above returns as expected:

> Command timed out in 3.12 seconds

However:

```php
<?php
use phpseclib\Crypt\RSA;
use phpseclib\Net\SSH2;

$key = new RSA();
$key->loadKey(file_get_contents('private-key.txt'));

$ssh = new SSH2('127.0.0.1');
if (!$ssh->login('username', $key)) {
    exit('Login Failed');
}

$ssh->setTimeout(3);

$start = microtime(true);
$ssh->exec('sleep 10');
$elapsed = microtime(true) - $start;

if ($ssh->isTimeout()) {
   echo sprintf('Command timed out in %01.2f seconds', $elapsed);
} else {
   echo sprintf('Command completed in %01.2f seconds', $elapsed);
}
```
I am expecting the command to time out after 3 seconds. But the second code runs for 10 seconds instead of 3 and outputs

> Command timed out in 10.11 seconds

What am I missing here?

terrafrost:
I'm able to reproduce the issue.

From the logs:

```
-> NET_SSH2_MSG_CHANNEL_REQUEST (since last: 0.0004, network: 0.0001s)
00000000  00:00:00:00:00:00:00:04:65:78:65:63:01:00:00:00  ........exec....
00000010  08:73:6c:65:65:70:20:31:30                       .sleep 10

[...]

PR #1905, opened 2023-03-31
(master branch) SSH/SFTP: create new openChannel() method to eliminate dupe code

terrafrost:
@rposky - since you offered to code review in #1888 I'm @'ing you!

### DIFF
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -923,6 +923,21 @@ class SSH2
     private bool $smartMFA = true;
 
     /**
+     * How many channels are currently opened
+     *
+     * @var int
+     */
+    private $channelCount = 0;
+
+    /**
+     * Does the server support multiple channels? If not then error out
+     * when multiple channels are attempted to be opened
+     *
+     * @var bool
+     */
+    private $errorOnMultipleChannels;
+
+    /**
      * Default Constructor.
      *
      * $host can either be a string, representing the host, or a stream resource.
@@ -1122,6 +1137,18 @@ class SSH2
             $this->bitmap = 0;
             throw new UnableToConnectException("Cannot connect to SSH $matches[3] servers");
         }
+
+        // Ubuntu's OpenSSH from 5.8 to 6.9 didn't work with multiple channels. see
+        // https://bugs.launchpad.net/ubuntu/+source/openssh/+bug/1334916 for more info.
+        // https://lists.ubuntu.com/archives/oneiric-changes/2011-July/005772.html discusses
+        // when consolekit was incorporated.
+        // https://marc.info/?l=openssh-unix-dev&m=163409903417589&w=2 discusses some of the
+        // issues with how Ubuntu incorporated consolekit
+        $pattern = '#^SSH-2\.0-OpenSSH_([\d.]+)[^ ]* Ubuntu-.*$#';
+        $match = preg_match($pattern, $this->server_identifier, $matches);
+        $match = $match && version_compare('5.8', $matches[1], '<=');
+        $match = $match && version_compare('6.9', $matches[1], '>=');
+        $this->errorOnMultipleChannels = $match;
 
         if (!$this->send_id_string_first) {
             fwrite($this->fsock, $this->identifier . "\r\n");
@@ -2432,28 +2459,7 @@ class SSH2
             throw new RuntimeException('If you want to run multiple exec()\'s you will need to disable (and re-enable if appropriate) a PTY for each one.');
         }
 
-        // RFC4254 defines the (client) window size as "bytes the other party can send before it must wait for the window to
-        // be adjusted".  0x7FFFFFFF is, at 2GB, the max size.  technically, it should probably be decremented, but,
-        // honestly, if you're transferring more than 2GB, you probably shouldn't be using phpseclib, anyway.
-        // see http://tools.ietf.org/html/rfc4254#section-5.2 for more info
-        $this->window_size_server_to_client[self::CHANNEL_EXEC] = $this->window_size;
-        // 0x8000 is the maximum max packet size, per http://tools.ietf.org/html/rfc4253#section-6.1, although since PuTTy
-        // uses 0x4000, that's what will be used here, as well.
-        $packet_size = 0x4000;
-
-        $packet = Strings::packSSH2(
-            'CsN3',
-            MessageType::CHANNEL_OPEN,
-            'session',
-            self::CHANNEL_EXEC,
-            $this->window_size_server_to_client[self::CHANNEL_EXEC],
-            $packet_size
-        );
-        $this->send_binary_packet($packet);
-
-        $this->channel_status[self::CHANNEL_EXEC] = MessageType::CHANNEL_OPEN;
-
-        $this->get_channel_packet(self::CHANNEL_EXEC);
+        $this->openChannel(self::CHANNEL_EXEC);
 
         if ($this->request_pty === true) {
             $terminal_modes = pack('C', TerminalMode::TTY_OP_END);
@@ -2534,6 +2540,42 @@ class SSH2
     }
 
     /**
+     * Opens a channel
+     */
+    protected function openChannel(string $channel, bool $skip_extended = false): bool
+    {
+        $this->channelCount++;
+
+        if ($this->channelCount > 1 && $this->errorOnMultipleChannels) {
+            throw new \RuntimeException("Ubuntu's OpenSSH from 5.8 to 6.9 doesn't work with multiple channels");
+        }
+
+        // RFC4254 defines the (client) window size as "bytes the other party can send before it must wait for the window to
+        // be adjusted".  0x7FFFFFFF is, at 2GB, the max size.  technically, it should probably be decremented, but,
+        // honestly, if you're transferring more than 2GB, you probably shouldn't be using phpseclib, anyway.
+        // see http://tools.ietf.org/html/rfc4254#section-5.2 for more info
+        $this->window_size_server_to_client[$channel] = $this->window_size;
+        // 0x8000 is the maximum max packet size, per http://tools.ietf.org/html/rfc4253#section-6.1, although since PuTTy
+        // uses 0x4000, that's what will be used here, as well.
+        $packet_size = 0x4000;
+
+        $packet = Strings::packSSH2(
+            'CsN3',
+            SSH2MessageType::CHANNEL_OPEN,
+            'session',
+            $channel,
+            $this->window_size_server_to_client[$channel],
+            $packet_size
+        );
+
+        $this->send_binary_packet($packet);
+
+        $this->channel_status[$channel] = SSH2MessageType::CHANNEL_OPEN;
+
+        return $this->get_channel_packet($channel, $skip_extended);
+    }
+
+    /**
      * Creates an interactive shell
      *
      * Returns bool(true) if the shell was opened.
@@ -2556,23 +2598,7 @@ class SSH2
             throw new InsufficientSetupException('Operation disallowed prior to login()');
         }
 
-        $this->window_size_server_to_client[self::CHANNEL_SHELL] = $this->window_size;
-        $packet_size = 0x4000;
-
-        $packet = Strings::packSSH2(
-            'CsN3',
-            MessageType::CHANNEL_OPEN,
-            'session',
-            self::CHANNEL_SHELL,
-            $this->window_size_server_to_client[self::CHANNEL_SHELL],
-            $packet_size
-        );
-
-        $this->send_binary_packet($packet);
-
-        $this->channel_status[self::CHANNEL_SHELL] = MessageType::CHANNEL_OPEN;
-
-        $this->get_channel_packet(self::CHANNEL_SHELL);
+        $this->openChannel(self::CHANNEL_SHELL);
 
         $terminal_modes = pack('C', TerminalMode::TTY_OP_END);
         $packet = Strings::packSSH2(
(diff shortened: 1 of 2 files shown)


phpseclib/Net/SSH2.php
```diff
@@ -2533,6 +2539,42 @@ public function exec(string $command, callable $callback = null)
...
 
+    /**
+     * Opens a channel
+     */
+    protected function openChannel(string $channel, bool $skip_extended = false): bool
+    {
+        $this->channelCount++;
+
+        if ($this->channelCount > 1 && $this->errorOnMultipleChannels) {
+            throw new \RuntimeException("Ubuntu's OpenSSH from 5.8 to 6.9 doesn't work with multiple channels");
```
rposky, line 2550: A little awkward to presume that errorOnMultipleChannels=true due to OpenSSH version, in consideration of the flag's generic name and possible future expansion of its use. I'd error with the more generic e.g. "Multiple channel feature has been disabled" or something to that effect.

phpseclib/Net/SSH2.php
```diff
@@ -2977,23 +2988,8 @@ public function ping(): bool
...
-
         try {
-            $this->send_binary_packet($packet);
-
-            $this->channel_status[self::CHANNEL_KEEP_ALIVE] = MessageType::CHANNEL_OPEN;
-
-            $response = $this->get_channel_packet(self::CHANNEL_KEEP_ALIVE);
+            $this->openChannel(self::CHANNEL_KEEP_ALIVE);
         } catch (\RuntimeException $e) {
             return $this->reconnect();
```
rposky, line 2994: The reconnect() is probably not desirable in the case that multiple channels have been disabled.

Perhaps more concerning is that with the current changes, ping() can be invoked at most 1 time when multiple channels are disabled, as subsequent pings will throw the multi channel exception.

phpseclib/Net/SSH2.php
```diff
@@ -922,6 +922,21 @@ class SSH2
...
+     */
+    private $channelCount = 0;
+
+    /**
+     * Does the server support multiple channels? If not then error out
+     * when multiple channels are attempted to be opened
+     *
+     * @var bool
+     */
+    private $errorOnMultipleChannels;
```
rposky, line 938: Consider exposing a setter for this as well, such that callers can enable the feature or forcibly disable if they know the OpenSSH version match to be a non-issue due to applied workarounds.

phpseclib/Net/SSH2.php
```diff
@@ -2533,6 +2539,42 @@ public function exec(string $command, callable $callback = null)
         }
     }
 
+    /**
+     * Opens a channel
+     */
+    protected function openChannel(string $channel, bool $skip_extended = false): bool
+    {
+        $this->channelCount++;
```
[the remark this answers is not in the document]
rposky: Also, if meant to track number of open channels, I do not see it decremented ever.

phpseclib/Net/SSH2.php
```diff
@@ -1123,6 +1138,18 @@ private function connect()
...
+        // https://bugs.launchpad.net/ubuntu/+source/openssh/+bug/1334916 for more info.
+        // https://lists.ubuntu.com/archives/oneiric-changes/2011-July/005772.html discusses
+        // when consolekit was incorporated.
+        // https://marc.info/?l=openssh-unix-dev&m=163409903417589&w=2 discusses some of the
+        // issues with how Ubuntu incorporated consolekit
+        $pattern = '#^SSH-2\.0-OpenSSH_([\d.]+)[^ ]* Ubuntu-.*$#';
+        $match = preg_match($pattern, $this->server_identifier, $matches);
+        $match = $match && version_compare('5.8', $matches[1], '<=');
+        $match = $match && version_compare('6.9', $matches[1], '>=');
+        $this->errorOnMultipleChannels = $match;
```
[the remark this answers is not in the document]
terrafrost: > What's the nature of the OpenSSH bug? Is it that no more than 1 channel can ever be established, or that multiple channels cannot co-exist?

Affected OpenSSH servers will kill the connection when you try to open up another channel. See https://marc.info/?l=openssh-unix-dev&m=163409903417589&w=2 for more detail.

phpseclib/Net/SSH2.php
```diff
@@ -2533,6 +2539,42 @@ public function exec(string $command, callable $callback = null)
         }
     }
 
+    /**
+     * Opens a channel
+     */
+    protected function openChannel(string $channel, bool $skip_extended = false): bool
```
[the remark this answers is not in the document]
terrafrost: Good catch!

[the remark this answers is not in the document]
terrafrost: Good idea.

As doing this kinda makes lines like this that proceed `$this->openChannel()` redundant I've removed them:

```php
        if ($this->isShellOpen()) {
            return false;
        }
```

phpseclib/Net/SSH2.php
terrafrost: The more generic error message you propose doesn't really explain the situation either. Like anyone affected by the issue is liable to be like "well can you enable the feature?" as tho _I'm_ the reason it's not working when, in reality, it's not working because of a buggy modification that the Ubuntu devs made to their fork of OpenSSH.

Sure, the variable name could be made less generic but, then again, there is something to be said about not having a variable name that's not 1KB in size as well lol. Like maybe I could have it named `$this->errorOnMultipleChannelsBecauseUbuntuForkOfOpenSSH58-69IsBroken` but that introduces it's own can of worms with it being impractically long.

phpseclib/Net/SSH2.php
terrafrost: > Perhaps more concerning is that with the current changes, ping() can be invoked at most 1 time when multiple channels are disabled, as subsequent pings will throw the multi channel exception.

Decrementing `$this->channelCount` should resolve that.

terrafrost:
Thanks for the feedback! I didn't respond to it earlier, even in spite of responded to other support requests, because this one is a lot more technical than most support requests.

I'll try to get my local changes on git either tonight or tomorrow.

Thanks again!

phpseclib/Net/SSH2.php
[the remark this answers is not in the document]
rposky: You wouldn't necessarily have to `unset($this->channel_status[$channel])`. An alternative would be to inspect the actual status values and to make a determination from there, rather than simply tallying up the number of channel status entries.

phpseclib/Net/SSH2.php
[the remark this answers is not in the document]
rposky: > As doing this kinda makes lines like this that proceed $this->openChannel() redundant I've removed them:

        if ($this->isShellOpen()) {
            return false;
        }
That looks sane to me.

phpseclib/Net/SSH2.php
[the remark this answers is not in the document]
terrafrost: I guess you could do `count(array_filter($this->channel_status, function($val) { return $val != NET_SSH2_MSG_CHANNEL_CLOSE; }))` but, as a matter of personal preference, I prefer `$this->channelCount`.

If this was a PR being done by someone else and they did `array_filter` instead of what I'm doing I prob wouldn't comment on it.

terrafrost:
@rposly - I made a new commit if you want to take a look see!

Thanks!

phpseclib/Net/SSH2.php
```diff
@@ -2725,7 +2761,7 @@ public function read(string $expect = '', int $mode = self::READ_SIMPLE, int $ch
             $channel = $this->get_interactive_channel();
         }
 
-        if (!$this->isInteractiveChannelOpen($channel)) {
+        if (!$this->isInteractiveChannelOpen($channel) && empty($this->channel_buffers[$channel])) {
```
rposky, line 2764: I had previously considered this change to ignore the connectivity requirement if buffers were available but decided against it. The reason being is that, in the READ_SIMPLE mode, connectivity will be required regardless. Once the buffers are drained, we're still executing within the `while` loop below, and back to needing a connection when we invoke `get_channel_packet`, or hitting disconnect exception.

We could confine the buffer emptiness check to the READ_NEXT mode for now, or rework the `while` loop to stop once the buffers are empty (and the channel is closed), but as-is, I think we're just changing the exception bound to be thrown with this change.

phpseclib/Net/SSH2.php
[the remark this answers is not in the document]
rposky: Right, and it can work either way. My preference for `channel_status` is just in that it involves less state, and mgmt thereof, than the additional `channel_count`.

phpseclib/Net/SSH2.php
terrafrost: In the unit test I added (`testReadingOfClosedChannel()` to tests/Functional/Net/SSH2Test.php) it's doing READ_SIMPLE and it's able to read from a closed buffer without issue.

So in the `while (true)` loop in `read()` there's this:

```php
            $response = $this->get_channel_packet($channel);
            if ($response === true) {
                return Strings::shift($this->interactiveBuffer, strlen($this->interactiveBuffer));
            }
```
$response is true because of this (from `get_channel_packet`):

```php
                $response = $this->get_binary_packet(true);
                if ($response === true && $this->is_timeout) {
                    if ($client_channel == self::CHANNEL_EXEC && !$this->request_pty) {
                        $this->close_channel($client_channel);
                    }
                    return true;
                }
```
Now that said, it does get a little janky when `$ssh->setTimeout(0)` is set. Like it outputs some of the output of `$ssh->write("ping 127.0.0.2\n", SSH2::CHANNEL_SHELL);` when it shouldn't ever output anything as it should essentially be caught in an infinite loop. This is because of of https://github.com/phpseclib/phpseclib/issues/1440#issuecomment-635970968 . I could undo it but idk I think what we really need are per channel timeout's instead of a global timeout but I think that'd be best for another PR.

Could you give some sample code that demonstrates the problem you were concerned about?

rposky: @terrafrost Yes, I'll try to get you that sample code tomorrow (or let you know that I've been unable to reproduce). Sorry it's taking me a bit to get to it.

rposky: So here's a unit test I've come up to illustrate my point.
```
public function testReadFromChannelBufferWhenDisconnected(): void
    {
        $ssh = $this->getSSH2Login();
        $ssh->openShell();
        $sshChannel = $ssh->getInteractiveChannelId();
        // start a command that will run for a bit and then close the channel
        $ssh->write("for i in {1..2}; do echo \"sleep 2...\" && sleep 2; done; exit\n", $sshChannel);

        $ssh->setTimeout(5);
        $ssh->enablePTY();
        $ssh->exec('bash');
        $execChannel = $ssh->getInteractiveChannelId();
        $ssh->write('echo "hello"', $execChannel);
        $ssh->read('', SSH2::READ_SIMPLE, $execChannel);
        $ssh->disablePTY();
        // simulate server disconnect and generate error on read()
        $ssh->disconnect();

        // make channel_buffers public and uncomment
        // $this->assertNotEmpty($ssh->channel_buffers[$sshChannel], 'Buffered data on SSH channel was expected');
        $this->assertFalse($ssh->isInteractiveChannelOpen($sshChannel));

        $ssh->setTimeout(1);
        $readOutput = $ssh->read('', SSH2::READ_SIMPLE, $sshChannel);
        $this->assertStringContainsString('sleep 2...', $readOutput, 'Shell echo output was not found');
        $this->assertFalse($ssh->isInteractiveChannelOpen($sshChannel));
    }
```
Against master, this generates the error `phpseclib3\Exception\InsufficientSetupException : Operation disallowed prior to login()` whereas the error `phpseclib3\Exception\InvalidArgumentException : fsock is not a resource.` is generated by the proposed change to pivot in `read()` based on channel buffer data.

So in the case that the server has disconnected, we're still throwing an exception, and a less descriptive one at that with the change. However, that's not to suggest that `Operation disallowed prior to login()` is a good error for this case, and I actually strongly dislike that `read()` and `write()` will attempt to auto re-open a shell which has exited (and which can actually succeed if the connection is alive). I would rather receive the error `'Data is not available on channel'` like we do for other channel types in the case that the shell is no longer open, and which at least allows for acknowledgement that the prior session is gone. That change could be disruptive however, but perhaps it's not too bad so long as we maintain the shell auto-open (and not re-open) on `read()` and `write()`.

There is another interesting caveat to the proposed change, and which is highlighted by this test case as well. If removing the call to `$ssh->disconnect()`, the change does allow for continued read against the closed shell (as intended), albeit only if the connection remains alive, whereas master throws the error `phpseclib3\Exception\RuntimeException : Unable to fulfill channel request` in a failed attempt to re-open the shell (probably worthy of investigation as well, but I didn't pursue). I'm not sure how likely a scenario it is to maintain an active SSH connection with unexpected closure of the shell though, but it strikes me as rare enough that I wouldn't introduce general logic to support it.

So in summary, I think where I'm at is that I do like enabling access to the buffered channel data, and that `read()` is probably the appropriate place to grant it, but that we haven't fully arrived at that goal with the proposed change. Technically, the access is already there through `$mode=READ_NEXT`, though the shell re-open behavior does make that hard to work with.

terrafrost: https://github.com/terrafrost/phpseclib/commit/f664ccb521819c075f727ffa9953cb4e7050e2e9 should take care of this.

And I do agree that the current behavior is suboptimal but, well, we've already talked about my long term plans lol. But right now my more immediate goals are completely rewriting ASN.1 / X.509. Like with what I have right now processing of ASN.1 stuff is deferred (ie. done on demand). So like if you want to get the total count of how many certs are in a CRL you don't need to parse the whole CRL - you can only parse the relevant parts of it. On a CRL with 40,000 entries this knocked the time down from 4s to 0.2s for that specific operation. I got other changes implemented, as well, but discussing them is a bit off topic in a PR for SSH2 lol

terrafrost:
I've made updates based on your feedback!

Unless I hear back from you in the next few days I'll prob merge this in the next few days!

rposky:
Thank ya. Looks good to me.

PR #1909, opened 2023-04-18
(3.0 branch) update exception message

browner12:
`fputs()` can return `int|false`.  if it fails and `$sent` is `false`, then our exception message looks a little confusing:

> Only   of XXX bytes were sent

This change updates the message to be more descriptive if the `fputs()` fails.

> Unable to write XXX bytes


browner12:
Great, thanks! Do you guys handle moving this stuff to master, or should I make a separate PR to that branch?

terrafrost:
I merged it into master.

Thanks!

PR #1917, opened 2023-06-12
Fix E_DEPRECATED in PHP 8.2 because of creation of a dynamic property (in 2.0)

seeschloss:
This commit just adds a "$name" property to the Crypto/Base and Crypto/Hash classes to prevent a E_DEPRECATED error in Net/SSH2 which uses this property.


terrafrost:
Good catch. I made another commit:

https://github.com/phpseclib/phpseclib/commit/16c67f439980c2e60ddd8373ff76c2525e59e7c0

PR #1940, opened 2023-09-18
Fix Undefined index jit notice

pafernandez-oesia:
If "jit" is not defined in $status array, it generates an "Undefined index: jit" notice.


terrafrost:
I released a new version - 3.0.23 - that includes this.

Thanks!

pafernandez-oesia:
Great, thank you.

Issue #1941, opened 2023-09-18
Fatal Error: Upgrading phpseclib/phpseclib (3.0.21 => 3.0.22) 


parallels999:
Does anyone else have this problem?
I'm trying to update, and it seems like it's a breaking change
```
Fatal error: 
  Uncaught UnexpectedValueException: 
    JIT on Windows is not currently supported in \vendor\phpseclib\phpseclib\phpseclib\bootstrap.php:28
    Stack trace:
      #0 \vendor\composer\autoload_real.php(41): require()
      #1 \vendor\composer\autoload_real.php(45): {closure}('decc78cc4436b12...', 'C:\\laravel\\ssw\\...') 
      #2 \vendor\autoload.php(25): ComposerAutoloaderInite3b2492fbaae1a014a5832dc98cf3829::getLoader()
```
On **3.0.21** works, How should I proceed?

parallels999:
On `php.ini` I add
```
opcache.jit=off
```
It is working, but I don't know if it's the right answer

terrafrost:
It may technically be a BC breaking change but the problem is that, on Windows, it turns out that [JIT _is_ broken](https://github.com/php/php-src/issues/11917). Sure, I've only ever seen issues when the PHP64 engine is used but, the problem is, if there are issues when that engine is being used then I can't be sure that other problems aren't lurking beneath the surface just waiting for someone to find and I don't want people expecting me to fix issues when the issues have nothing to do with phpseclib anyway. I mean, I suppose I could write a new engine just for JIT enabled Windows PHP installs but without really having a good handle on what is causing the problem, anyway, (well, aside from the fact that I know the problem is caused by PHP lol), I can't guarantee that problems could still exist in whatever rewrite I do.

In spite of that, if you want to use JIT on Windows, you can do `define('PHPSECLIB_ALLOW_JIT', true);` at the top of your file.

parallels999:
Thanks for responding, `define('PHPSECLIB_ALLOW_JIT', true);` seems like a good option, but this is just a test environment, I can disable it, it doesn't matter

rossaddison:
Here is my stack trace: 

````
C:\wamp64\www\yii3-i-4>php ./vendor/bin/psalm
Uncaught UnexpectedValueException: JIT on Windows is not currently supported in C:\wamp64\www\yii3-i-4\vendor\phpseclib\phpseclib\phpseclib\bootstrap.php:28
Stack trace:
[...]

PR #1942, opened 2023-09-20
Move JIT check to BigInteger

danog:
This fixes static analysis with Psalm v5 (which enables JIT by default) of projects which require phpseclib as a dependency.

### DIFF
--- a/phpseclib/Math/BigInteger.php
+++ b/phpseclib/Math/BigInteger.php
@@ -136,16 +136,6 @@ class BigInteger implements \JsonSerializable
     private static function initialize_static_variables()
     {
         if (!isset(self::$mainEngine)) {
-            // see https://github.com/php/php-src/issues/11917
-            if (strtoupper(substr(PHP_OS, 0, 3)) === 'WIN' && function_exists('opcache_get_status') && !defined('PHPSECLIB_ALLOW_JIT')) {
-                $status = opcache_get_status();
-                if ($status && isset($status['jit']) && $status['jit']['enabled'] && $status['jit']['on']) {
-                    throw new UnexpectedValueException(
-                        'JIT on Windows is not currently supported'
-                    );
-                }
-            }
-
             $engines = [
                 ['GMP', ['DefaultEngine']],
                 ['PHP64', ['OpenSSL']],
@@ -154,13 +144,16 @@ class BigInteger implements \JsonSerializable
                 ['PHP64', ['DefaultEngine']],
                 ['PHP32', ['DefaultEngine']]
             ];
+
             foreach ($engines as $engine) {
                 try {
                     self::setEngine($engine[0], $engine[1]);
-                    break;
+                    return;
                 } catch (\Exception $e) {
                 }
             }
+
+            throw new UnexpectedValueException('No valid BigInteger found. This is only possible when JIT is enabled on Windows and neither the GMP or BCMath extensions are available so either disable JIT or install GMP / BCMath');
         }
     }
 
--- a/phpseclib/Math/BigInteger.php
+++ b/phpseclib/Math/BigInteger.php
@@ -29,6 +29,7 @@ namespace phpseclib3\Math;
 
 use phpseclib3\Exception\BadConfigurationException;
 use phpseclib3\Math\BigInteger\Engines\Engine;
+use UnexpectedValueException;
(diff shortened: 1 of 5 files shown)


terrafrost:
First, I apologize for the delay. I'm currently vacationing and am doing a bunch of outdoorsy stuff wherein I don't have access to my computer.

Anyway, I'm kinda confused by this. How does Psalm v5 enable JIT by default?

Using the install instructions at https://psalm.dev/docs/running_psalm/installation/ (either with Composer or the Phar) it'd use your existing PHP setup.

That said, I suppose that's somewhat immaterial to the merits of the proposed change. So in-so-far as those are concerned... I can see pros and cons of going with either approach.

**Pros of the current approach**
So I still stand by what I said in https://github.com/phpseclib/phpseclib/issues/1941#issuecomment-1724032646 - if JIT on Windows is going to behave unpredictably in one scenario I can't be sure that there aren't other scenarios in which where it'll behave just as unpredictably. Maybe once https://github.com/php/php-src/issues/11917 is fixed I could look at the fix and see if there specific conditions for the issue and then implement workarounds when those conditions are met but, as is, I don't have a good enough grasp on the underlying issue to be able to do that.

**Pros of your approach**
Thus far the only confirmed problem with JIT on Windows manifests itself in BigInteger when the PHP64 engine is used. It's possible that other issues that have defied diagnose (ie. maybe someone makes a bug report and then never follows up when I ask for more info) are attributable to this but I have no way of knowing that for sure. If I were able to identify other scenarios in which this problem occurs that might help the PHP devs hone in on and ultimately fix the underlying issue. Of course that does put more of the onus on me.

So in lieu of knowing how Psalm v5 enables JIT by default and in recognition of your own contributions to open source I guess what I'm thinking is this: move the check over to Math/BigInteger/Engines/PHP64.php and PHP32.php. Make `isValidEngine` return false if JIT is enabled on Windows. Don't throw an exception, don't check for `PHPSECLIB_ALLOW_JIT`, just return false. Or better yet, in PHP.php (which both PHP64.php and PHP32.php extend), add a new method - `testJITOnWindows()` - and then call that method from `isValidEngine()` in both PHP32.php and PHP64.php.

Then, in `initialize_static_variables()`, in Math/BigInteger.php, make it so that if no valid engine is found that an exception is thrown. eg. maybe do `$foundEngine = false;` before the `for` loop and set that to true after `self::setEngine($engine[0], $engine[1]);`. Then, after the for loop, if `!$foundEngine` then do `throw new UnexpectedValueException('No valid BigInteger found. This is only possible when JIT is enabled on Windows and neither the GMP or BCMath extensions are available so either disable JIT or install GMP / BCMath');` or some such.

The advantage of doing that vs what you're doing is that people can still use the GMP or BCMath engines on Windows even when JIT is enabled.

danog:
Hi, thank you for the review!

>Anyway, I'm kinda confused by this. How does Psalm v5 enable JIT by default?

Sorry, should've been more detailed in the description of the PR :)
Psalm v5 enables function JIT by default in order to improve performance, currently we have around a 20% improvement!

Using tracing JIT can probably raise number that even more but as you saw, tracing JIT is quite buggy right now, plus function JIT does not rely on runtime type information, so bugs are more easily reproducible than with tracing JIT.

>A Possible Middleground?

Thank you, implemented the changes you suggested: I did it without the additional `$foundEngine` boolean variable since just returning from the `initialize_static_variables` function on success seemed like the simplest solution to me, can add it back if you want to.

Also, I've kept the `PHPSECLIB_ALLOW_JIT`, mostly because I intend to submit another PR to php-src to add a nightly test suite for Windows as well, which will test phpseclib (and other popular FOSS libraries) with JIT to find bugs and regressions, and that flag will really come in useful.

richard67:
Will there be a release which includes this change? If I see right, the latest 3.0.23 doesn't include it.

terrafrost:
I'll try to do a release this week.

terrafrost:
3.0.33 has been released!

richard67:
Yes, I saw. Thanks a lot.

AltamashShaikh:
@terrafrost @danog The exception added with this PR seems to create issue for us, as we are using the [bcmath](https://github.com/phpseclib/bcmath_compat/) polyfill

PR #1949, opened 2023-10-05
Update Base.php

SneakBug8:
At some hosting providers php_uname('m') function returns NULL which causes fatal php error preventing further execution (the bit with `php_uname('m') & "\xDF\xDF\xDF"`).

Applied a fix that would use PHP_OS const in such cases instead.


terrafrost:
I went ahead and implemented a fix here:

https://github.com/phpseclib/phpseclib/commit/173bb38e67e30429edb69e21a8b16d8a0c86c355

Thanks!

PR #1953, opened 2023-10-18
access $disconnect_reasons using self instead of static

MarkLittlewood:
Fatal error: Uncaught Error: Cannot access property phpseclib3\Net\SFTP::$disconnect_reasons was being received when login failed. 
changing static:: to self:: fixed the issue

### DIFF
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -3728,7 +3728,7 @@ class SSH2
             case NET_SSH2_MSG_DISCONNECT:
                 Strings::shift($payload, 1);
                 list($reason_code, $message) = Strings::unpackSSH2('Ns', $payload);
-                $this->errors[] = 'SSH_MSG_DISCONNECT: ' . static::$disconnect_reasons[$reason_code] . "\r\n$message";
+                $this->errors[] = 'SSH_MSG_DISCONNECT: ' . self::$disconnect_reasons[$reason_code] . "\r\n$message";
                 $this->bitmap = 0;
                 return false;
             case NET_SSH2_MSG_IGNORE:


terrafrost:
I'm not able to reproduce the issue that this PR attempting to fix.

So you're modifying the filter() method of SSH2.php. I added the following right after the function definition (just before the switch statement) and then ran the PHP script:

```php
print_r(static::$disconnect_reasons); exit;
```
I got no errors, not even after calling `error_reporting(E_ALL);` before initializing the SFTP() object and I'm running PHP 8.2.9. I mean, `print_r(self::$disconnect_reasons)` works, as well, but I'd like to know what, specifically, is causing the issue for you.

And, to be fair, altho the static keyword enables [late static binding](https://www.php.net/manual/en/language.oop5.late-static-bindings.php) it's not of huge concern because `$disconnect_reasons ` is private in SSH2.php.

terrafrost:
I tried it with PHP 8.3.0-dev, as well, and am still not able to reproduce the issue you described...

In both cases I did `print_r(static::$disconnect_reasons)` and `print_r(self::$disconnect_reasons)` and both worked just as well as one other.

terrafrost:
Maybe you can post the output of `phpinfo()`?

MarkLittlewood:
Unfortunatley I am still stuck on php v7.3.33 - some organisations take a long time to move versions.
That may be why I am getting the issue and you are not.

terrafrost:
I was able to reproduce this on 7.3.33 and, in light of that, went ahead and merged this!

Even if PHP 7.3.33 is old I "advertise" phpseclib 3.0 as working on PHP 5.6+ and I'm committed to that! The master branch (which will eventually become v4.0) is a different story but you're not using that version!

MarkLittlewood:
Fantastic thank you

PR #1969, opened 2023-12-14
Update composer.json

Traineratwot:
You use `bcmod` in your code,  It is necessary to add a dependency on the corresponding php extension

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -58,6 +58,7 @@
     ],
     "require": {
         "php": ">=8.1",
+        "ext-bcmath": "*",
         "paragonie/constant_time_encoding": "^2"
     },
     "require-dev": {


terrafrost:
Consider this code:

```php
if (function_exists('bcmod')) {
    echo bcmod($x, $y);
} else {
   // pure-PHP implementation of bcmod
}
```
If you do something like that then bcmath is *not* a dependency. It's preferentially used if it's available but it is 110% not a hard requirement.

In phpseclib 3+ the bcmath function calls are in the BCMath BigInteger engine only.  A grep of the codebase for bcmod reaffirms this:

```
$ grep -r bcmod .
./Math/BigInteger/Engines/BCMath/Reductions/Barrett.php:            return bcmod($n, $m);
./Math/BigInteger/Engines/BCMath/Reductions/Barrett.php:            return bcmod($x, $n);
./Math/BigInteger/Engines/BCMath/Reductions/EvalBarrett.php:            $code = 'return bcmod($x, $n);';
./Math/BigInteger/Engines/BCMath.php:            $temp = bcmod($current, '16777216');
./Math/BigInteger/Engines/BCMath.php:        $remainder->value = bcmod($this->value, $y->value);
./Math/BigInteger/Engines/BCMath.php:            $result->value = bcmod($result->value, $result->bitmask->value);
./Math/BigInteger/Engines/BCMath.php:            $r = bcmod($this->value, (string)$prime);
./Math/BigInteger/Engines/BCMath.php:            bcmod($this->value, bcpow('2', $x + 1, 0)),
```
Math/BigInteger.php only uses the BCMath engine **_if_** the bcmath extension is available.  Note the --**_if_**-- part of what I said.

Sure, you can call Math/BigInteger/Engines/BCMath.php directly and kinda use that as a drop in replacement for Math/BigInteger.php. That's useful for unit testing purposes but that's not documented behavior. You shouldn't be doing that and if you are then that's on you.

If you can find a place where BCMath is used unconditionally lmk. The correct course of action in that scenario isn't to update composer.json - it would be to put that BCMath function call in an if statement or something.

jack-worman:
Should it be added to "suggest" then?

terrafrost:
If the GMP extension is available the BCMath extension isn't used because nothing can beat the GMP extension. The suggest that's already in place reflects that.

I mean I could still put it in the suggest section, I suppose, even tho if you've installed every extension in that section BCMath will never be used, anyway, but idk, that kinda feels weird lol.

But, that said, I guess the comments in the suggest could be updated, as well. eg. for ext-bcmath it could say "This is recommended for 32-bit PHP installs without the GMP extension installed" (64-bit PHP installs don't use bcmath because it turns out that the pure-PHP implementation is faster than BCMath for 64-bit PHP installs)

PR #1988, opened 2024-03-15
[PHP 8.4] Fixes for implicit nullability deprecation

Ayesh:
Fixes all issues that emit deprecation notices on PHP 8.4 for implicit nullable parameter type declarations.

See:
 - [RFC](https://wiki.php.net/rfc/deprecate-implicitly-nullable-types)
 - [PHP 8.4: Implicitly nullable parameter declarations deprecated](https://php.watch/versions/8.4/implicitly-marking-parameter-type-nullable-deprecated)

### DIFF
--- a/phpseclib/Crypt/Common/AsymmetricKey.php
+++ b/phpseclib/Crypt/Common/AsymmetricKey.php
@@ -282,7 +282,7 @@ abstract class AsymmetricKey
      *
      * @param string|null $method optional
      */
-    protected static function validatePlugin(string $format, string $type, string $method = null)
+    protected static function validatePlugin(string $format, string $type, ?string $method = null)
     {
         $type = strtolower($type);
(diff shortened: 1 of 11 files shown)


nickvergessen:
@terrafrost any chance to get this into the 2.0 branch too and get a release from there?

terrafrost:
> @terrafrost any chance to get this into the 2.0 branch too and get a release from there?

phpseclib 2.0 doesn't use type hinting. phpseclib 3.0 does but (1) 3.0 doesn't have strict_types enabled and (2) [nullable types](https://wiki.php.net/rfc/nullable_types) weren't introduced until PHP 7.1 and phpseclib 3.0 is supposed to work for versions of PHP as low as 5.6, so that'd be a BC break.

nickvergessen:
Ah okay, then I guess it should be fine. I have to admit, I just searched for 8.4/nullable related PRs in all our dependencies, not checking if they actually spam the log as well.

PR #1995, opened 2024-05-07
Agent: reset supported_private_key_algorithms for every key

forrest79:
Hi, I think I found a bug in the agent login.

I have ssh-agent with 3 certificates loaded. I'm able to log in to my server with a classic ssh client but not with Agent from this library. I figured out that the problem is after the first key is not accepted (if it is not accepted), in the `supported_private_key_algorithms` property are ommited `rsa-sha2-256` and `rsa-sha2-512` algorithms and only `ssh-rsa` is kept. But `ssh-rsa` is disabled by default in my `sshd` and then the second and next keys are tried only with `ssh-rsa` which is declined by `sshd`.

I think every key from the agent should start with all algorithms in the `supported_private_key_algorithms` property. I'm not an expert in this nor in this library - I prepare a simple fix that works for me. But I suppose that the correct solution will be different :-)

### DIFF
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -2335,10 +2335,12 @@ class SSH2
     {
         $this->agent = $agent;
         $keys = $agent->requestIdentities();
+        $orig_algorithms = $this->supported_private_key_algorithms;
         foreach ($keys as $key) {
             if ($this->privatekey_login($username, $key)) {
                 return true;
             }
+            $this->supported_private_key_algorithms = $orig_algorithms;
         }
 
         return false;


terrafrost:
Nice find!

Altho what'd be better is if you did something like `$orig_algorithms = $this->supported_private_key_algorithms` before the `for` loop and then did `$this->supported_private_key_algorithms = $orig_algorithms` after each `$this->privatekey_login($username, $key)` call (ie. same place you have your current code changes).

The reason this would be better than what you're currently doing is that `$this->server_host_key_algorithms` doesn't take NET_SSH2_MSG_EXT_INFO (RFC8308) into consideration.

I could make this change myself but I'll hold off for a few days in case you want to do it? Like if I do it your name won't be in the commit history. idk if that matters to you or not lol.

That said, if you want to make the code change vs me, don't use tabs to indent. In your change the line you added is preceded by one tab and then eight spaces. The tab should be replaced by four spaces (bringing the total number of spaces to twelve).

Thanks!

forrest79:
Hi, thank you for a quick response:

> Altho what'd be better is if you did something like `$orig_algorithms = $this->supported_private_key_algorithms` before the `for` loop and then did `$this->supported_private_key_algorithms = $orig_algorithms` after each `$this->privatekey_login($username, $key)` call (ie. same place you have your current code changes).
> 
> The reason this would be better than what you're currently doing is that `$this->server_host_key_algorithms` doesn't take NET_SSH2_MSG_EXT_INFO (RFC8308) into consideration.
> 
> I could make this change myself but I'll hold off for a few days in case you want to do it? Like if I do it your name won't be in the commit history. idk if that matters to you or not lol.

I updated the patch according to your advice. I'm totally OK with not having my name in the commit history, so if this update is still not perfect, feel free to fix this on your own :-)

> That said, if you want to make the code change vs me, don't use tabs to indent. In your change the line you added is preceded by one tab and then eight spaces. The tab should be replaced by four spaces (bringing the total number of spaces to twelve).

Sorry, I did a quick commit right in the GitHub web interface, not it's OK.

terrafrost:
Looks good! I'll try to merge this this evening. More specifically, I'll cherry pick this to the 1.0 branch and then merge into the 2.0 branch, the 3.0 branch and then the master branch.

terrafrost:
See https://github.com/phpseclib/phpseclib/commit/3b0fb1c05f7ed53a41c10107395c60e1b9a0638f

Thanks!

PR #2000, opened 2024-05-20
Support loading Ed448 in PKCS8 format

LeonMelis:
For Ed448, the private key component is expected to be 57 bytes in length, not 32 like with Ed25519.

This patch will properly support loading Ed448 keys in PKCS8 format

To test, create an Ed448 key with OpenSSL:

```shell
openssl genpkey -algorithm ed448 -out ed448-key.pem
```

Then load the key with PHPSecLib
```php
$private_key_pem = file_get_contents('ed448-key.pem');
$private_key = EC::loadPrivateKey($private_key_pem);
```

Without this proposed patch, PHPSecLib would not be able to load this Ed448 key.

### DIFF
--- a/phpseclib/Crypt/EC/Formats/Keys/PKCS8.php
+++ b/phpseclib/Crypt/EC/Formats/Keys/PKCS8.php
@@ -129,13 +129,22 @@ abstract class PKCS8 extends Progenitor
         $components = [];
 
         if (isset($key['privateKey'])) {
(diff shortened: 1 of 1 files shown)


terrafrost:
The code change looks good but could you add a unit test and re-submit this PR to target the 3.0 branch?

Thanks!

LeonMelis:
I did a revision of the PR here: https://github.com/phpseclib/phpseclib/pull/2003

PR #2011, opened 2024-06-21
fix #2010 back-ported for master of phpseclib

timmit-nl:
fix #2010 back-ported  for master from the 3.0 branch . As the 3.0 branch is different then the master branch


timmit-nl:
made this so you don't have to backport the fix yourself.

PR #2012, opened 2024-06-28
Correction to stream timeout for keep alive

rposky:
Looks like I got sloppy around the keep alive changes, w.r.t. #2009, #2010, #2011. Sorry about that!

I see it's still not quite right though, as we're basing the stream timeout on time since the last read as opposed to time remaining in the keep alive interval. This attempts to fix that along with adding a counter for packet read time and ensuring that `last_packet` is always incremented, even if logging is not enabled.

### DIFF
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -825,11 +825,9 @@ class SSH2
     private $quiet_mode = false;
 
     /**
-     * Time of first network activity
-     *
-     * @var float
+     * Time of last read/write network activity
      */
-    private $last_packet;
+    private $last_packet = null;
 
     /**
      * Exit status returned from ssh if any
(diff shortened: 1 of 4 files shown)


phpseclib/Net/SSH2.php
```diff
@@ -3693,10 +3694,10 @@ private function get_binary_packet()
             $current = microtime(true);
             $message_number = isset(self::$message_numbers[ord($payload[0])]) ? self::$message_numbers[ord($payload[0])] : 'UNKNOWN (' . ord($payload[0]) . ')';
             $message_number = '<- ' . $message_number .
-                              ' (since last: ' . round($current - $this->last_packet, 4) . ', network: ' . round($elapsed, 4) . 's)';
+                              ' (since last: ' . round($current - $this->last_packet, 4) . ', network: ' . round($packet->read_time, 4) . 's)';
             $this->append_log($message_number, $payload);
-            $this->last_packet = $current;
         }
+        $this->last_packet = microtime(true);
```
rposky, line 3700: Use of `microtime(true)` over `$current` in populating `last_packet` discounts any time spent in the logging subroutine.

PR #2016, opened 2024-07-09
style: fix cs and type in `SSH2::get_stream_timeout()`

simPod:


phpseclib/Crypt/Blowfish.php
```diff
@@ -532,7 +532,7 @@ private static function expand0state(array $key, array &$sbox, array &$p): void
         ];
 
         // @codingStandardsIgnoreStart
-        [ $p[0],  $p[1]] = self::encryptBlockHelperFast(     0,      0, $sbox, $p);
```
terrafrost, line 535: Here's what I don't like about this and why I've refrained from doing anything to fix it myself:

```
        [ $p[0],  $p[1]] = self::encryptBlockHelperFast(     0,      0, $sbox, $p);
        [ $p[2],  $p[3]] = self::encryptBlockHelperFast($p[ 0], $p[ 1], $sbox, $p);
        [ $p[4],  $p[5]] = self::encryptBlockHelperFast($p[ 2], $p[ 3], $sbox, $p);
        [ $p[6],  $p[7]] = self::encryptBlockHelperFast($p[ 4], $p[ 5], $sbox, $p);
        [ $p[8],  $p[9]] = self::encryptBlockHelperFast($p[ 6], $p[ 7], $sbox, $p);
        [$p[10], $p[11]] = self::encryptBlockHelperFast($p[ 8], $p[ 9], $sbox, $p);
        [$p[12], $p[13]] = self::encryptBlockHelperFast($p[10], $p[11], $sbox, $p);
        [$p[14], $p[15]] = self::encryptBlockHelperFast($p[12], $p[13], $sbox, $p);
        [$p[16], $p[17]] = self::encryptBlockHelperFast($p[14], $p[15], $sbox, $p);
```
Note how everything lines up.

With this change we'd have this:

```
        [ $p[0],  $p[1]] = self::encryptBlockHelperFast(     0, 0, $sbox, $p);
        [ $p[2],  $p[3]] = self::encryptBlockHelperFast($p[ 0], $p[ 1], $sbox, $p);
        [ $p[4],  $p[5]] = self::encryptBlockHelperFast($p[ 2], $p[ 3], $sbox, $p);
        [ $p[6],  $p[7]] = self::encryptBlockHelperFast($p[ 4], $p[ 5], $sbox, $p);
        [ $p[8],  $p[9]] = self::encryptBlockHelperFast($p[ 6], $p[ 7], $sbox, $p);
        [$p[10], $p[11]] = self::encryptBlockHelperFast($p[ 8], $p[ 9], $sbox, $p);
        [$p[12], $p[13]] = self::encryptBlockHelperFast($p[10], $p[11], $sbox, $p);
        [$p[14], $p[15]] = self::encryptBlockHelperFast($p[12], $p[13], $sbox, $p);
        [$p[16], $p[17]] = self::encryptBlockHelperFast($p[14], $p[15], $sbox, $p);
```
Things are no longer lining up. If we're going to remove the white space before the second parameter then why not also remove it before the first parameter?

`@codingStandardsIgnoreStart` isn't there to look pretty - it's there because blind adherence to coding standards doesn't make sense here and actively imho makes the code less readable.

If `@codingStandardsIgnoreStart` doesn't work the what will?

If there is no alternative than the options are:

- consider dropping whatever tool (psalm?) is resulting in causing unit test failures (which would be a shame because the tool does find legit issues too)
- file a feature request asking that they implement this change

As is this change seems rather like [teaching to the test](https://en.wikipedia.org/wiki/Teaching_to_the_test). Like this is a change that isn't maximizing readability to people - it's maximizing readability to the static analysis tool, even tho at the end of the day it's people who are responsible for fixing stuff - not static analysis tools.

Maybe this line could be added to this:

https://github.com/phpseclib/phpseclib/blob/master/build/psalm_baseline.xml

But, honestly, I really don't like that either. Almost half a megabyte of hard coded exceptions. Maybe some of them can and should be eliminated but maybe some are like this that actively make the code less readable.

Right now I'm working on completely redoing how X.509 and ASN.1 work. That's probably months or years away. After that I figure my next big push will be to fix all the psalm issues. Either fix them in phpseclib, submit bug reports against pslam or something idk. But, honestly, I expect that'll take many months to possibly years, as well.

phpseclib/Crypt/Blowfish.php
```diff
@@ -583,7 +583,7 @@ private static function expandstate(array $data, array $key, array &$sbox, array
         ];
 
         // @codingStandardsIgnoreStart
-        [ $p[0],  $p[1]] = self::encryptBlockHelperFast($data[ 0]         , $data[ 1]         , $sbox, $p);
```
terrafrost, line 586: Same as above

terrafrost (changes requested):
I don't know how I missed this - I apologize!

I mean, on July 12 I flew from Texas to Tokyo on vacation. You submitted this PR on July 9. So that prob had something to do with it. But I feel like I could have reviewed it on either July 10 or July 11 meh.

Anyway, I did fix most of these issues with https://github.com/phpseclib/phpseclib/commit/44d5ca2ab572b047caf09bf7e2397a8fb6bf7b10 . You and I implemented different fixes for one issue (which is causing a merge conflict) and one of the changes I strongly disagree with, as my comment elaborates upon.

phpseclib/Crypt/Blowfish.php
simPod: Hm, there's currently no way to ignore certain lines :/ https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/issues/4512

build/php-cs-fixer.php
```diff
@@ -3,7 +3,10 @@
 declare(strict_types=1);
 
 return (new PhpCsFixer\Config())
-    ->setFinder(PhpCsFixer\Finder::create()->in(__DIR__ . '/..'))
+    ->setFinder(PhpCsFixer\Finder::create()
+        ->in(__DIR__ . '/..')
+        // https://github.com/phpseclib/phpseclib/pull/2016#discussion_r1691282898
+        ->notPath('phpseclib/Crypt/Blowfish.php'))
```
simPod, line 9: I think we should ignore it for now.

terrafrost: tbh if we're going to be ignoring files I think it'd better just to drop PhpCsFixer all together.

In reviewing how PhpCsFixer came to be included in phpseclib...  so there was #1748 for the master branch that I cherry-picked into the 3.0 branch. Then there was #1754 but it ultimately wasn't merged because of whitespace issues similar to what we're dealing with, now. I proposed codesniffer be used instead and then we got #1760. So right now we're doing PhpCsFixer, codesniffer and pslam. Maybe we jut do codesniffer and pslam.

It'd be a shame to lose codesniffer because it does make good finds but I greatly dislike the idea of excluding entire files and of making the types of whitespace changes that it's proposing.

I suppose another idea would be to take the output of PhpCSFixer and filter out errors from lines of code that have `// @codingStandardsIgnoreStart` before it and `// @codingStandardsIgnoreEnd` after.

simPod: I use (and contribute to) https://github.com/doctrine/coding-standard. Can consider to switch to that.

PR #2019, opened 2024-07-11
Fixed IEEE signature format non-fixed length signature problem

toramanlis:
IEEE signature format as referred [here](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/sign#ecdsa) is proposed as following [RFC-4754 specs](https://datatracker.ietf.org/doc/html/rfc4754#section-7).

The specs state

>  The bit lengths of r and s are enforced, if necessary, by pre-pending
   the value with zeros.

In order to fully adhere to these specs, we need to pad the r and s values with zeros up to either 256, 384 or 528 bits (32, 48, or 66 bytes respectively).

I implemented the logic that chooses the smallest of these options that is bigger than both the r and s values.

### DIFF
--- a/phpseclib/Crypt/EC/Formats/Signature/IEEE.php
+++ b/phpseclib/Crypt/EC/Formats/Signature/IEEE.php
@@ -26,6 +26,8 @@ use phpseclib3\Math\BigInteger;
  */
 abstract class IEEE
 {
+    const PART_LENGTHS = [32, 48, 66];
+
     /**
      * Loads a signature
      *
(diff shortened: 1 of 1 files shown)


terrafrost:
Thanks for this! I'm on a plane to Japan right now (I live in Texas so it's
a ways away) and altho I have internet it would seem that github is having
some difficulty loading.

I'll try to look at this as time permits but for this coming week imma be
distracted with touristy stuff!

toramanlis:
Dude, absolutely don't work on a holiday anyway. Be a nice boss to yourself :) Cheers

terrafrost:
So altho the RFC only mentions three different curves there's no reason to assume that the signature format can only be used with those 3x curves. idk what the IEEE specification says but what I'm thinking is that maybe a third parameter - be it `int $length` or `Curve $curve` could be added to `save()`?:

So like if someone is using secp112r1 one could get the appropriate length thusly:

```
$curve = new \phpseclib3\Crypt\EC\Curves\secp112r1();
echo $curve->getLength();
```

If this signature format can only be used with the 3x curves mentioned then passing in the `$curve` object might be preferrable. Then the name could be extracted through reflection and an exception thrown if the curve isn't among the supported curves.

terrafrost:
This should work better as a result of https://github.com/phpseclib/phpseclib/commit/04a559debb1f78f7f92da7e9a6be1497f93cb432 and https://github.com/phpseclib/phpseclib/commit/a015cded00e8eda8b162a69af98c11cf30b7e004 . Still need to merge into master but I'll try to do that tonight or tomorrow.

Thank for bringing this to my attention!

Issue #2020, opened 2024-07-19
Maximum execution time exceeded in version 3.0.38


liborm85:
Version 3.0.38 changed behavior for execution time.

If multiple exec commands are run in version 3.0.38 (or 3.0.39) is throw error maximum execution time exceeded. It works correctly in version 3.0.37.

Example for reproduce problem (sleep 1 simulate more difficult command):
```php
<?php

set_time_limit(30);

$sftp = new \phpseclib3\Net\SFTP('host', 22, 60);
$sftp->login('login', 'password');
$sftp->setTimeout(0);

$idx = 0;
do {
    $sftp->exec('sleep 1; echo test;');
    var_dump($idx);

    $idx++;
} while ($idx < 60);

echo 'OK';
```

Correct output in version 3.0.37:
```
int(0) int(1) int(2) int(3) int(4) int(5) int(6) int(7) int(8) int(9) int(10) int(11) int(12) int(13) int(14) int(15) int(16) int(17) int(18) int(19) int(20) int(21) int(22) int(23) int(24) int(25) int(26) int(27) int(28) int(29) int(30) int(31) int(32) int(33) int(34) int(35) int(36) int(37) int(38) int(39) int(40) int(41) int(42) int(43) int(44) int(45) int(46) int(47) int(48) int(49) int(50) int(51) int(52) int(53) int(54) int(55) int(56) int(57) int(58) int(59) OK
```

Maximum execution time exceeded throw in version 3.0.38 or 3.0.39:
```
int(0) int(1) int(2) int(3) int(4) int(5) int(6) int(7) int(8) int(9) int(10) int(11) int(12) int(13) int(14) int(15) int(16) int(17) int(18) int(19) int(20) int(21) int(22) int(23) int(24) int(25) int(26) int(27)
Fatal error: Maximum execution time of 30 seconds exceeded in /vendor/phpseclib/phpseclib/phpseclib/Net/SSH2.php on line 3554
```

terrafrost:
First, apologies for the delay. I traveled to Japan from Texas for a week on vacation and just got back.

Second, I'm not able to duplicate the issue on either 3.0.38 or 3.0.39. When I do `define('NET_SSH2_LOGGING', 3)` on 3.0.39 I get this between each loop:

```
-> NET_SSH2_MSG_CHANNEL_OPEN (since last: 0.0005, network: 0s)
[...]

PR #2023, opened 2024-08-01
SSH2: Defer to default socket timeout in absence of more specific value

rposky:
Attempting to address #2020, this issue formalizes the implied behavior prior to #1999 regarding PHP stream timeouts, by defaulting to PHP's `default_socket_timeout` for stream read functions when the user has not requested a timeout or keep-alive interval. This change mirrors the default timeout behavior of PHP socket streams, as evidenced [here](https://github.com/php/php-src/blob/master/main/streams/xp_socket.c#L964), and thus the behavior prior to #1999 changes, when `stream_set_timeout` was conditionally invoked. It is noteworthy that the prior implementation would not permit an update to remove a timeout via `SSH2:setTimeout`, given the conditional invocation of `stream_set_timeout`, now supported.

The following simple test confirms the underlying use of the default socket timeout when `stream_set_timeout` is never invoked, as execution terminates after 10 seconds, as defined.
```
<?php
$start = time();
ini_set('default_socket_timeout', 10);
$fsock = fsockopen('127.0.0.1', 22);
stream_get_contents($fsock);
printf("Time: %d", (time() - $start));
```

### DIFF
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -2747,7 +2747,7 @@ class SSH2
      * Set Timeout
      *
      * $ssh->exec('ping 127.0.0.1'); on a Linux host will never return and will run indefinitely.  setTimeout() makes it so it'll timeout.
-     * Setting $timeout to false or 0 will mean there is no timeout.
+     * Setting $timeout to false or 0 will revert to the default socket timeout.
      *
      * @param mixed $timeout
      */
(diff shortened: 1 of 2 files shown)


terrafrost:
Looks good!

Also, apologies for the delay - I'm once again on vacation lol. I'm doing Black Hat USA in Vegas this week but left a few days early to do sight seeing.

Issue #2021, opened 2024-07-23
Fortify on Demand Issue - Possible Variable Overwrite: Global Scope


felipef0xx:
### **Summary**
The program invokes `extract()`, which can overwrite global variables and might open the door for attackers. The program invokes a function that can overwrite global variables, which can open the door for attackers.

### **Explanation**
Taking **MSBLOB.php** as an example:
Functions that can overwrite global variables that are already initialized can allow an attacker to influence the execution of code that relies on the overwritten variables.
In this case the function extract() in MSBLOB.php on line 119 can overwrite global variables.

**Example 1:** If an attacker supplies a malicious value for str in the following segment of PHP code, then the call to mb_parse_str() might overwrite any arbitrary variables, including first. In this case, if a malicious value that contains JavaScript overwrites first, then the program is vulnerable to cross-site scripting.

```php
<?php
$first="User";
...
$str =  $_SERVER['QUERY_STRING'];
mb_parse_str($str);
echo $first;
?>
```

### **Locations**
phpseclib/phpseclib/Common/Functions/Strings.php : 122
phpseclib/phpseclib/Crypt/RSA.php : 355
phpseclib/phpseclib/Crypt/Common/Formats/Keys/PKCS8.php : 341
phpseclib/phpseclib/Crypt/Common/Formats/Keys/PKCS8.php : 359
phpseclib/phpseclib/Crypt/Common/Formats/Keys/PKCS8.php : 369
phpseclib/phpseclib/Crypt/Common/Formats/Keys/PKCS8.php : 378
phpseclib/phpseclib/Crypt/Common/Formats/Keys/PKCS8.php : 405
phpseclib/phpseclib/Crypt/Common/Formats/Keys/PuTTY.php : 189
phpseclib/phpseclib/Crypt/Common/Formats/Keys/PuTTY.php : 217
phpseclib/phpseclib/Crypt/Common/Formats/Keys/PuTTY.php : 309
phpseclib/phpseclib/Crypt/DSA/PrivateKey.php : 91
phpseclib/phpseclib/Crypt/DSA/PublicKey.php : 44
phpseclib/phpseclib/Crypt/DSA/Formats/Keys/PuTTY.php : 62
phpseclib/phpseclib/Crypt/EC/PrivateKey.php : 159
phpseclib/phpseclib/Crypt/EC/PublicKey.php : 119
phpseclib/phpseclib/Crypt/RSA/Formats/Keys/MSBLOB.php : 86
[...]

PR #2045, opened 2024-10-25
Fix - Possible Variable Overwrite: Global Scope

felipef0xx:
This PR applies the recommended changes by Fortify to remediate the `Possible Variable Overwrite: Global Scope` vulnerability.

> Invoke extract(array $var_array [, int $extract_type [, string $prefix]]) with the second argument set to EXTR_SKIP, which prevents the function from overwriting global variables that are already defined.

CWE ID: 473

This was previously discussed on #2021 .


terrafrost:
I've reverted this. This commit caused the 178 failures in the unit tests. Should have done a better job of checking before merging this. If you can fix those errors I'll consider re-merging this.

Thanks!

PR #2055, opened 2024-12-02
fix non-existent variable usage

jmarchan-ByTel:


terrafrost:
It's a good observation but the fix has left all of the unit tests failing.

I went ahead and implemented my own fix:

https://github.com/phpseclib/phpseclib/commit/0efd3176d8b449bf61547e8f479379039c08531d

Thanks!

PR #2057, opened 2024-12-03
FIX undefined var

jmarchan-ByTel:


terrafrost:
Nice find but the code change, itself, is not compliant with phpseclib's CS standards. Consequently, I just implemented a version of your code changes that are compliant:

https://github.com/phpseclib/phpseclib/commit/573f4aaf02bd3faf75bffebfa3cbef1ec56b6f52

PR #2059, opened 2024-12-20
Ensure that users of PHP 8.2.26-nmm1 users don't utilize the GMP functions to prevent base and exponent error from occurring

andtown:
It appears that PHP 8.2.26-nmm1 users are still experiencing the "base and exponent overflow" issue during the call to `gmp_pow()` function

PHP 8.2.26-nmm1 is considered lower than PHP 8.2.26 thus GMP extension will be used for arbitrary-length integer operations. Therefore, we'd like to propose a change in this PR.


terrafrost:
I cherry-picked this into the 1.0 branch and merged that into 2.0 / 3.0 / master:

https://github.com/phpseclib/phpseclib/commit/3111903089f1c40b61a5ddb730fdb028b40bf0ce

Thanks!

PR #2063, opened 2025-01-02
Remove extract function - Possible Variable Overwrite

felipef0xx:
This PR attempts to remediate the `Possible Variable Overwrite: Global Scope` vulnerability (CWE 473).

I opened PR #2045, but it was reverted due to too many failed tests.
Basically, it fails because the `EXTR_SKIP` parameter in `extract()` function were not writing values in existing variables, which is not the intended behavior in some cases.

So I'm proposing a new approach: replace `extract()` with variable assignment.
In addition to preventing user input from rewriting global variables, this improves code readability, as it becomes clear which variables are being assigned.

### DIFF
--- a/phpseclib/File/ASN1.php
+++ b/phpseclib/File/ASN1.php
@@ -272,7 +272,7 @@ abstract class ASN1
             // tags of indefinte length don't really have a header length; this length includes the tag
             $current += ['headerlength' => $length + 2];
             $start += $length;
-            extract(unpack('Nlength', substr(str_pad($temp, 4, chr(0), STR_PAD_LEFT), -4)));
+            ['length' => $length] = unpack('Nlength', substr(str_pad($temp, 4, chr(0), STR_PAD_LEFT), -4));
             /** @var integer $length */
         } else {
             $current += ['headerlength' => 2];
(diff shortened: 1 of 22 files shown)


phpseclib/File/ASN1.php
```diff
@@ -272,7 +272,7 @@ private static function decode_ber(string $encoded, int $start = 0, int $encoded
             // tags of indefinte length don't really have a header length; this length includes the tag
             $current += ['headerlength' => $length + 2];
             $start += $length;
-            extract(unpack('Nlength', substr(str_pad($temp, 4, chr(0), STR_PAD_LEFT), -4)));
+            ['length' => $length] = unpack('Nlength', substr(str_pad($temp, 4, chr(0), STR_PAD_LEFT), -4));
             /** @var integer $length */
```
terrafrost, line 276: Might as well delete the `/** @var integer $length */` style comments

phpseclib/Net/SFTP/Stream.php
```diff
@@ -124,7 +124,33 @@ public function __construct()
...
+        }
+        if (isset($url['path'])) {
+            $path = $url['path'];
+        }
+        if (isset($url['query'])) {
+            $query = $url['query'];
+        }
+        if (isset($url['fragment'])) {
+            $fragment = $url['fragment'];
+        }
```
terrafrost, line 152: I guess it's just me but I'd almost prefer this:

```php
$keys = ['scheme', 'host', 'port', 'user', 'pass', 'path', 'query', 'fragment'];
foreach ($keys as $key) {
    $$key = $url[$key];
}
```

terrafrost (commented):
Haven't used the PHP 7.1+ associate array destructuring before. I've given this a cursory review and have posted my thoughts. I'll give it a more thorough review later, as time permits.

Thanks!

Issue #1367, opened 2019-04-28
UUID OIDs and OIDs 2.40+ are not supported


danielmarschall:
I have looked at your implementation of Object Identifiers and found out that there are many issues.

(1) The [modulo-40 code](https://github.com/phpseclib/phpseclib/blob/master/phpseclib/File/ASN1.php#L462) is not correct, since it does not support the case of OID 2.40, 2.41, etc. Your modulo-40 code would encode 2.40 as 3.0.

(2) Since no BigInteger code is used, the max size of arcs is limited, and therefore UUID-OIDs (OIDs inside 2.25) are not supported.

You can find a implementation which solves all problems [here](https://misc.daniel-marschall.de/asn.1/oid-converter/php/). Testcases/checks of the DER encoding can be done [here](https://misc.daniel-marschall.de/asn.1/oid-converter/online.php).

bantu:
@danielmarschall Thanks you for your report.

I did not have a change to look into your implementation and whether it could be integrated into phpseclib. In case we would like to, would you be willing to release your code under the MIT license? You are welcome to submit a pull request introducing tests reproducing the issues you're seeing and fixing them.

danielmarschall:
I was actually changing a lot of my source codes from GPLv3 to Apache2. Would that fit better? I guess Apache2 and MIT are compatible with each other, so you could include my library in your project.

terrafrost:
You appear to be correct. [ITU-T X.690, ISO/IEC 8825-1 § 8.19.5](https://www.itu.int/ITU-T/studygroups/com17/languages/X.690-0207.pdf#page=22) gives 2.100.3 as an example and shows the encoding to be 0x813403.

I'll look at your library and test cases and see what I can do, as time permits, to improve how it works in phpseclib. As for including you're library...  phpseclib doesn't have the gmp extension as a dependency and I don't intend to make it one. Also, to facilitate third party use, I'd recommend you make your package install-able via Composer / Packagist.

Anyway, thanks for the report and, like I said, I'll try to work on this as time permits.

Thanks!

terrafrost:
Here's what I've done so far for decoding OIDs:

https://github.com/terrafrost/phpseclib/commit/8784f4bacb116f43b1d4ced75da513652008d0be

terrafrost:
I've finished up the encoding step as well along with the unit test:

https://github.com/terrafrost/phpseclib/commit/e793461543d16a5d52f8805882bf6d45bf8c6cde

[...]

PR #2066, opened 2025-01-13
Support ASN.1 tag encoding with values greater than 30.

mmauv:
Allow ASN.1 DER encoding of tags with values greater than 30.
From X.690-0207.pdf#page=12 section 8.1.2.4:
> For tags with a number greater than or equal to 31, the identifier shall comprise a leading octet followed by
one or more subsequent octets.
> ...
> c) bits 5 to 1 shall be encoded as 11111<sub>2</sub> .
> ...
> a) bit 8 of each octet shall be set to one unless it is the last octet of the identifier octets;
> b) bits 7 to 1 of the first subsequent octet, followed by bits 7 to 1 of the second subsequent octet, followed in
turn by bits 7 to 1 of each further octet, up to and including the last subsequent octet in the identifier
octets shall be the encoding of an unsigned binary integer equal to the tag number, with bit 7 of the first
subsequent octet as the most significant bit

### DIFF
--- a/phpseclib/File/ASN1.php
+++ b/phpseclib/File/ASN1.php
@@ -931,7 +931,7 @@ abstract class ASN1
                                     $subtagvalue = $constant->bitwise_and($mask);
                                     $subtagvalue->setPrecision(8);
                                     $subtag = (chr(0x80) | $subtagvalue->toBytes()) . $subtag;
-                                    $constant = $constant->bitwise_rightShift(shift: 7);
+                                    $constant = $constant->bitwise_rightShift(7);
                                 }
                                 $subtag[strlen($subtag) - 1] = $subtag[strlen($subtag) - 1] & chr(0x7F);
                                 $subtag = chr((self::CLASS_CONTEXT_SPECIFIC << 6) | 0x20 | 0x1f) . $subtag;
--- a/phpseclib/File/ASN1.php
+++ b/phpseclib/File/ASN1.php
@@ -923,15 +923,12 @@ abstract class ASN1
                             if ($child['constant'] <= 30) {
                                 $subtag = chr((self::CLASS_CONTEXT_SPECIFIC << 6) | 0x20 | $child['constant']);
                             } else {
-                                $zero = new BigInteger();
-                                $mask = new BigInteger(0x7F);
-                                $constant = new BigInteger($child['constant']);
+                                $constant = $child['constant'];
                                 $subtag = '';
-                                while (!$constant->equals($zero)) {
-                                    $subtagvalue = $constant->bitwise_and($mask);
-                                    $subtagvalue->setPrecision(8);
-                                    $subtag = (chr(0x80) | $subtagvalue->toBytes()) . $subtag;
-                                    $constant = $constant->bitwise_rightShift(7);
+                                while ($constant > 0) {
+                                    $subtagvalue = $constant & 0x7F;
+                                    $subtag = (chr(0x80 | $subtagvalue )) . $subtag;
+                                    $constant = $constant >> 7;
                                 }
                                 $subtag[strlen($subtag) - 1] = $subtag[strlen($subtag) - 1] & chr(0x7F);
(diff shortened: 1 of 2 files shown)


terrafrost (changes requested):
Why are you using BigInteger? $child['constant'] is an integer and converting that to a BigInteger at the point where you are isn't going to all of a sudden enable you to use bigger integers. And although there's absolutely a use case for being able to do constants > 30 is there one for being able to do constants > PHP_INT_MAX?

Other than that I didn't realize that this wasn't supported. Like phpseclib can decode ASN.1 stuff with > 30 but it can't encode them. So nice catch there!

A unit test would be good, as well.

If it helps (and partially for my own reference lol) this is what I used to test it:

```php
$map = [
    'type'     => ASN1::TYPE_SEQUENCE,
    'children' => [
        'demo' => [
            'constant' => 0xFFFFFFFF,
            'optional' => true,
            'explicit' => true,
            'default'  => 'v1',
            'type'     => ASN1::TYPE_INTEGER,
            'mapping' => ['v1', 'v2', 'v3']
        ]
    ]
];

$data = ['demo' => 'v3']; // should be 300abf8fffffff7f03020102

$str = ASN1::encodeDER($data, $map);

$decoded = ASN1::decodeBER($str);
$arr = ASN1::asn1map($decoded[0], $map);
```

mmauv:
> Why are you using BigInteger? $child['constant'] is an integer and converting that to a BigInteger at the point where you are isn't going to all of a sudden enable you to use bigger integers. And although there's absolutely a use case for being able to do constants > 30 is there one for being able to do constants > PHP_INT_MAX?

BigInteger was used for coherence with the implementation of object identifier values [File/ASN1.php line 1232](https://github.com/mmauv/phpseclib/blob/a004cbc2108e70cce05c6ac4f740e238d70e0a68/phpseclib/File/ASN1.php#L1232). If it's not relevant, the implementation can be done without using BigInteger.

I'm not sure if there's a use case with constants > PHP_INT_MAX, since PHP_INT_MAX is quite big.

Other requested changes were done in the last commit.

terrafrost:
> BigInteger was used for coherence with the implementation of object identifier values [File/ASN1.php line 1232](https://github.com/mmauv/phpseclib/blob/a004cbc2108e70cce05c6ac4f740e238d70e0a68/phpseclib/File/ASN1.php#L1232). If it's not relevant, the implementation can be done without using BigInteger.

Good catch - that's prob something that ought to be changed. I can do that.

That said, altho encoding with tags using BigInteger is consistent with how phpseclib handles OIDs, it's inconsistent with how phpsecib _decodes_ tags:

https://github.com/phpseclib/phpseclib/blob/baba459ca1b2619b173805ca3dfedc4e5a88eac6/phpseclib/File/ASN1.php#L234-L253

So if you could change it to use ints that'd be great. Meanwhile, I'll update how OIDs are handled.

Thanks!!

terrafrost:
>> BigInteger was used for coherence with the implementation of object identifier values [File/ASN1.php line 1232](https://github.com/mmauv/phpseclib/blob/a004cbc2108e70cce05c6ac4f740e238d70e0a68/phpseclib/File/ASN1.php#L1232). If it's not relevant, the implementation can be done without using BigInteger.
>
> Good catch - that's prob something that ought to be changed. I can do that.

So I modified ASN1.php to make it stop using BigIntegers for OIDs and one of the unit tests then failed. Turns out phpseclib started using BigInteger's because of #1367.

So I guess there is a valid use-case for using BigInteger's for OIDs but I am aware of no valid use case for using them with tags. And plus, as previously stated, BigInteger's are not used for decoding tags.

Anyway, imma try to squash your commits into one and then cherry-pick that to the 3.0 branch. Maybe this evening or tomorrow or some such.

Thanks!

terrafrost:
This has been squashed into one commit and cherry-picked to the 3.0 branch:

https://github.com/phpseclib/phpseclib/commit/a02712f1c327524b4ef8cb3f62d13d9bc0eeec58

Thanks!

Issue #2082, opened 2025-05-15
ConnectionClosedException thrown when other lines precede version string


joecoons:
Existing comment in `SSH2::connect()`:
```php
/* According to the SSH2 specs,

          "The server MAY send other lines of data before sending the version
           string.  Each line SHOULD be terminated by a Carriage Return and Line
           Feed.  Such lines MUST NOT begin with "SSH-", and SHOULD be encoded
           in ISO-10646 UTF-8 [RFC3629] (language is not specified).  Clients
           MUST be able to process such lines." */
```
Link to specs in question: [RFC 4253: Section 4.2](https://datatracker.ietf.org/doc/html/rfc4253#section-4.2)

Starting in version **3.0.42** the assignment of `$this->server_identifier` changed from using `$temp` to using `$data`. This change causes it to include any lines *before* the SSH version indicator. This apparently causes problems, possibly causing it to read to the end of the buffer early. The exception is thrown later on in `get_binary_packet()` on line 3618.

We can use 3.0.41 for now, but this stops us from upgrading further.

My solution is to simply revert the single change to use `$temp` again.

joecoons:
Thanks again!

PR #2083, opened 2025-05-15
Excluding extra lines before version string from server_identifier

joecoons:
https://github.com/phpseclib/phpseclib/issues/2082

Existing comment in `SSH2::connect()`:
```php
/* According to the SSH2 specs,

          "The server MAY send other lines of data before sending the version
           string.  Each line SHOULD be terminated by a Carriage Return and Line
           Feed.  Such lines MUST NOT begin with "SSH-", and SHOULD be encoded
           in ISO-10646 UTF-8 [RFC3629] (language is not specified).  Clients
           MUST be able to process such lines." */
```
Link to specs in question: [RFC 4253: Section 4.2](https://datatracker.ietf.org/doc/html/rfc4253#section-4.2)

Starting in version **3.0.42** the assignment of `$this->server_identifier` changed from using `$temp` to using `$data`. This change causes it to include any lines *before* the SSH version indicator. This apparently causes problems, possibly causing it to read to the end of the buffer early. The exception is thrown later on in `get_binary_packet()` on line 3618.

We can use 3.0.41 for now, but this stops us from upgrading further.

My solution is to simply revert the single change to use `$temp` again.

### DIFF
--- a/phpseclib/Net/SSH2.php
+++ b/phpseclib/Net/SSH2.php
@@ -1487,7 +1487,7 @@ class SSH2
 
         $extra = $matches[1];
 
-        $this->server_identifier = trim($data, "\r\n");
+        $this->server_identifier = trim($temp, "\r\n");
         if (strlen($extra)) {
             $this->errors[] = $data;
         }


terrafrost:
Here's the commit that introduced this behavior:

https://github.com/phpseclib/phpseclib/commit/778035aa7172ddc00d21aa12369637d2be915c4c

> This change causes it to include any lines _before_ the SSH version indicator.

Sounds consistent with what the RFC says.

> This apparently causes problems, possibly causing it to read to the end of the buffer early.

Can you elaborate on the problems? What are you using as the SSH server? What is the server identification string?

That said, it's also not clear how your code change would prevent an exception being thrown if one is already being thrown. Like your commit doesn't change how much data is being read from the buffer / socket connection / whatever - it just changes what `$this->server_identifier` is being set to. Like as is your fix seems akin to saying that `$a = file_get_contents('100gb file')` is causing PHP to run out of memory BUT adding `$a = '';` after makes the issue go away. I mean, sure, in this scenario, maybe [dead-code elimination](https://en.wikipedia.org/wiki/Dead-code_elimination) could result in that behavior idk but barring that it's unclear how changing what a variable is set to would prevent issues, should issues exist.

terrafrost:
I think I figured it out.

So in the part of RFC4253 that you and I quoted there's this bit

> The server MAY send other lines of data before sending the version string.

The implication of that is that the other lines that come before the version string are _not_ part of the version string.

As for why this is an issue...  later in the same RFC there's this:

```
   The hash H is computed as the HASH hash of the concatenation of the
   following:

      string    V_C, the client's identification string (CR and LF
                excluded)
      string    V_S, the server's identification string (CR and LF
                excluded)
      string    I_C, the payload of the client's SSH_MSG_KEXINIT
      string    I_S, the payload of the server's SSH_MSG_KEXINIT
      string    K_S, the host key
      mpint     e, exchange value sent by the client
      mpint     f, exchange value sent by the server
      mpint     K, the shared secret
```
So the server identification string needs to be included in the hash. With your code change `SSH-2.0-JSCAPE9` is being fed into the hash whereas with mine it's `Evox Images Portal FTP Server` and `SSH-2.0-JSCAPE9`.

So per that I guess `getServerIdentification()` prob should return just `SSH-2.0-JSCAPE9`, however, I think `getLog()` should return both.

I'll do a little more testing. Like with your changes what if you have a server identification string that's 300 bytes long? ie. `SSH-2.0-zzzzzzzz` where z is repeated 300x or whatever?

I gotta head out very shortly but I'll take a look at this further prob tomorrow.

Thank you for your patience!

terrafrost:
https://github.com/phpseclib/phpseclib/commit/9d8e42d41ed91c6ffb2100d188c3331155ec7853 should fix this.

Thanks!

terrafrost:
Still need to backport the fix to the 2.0 and possibly the 1.0 branches but I just wanted to get the 3.0 fix out before that.

PR #2092, opened 2025-07-12
Remove mcrypt suggestion

AJenbo:
This was already done on the master branch a few years ago. Having the recommendation can be misleading especially when OpenSSL is already installed as it's not clear that OpenSSL will be used already. Mcrypt is abandoned and probably not a good idea to install in the first place but can also be problematic to install at all.

If you still prefer that mcrypt is listed as recommended extensions then it would at least be good to update the description to make it more clear that it's not needed in case openssl is already installed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -61,7 +61,6 @@
     "suggest": {
         "ext-libsodium": "SSH2/SFTP can make use of some algorithms provided by the libsodium-php extension.",
         "ext-openssl": "Install the OpenSSL extension in order to speed up a wide variety of cryptographic operations.",
-        "ext-mcrypt": "Install the Mcrypt extension in order to speed up a few other cryptographic operations.",
         "ext-gmp": "Install the GMP (GNU Multiple Precision) extension in order to speed up arbitrary precision integer arithmetic operations.",
         "ext-dom": "Install the DOM extension to load XML formatted public keys."
     },


terrafrost:
mcrypt isn't used in the master branch, however, the 3.0 branch works on versions of PHP as low as 5.6 and mcrypt wasn't removed [until PHP 7.1](https://wiki.php.net/rfc/mcrypt-viking-funeral).

And sure, even on PHP 5.6 and 7.0, one might think OpenSSL would still be the faster option, but that's not always the case. Two examples:

- AES is Rijndael with a 128 bit block size, however, Rijndael also supports block sizes of 160, 192, 224 and 256 bits. OpenSSL, however, only supports AES (ie. Rijndael with a 128-bit block size) so if you need to use Rijndael with a 256 bit block size, the only two options are to use the pure PHP implementation or to use mcrypt and, if mcrypt is available, it's going to be faster than the pure PHP implementation.

- OpenSSL doesn't support "continuous mode" (which SSH uses). For block ciphers phpseclib can emulate it with OpenSSL but for stream ciphers (such as RC4) phpseclib cannot emulate it (ChaCha20 is a stream cipher as well but my recollection is that SSH doesn't implement ChaCha20 the same way that it does RC4 but I'd need to freshen up on that part of the code to be sure of that)

Finally, technically, mcrypt can still be installed on versions of PHP as new as 8.4. It just needs to be installed via PECL: https://pecl.php.net/package/mcrypt . Indeed, the PHP 8.4 Docker container in https://github.com/phpseclib/docker-php/blob/8.4/Dockerfile installs mcrypt via pecl. That said, I don't imagine too many people do that, and for that reason, the master branch (which'll become v4.0 in the fullness of time) does not even _try_ to make use of mcrypt, even if it's available, however, the 3.0 branch is a different story.

I mean, if I were to bump the PHP requirements for the 3.0 branch to PHP 7.2, then sure, I guess removing mcrypt might make more sense, but, imho, doing that would be a BC break, and I'm not going to knowingly break BC. https://phpseclib.com/docs/versioning talks more about my view w.r.t. LTS and BC and what not in more detail.

AJenbo:
I know that it can be installed on 8.4. i did mention that in my description. But even from PECL it is a bit tricky to install, for example it fails on the official PHP 8.4 docker image ... which is why I came here.

I'm not suggesting to remove mcrypt support, just either remove it from suggestions or clarify that there is not need to put effort into installing it if openssl is already installed.

terrafrost:
I think updating the description would be better. Maybe having it say something like "_Install the Mcrypt extension in order to speed up a few less commonly used symmetric key ciphers that the OpenSSL extension doesn't support_".

Like it may well be that, depending on your use case, you could benefit from the mcrypt extension being installed, even on PHP 8.4 and even in spite of the difficulty that I guess installing it might be in Docker (altho that said, [I had no difficulty getting it working in Docker with the PHP 8.4 container](https://github.com/phpseclib/docker-php/blob/8.4/Dockerfile)). But I do think the use cases that would benefit from the mcrypt extension being installed are pretty far and few between.

Issue #1118, opened 2017-04-18
Security issue - throwing username and password in stacktrace


gandra:
I am playing with [thephpleague flysystem](https://flysystem.thephpleague.com) and under the hood it use phpseclib. When there is a connection problem in the stack trace are thrown credentials:

```
in SSH2.php line 1011
at HandleExceptions->handleError(1024, 'Cannot connect to my-host.com:22. Error 0. php_network_getaddresses: getaddrinfo failed: Name or service not known', '/home/vagrant/Code/my-project/vendor/phpseclib/phpseclib/phpseclib/Net/SSH2.php', 1011, array('start' => 1492543756.398536, 'errno' => 0, 'errstr' => 'php_network_getaddresses: getaddrinfo failed: Name or service not known', 'host' => 'my-host.com:22'))
at user_error('Cannot connect to my-host.com:22. Error 0. php_network_getaddresses: getaddrinfo failed: Name or service not known') in SSH2.php line 1011
at SSH2->_connect() in SSH2.php line 1888
at SSH2->_login('username', 'secret')
at call_user_func_array(array(object(SFTP), '_login'), array('username', 'secret')) in SFTP.php line 390
at SFTP->login('username', 'secret') in SftpAdapter.php line 193
...
```

I consider it as a security hole. Username and password should not be thrown in the stacktrace. 
At least by default. If for some reason you still prefer to throw password in the stacktrace it should be enabled by explicit configuration. By default password should not be thrown as part of a stack trace.

GrahamCampbell:
Not a security bug imo.

> By default password should not be thrown as part of a stack trace.

Not how stack traces work.

gandra:
@GrahamCampbell ,
I have seen my password in the stacktrace. I think this should not happen. Password should not appear in the stacktrace.

GrahamCampbell:
> Password should not appear in the stacktrace.

Right, but my point is, how do you propose we do this...

terrafrost:
I guess thephpleague takes errors and turns them into exceptions that are thrown. So in that case you could simply put the login() method around a try / catch block and catch all Exceptions.

Sample code:

```php
<?php
function login($username, $password) {
    throw new Exception();
}

login('user', 'pass');
```
Oh noes, you can see the user / pass in the stack trace.

```php
<?php
function login($username, $password) {
    throw new Exception();
}

try {
    login('user', 'pass');
} catch (Exception $e) {}
```
Yay, no full stack trace.

[...]

PR #2096, opened 2025-08-21
Add #[SensitiveParameter]-Attribute for password parameters to prevent leakage in stack traces

fox34:
Fixes issue #1118, but requires PHP 8.2. As such, a generic polyfill is added to bootstrap.php for the time being.

### DIFF
--- a/phpseclib/Crypt/Common/AsymmetricKey.php
+++ b/phpseclib/Crypt/Common/AsymmetricKey.php
@@ -127,7 +127,7 @@ abstract class AsymmetricKey
      * @param string|array $key
      * @return PublicKey|PrivateKey
      */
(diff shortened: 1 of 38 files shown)


mvorisek:
https://github.com/symfony/polyfill-php82/blob/v1.33.0/Resources/stubs/SensitiveParameter.php dependency should be preferred, the current bootstrap is conditioned only on PHP version, but 2 bootstraps like this will cause duplicate class error.

fox34:
> https://github.com/symfony/polyfill-php82/blob/v1.33.0/Resources/stubs/SensitiveParameter.php dependency should be preferred, the current bootstrap is condition only on PHP version, but 2 bootstraps like this will cause duplicate class error.

Nice catch. Would you mind creating a PR?

PR #2106, opened 2025-11-13
fix: argument typing

vitormattos:
Looking the first method row, is possible to see that this method can accept array or string.

### DIFF
--- a/phpseclib/File/X509.php
+++ b/phpseclib/File/X509.php
@@ -2294,8 +2294,12 @@ class X509
 
     /**
      * Load a Certificate Revocation List
+     *
+     * @param string|array $crl
+     * @param int $mode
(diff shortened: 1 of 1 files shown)


terrafrost:
There's not a lot of point to merging this PR because it'd just be undone (and conflict) with #2105, which I'll prob merge in this weekend. That PR splits CRL (and SPKAC / CSR) into their own classes and makes the load method static. Union types are also employed:

```php
public static function load(string|array|Constructed $crl, int $mode = ASN1::FORMAT_AUTO_DETECT): CRL
```
I haven't merged #2105 in, sooner, because I wanted to give people the chance to opine in, since it's a complete paradigm shift w.r.t. ASN1 / X509s / CRLs / etc.

PR #2121, opened 2026-01-09
feat: add OpenSSL support for Ed25519/Ed448

binaryfire:
PHP 8.4 added `OPENSSL_KEYTYPE_ED25519` and `OPENSSL_KEYTYPE_ED448`, so phpseclib can now use OpenSSL for these curves instead of requiring libsodium.

## Why?

phpseclib already prefers OpenSSL for other EC curves (secp256k1, secp384r1, etc). This brings Ed25519/Ed448 in line with that.

## Changes

- `EC::createKey()` - use OpenSSL for Ed25519/Ed448 key generation when available
- `EC\PrivateKey::sign()` - use OpenSSL for signing
- `EC\PublicKey::verify()` - use OpenSSL for verification
- Added tests for the new code paths

Falls back to libsodium (then pure-PHP) when OpenSSL isn't available, so fully backward compatible.

Ed25519ctx still uses the existing implementation since OpenSSL doesn't support contexts.

## Testing

All existing tests pass. New tests cover key generation, signing, verification, format round-trips, and libsodium interop.

### DIFF
--- a/phpseclib/Crypt/EC.php
+++ b/phpseclib/Crypt/EC.php
@@ -153,7 +153,7 @@ abstract class EC extends AsymmetricKey
 
         $curve = strtolower($curve);
 
-        // Use OpenSSL for Ed25519/Ed448 if available (PHP 8.4+ with OpenSSL 3.0+)
+        // OpenSSL supports Ed25519/Ed448 when OPENSSL_KEYTYPE_ED25519/ED448 constants are defined
         if (self::$engines['OpenSSL'] && $curve == 'ed25519' && defined('OPENSSL_KEYTYPE_ED25519')) {
             $key = openssl_pkey_new(['private_key_type' => OPENSSL_KEYTYPE_ED25519]);
             openssl_pkey_export($key, $privateKeyStr);
--- a/phpseclib/Crypt/EC/PrivateKey.php
+++ b/phpseclib/Crypt/EC/PrivateKey.php
@@ -115,13 +115,12 @@ final class PrivateKey extends EC implements Common\PrivateKey
         }
 
         if ($this->curve instanceof TwistedEdwardsCurve) {
-            // Use OpenSSL for Ed25519/Ed448 signing if available (PHP 8.4+)
-            // Ed25519ctx (with context) is not supported by OpenSSL, so skip if context is set
+            // OpenSSL supports Ed25519/Ed448 but not Ed25519ctx (context), so skip if context is set
             if (self::$engines['OpenSSL'] && !isset($this->context)) {
                 $keyTypeConstant = $this->curve instanceof Ed25519 ? 'OPENSSL_KEYTYPE_ED25519' : 'OPENSSL_KEYTYPE_ED448';
                 if (defined($keyTypeConstant)) {
                     $result = '';
-                    // Ed25519/Ed448 use algorithm 0 (no separate hash - EdDSA has built-in hash)
+                    // algorithm 0 is used because EdDSA has a built-in hash
                     openssl_sign($message, $result, $this->withPassword()->toString('PKCS8'), 0);
                     $signature = $shortFormat == 'SSH2'
                         ? Strings::packSSH2('ss', 'ssh-' . strtolower($this->getCurve()), $result)
--- a/phpseclib/Crypt/EC/PublicKey.php
+++ b/phpseclib/Crypt/EC/PublicKey.php
@@ -58,13 +58,11 @@ final class PublicKey extends EC implements Common\PublicKey
                 [, $signature] = Strings::unpackSSH2('ss', $signature);
             }
 
-            // Use OpenSSL for Ed25519/Ed448 verification if available (PHP 8.4+)
-            // Ed25519ctx (with context) is not supported by OpenSSL, so skip if context is set
+            // OpenSSL supports Ed25519/Ed448 but not Ed25519ctx (context), so skip if context is set
             if (self::$engines['OpenSSL'] && !isset($this->context)) {
                 $keyTypeConstant = $this->curve instanceof Ed25519 ? 'OPENSSL_KEYTYPE_ED25519' : 'OPENSSL_KEYTYPE_ED448';
                 if (defined($keyTypeConstant)) {
-                    // Ed25519/Ed448 use algorithm 0 (no separate hash)
-                    // Returns 1 if valid, 0 if invalid, -1 on error
+                    // algorithm 0 is used because EdDSA has a built-in hash
                     return openssl_verify($message, $signature, $this->toString('PKCS8'), 0) === 1;
                 }
             }
--- a/tests/Unit/Crypt/EC/KeyTest.php
+++ b/tests/Unit/Crypt/EC/KeyTest.php
@@ -445,13 +445,10 @@ Private-MAC: 8a06821a1c8b8b40fc40f876e543c4ea3fb81bb9
         $this->assertSameNL($expected, $key->toString('libsodium'));
     }
 
-    /**
-     * Test OpenSSL Ed25519 key generation, signing, and verification
-     */
     public function testOpenSSLEd25519(): void
     {
         if (!extension_loaded('openssl') || !defined('OPENSSL_KEYTYPE_ED25519')) {
-            self::markTestSkipped('OpenSSL Ed25519 support is not available (requires PHP 8.4+ with OpenSSL 3.0+).');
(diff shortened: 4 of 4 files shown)


terrafrost:
I like it!

A few thoughts.

**Re: Prioritizing OpenSSL over libsodium**
idk that OpenSSL should be considered a "better" engine than libsodium. Two weak arguments for why it ought not be:

 1. libsodium is considered the "better" engine for most everything else (eg. [Common/SymmetricKey.php::setEngine](https://github.com/phpseclib/phpseclib/blob/3.0.48/phpseclib/Crypt/Common/SymmetricKey.php#L2351)) so why have it be the "better" engine for this as well?
 2. Historically OpenSSL has sometimes had the issue where the header and library versions [don't match](https://github.com/phpseclib/phpseclib/commit/54f0f2131bf13a8f4e968118586ef7cc82f416c5). The code that dealt with that isn't present in the 3.0 branch, however, to my knowledge, libsodium has _never_ had such issues.

That said, I realize that this makes testing the OpenSSL implementation a little more challenging. I guess reflection could be used to change `EC::$engines` or something.

Maybe a better solution would be to introduce a new method - `forceEngine('OpenSSL')` or some such. If you call that method and you then do `EC::createKey('Ed25519')` whilst OpenSSL isn't an option then you get an exception. But reflection seems like it'd be less effort. A hypothetical `forceEngine()` method could be introduced in a later PR.

The `setPreferredEngine()` approach that SymmetricKey.php uses wouldn't work out super well in this case because `EC::createKey()` creates the key as a one step process whereas with classes extending SymmetricKey.php it's a multi step process. eg. `$cipher = new AES('ctr'); $cipher->setKey('...'); $cipher->setIV('...'); $cipher->encrypt('...')`. Like during any of those steps you can do `$cipher->getEngine()` and see if that's the engine that you're hoping that it is and, if not, do `self::markTestSkipped()`.

**getEngine() method needs to be updated**
I haven't tested it but, with the way it's written right now, it looks like the [getEngine() method](https://github.com/phpseclib/phpseclib/blob/3.0.48/phpseclib/Crypt/EC.php#L330) could falsely imply that libsodium is being used for Ed25519 when, in fact, it's OpenSSL.

**Make it so OpenSSL can be used to generate other EC keys?**
It doesn't look like OpenSSL is used to generate any other EC keys. I assume the reason for that is kinda hinted at in the comments of `EC/PrivateKey.php::sign()`:

https://github.com/phpseclib/phpseclib/blob/274295f86709d1ab3f5440a089ae9f7d321247e5/phpseclib/Crypt/EC/PrivateKey.php#L148-L152

Would you be up for implementing support for that? We could do `version_compare(PHP_VERSION, '7.1.0') >= 0 && in_array($curve, openssl_get_curve_names())`, however, some care would need to be taken because phpseclib supports some curve names as aliases of other curves. eg. nistp256 == secp256r1 == prime256v1.

I suppose for that what we could do is add `const OPENSSL_NAME = 'prime256v1'` to [secp256r1.php](https://github.com/phpseclib/phpseclib/blob/master/phpseclib/Crypt/EC/Curves/secp256r1.php) and [nistp256.php](https://github.com/phpseclib/phpseclib/blob/master/phpseclib/Crypt/EC/Curves/nistp256.php). Then what we could do is pass the [short class name](https://www.php.net/manual/en/reflectionclass.getshortname.php) to `in_array()` _unless_ `$curve::OPENSSL_NAME` is defined, at which point, we'd pass that, instead.

**Make this target the 3.0 branch?**
There's no reason this PR can't live in the 3.0 branch. If you were to resubmit it targeting the 3.0 branch some of your changes would need to be removed. eg.

```php
if ($source instanceof Signable) {
    $source->setSignature($signature);
}
```
However, maybe you could submit two PR's? One for the 3.0 branch and one for the master branch that merges the 3.0 changes into the PR for the master branch and that keeps the `$source->setSignature($signature);` stuff?

Thanks!

binaryfire:
@terrafrost Sounds good! Will get onto it as soon as I have some time.

terrafrost:
So I'm working on the changes I proposed now. I squashed your two commits into one, cherry-picked that commit to the the 3.0 branch and I'm now working on removing `getEngine()`, `useBestEngine()` and `useInternalEngine()` and what not and am replacing them with `forceEngine(string $engineName)`. If you try to force an engine that isn't used an exception will be thrown.

It's a BC breaking change for the 3.0 branch but it's hard to imagine anyone actually using those features anyway. They were mainly added to facilitate unit testing. I talk more about phpseclib's versioning strategy here:

https://phpseclib.com/docs/versioning

binaryfire:
> So I'm working on the changes I proposed now. I squashed your two commits into one, cherry-picked that commit to the the 3.0 branch and I'm now working on removing `getEngine()`, `useBestEngine()` and `useInternalEngine()` and what not and am replacing them with `forceEngine(string $engineName)`. If you try to force an engine that isn't used an exception will be thrown.
> 
> It's a BC breaking change for the 3.0 branch but it's hard to imagine anyone actually using those features anyway. They were mainly added to facilitate unit testing. I talk more about phpseclib's versioning strategy here:
> 
> https://phpseclib.com/docs/versioning

Thanks for picking this up. Sorry I haven't been able to get back to this sooner. Life has been getting in the way.

terrafrost:
So I'm still working on this. Among other things, I got this running on 3.0 and am making it so that openssl_sign() / openssl_verify() / openssl_public_encrypt() / etc aren't just used for Ed25519 / Ed448 but for _all_ elliptic curves and for RSA and DSA as well. At least that's the idea.

Turns out it can't use `openssl_verify()` for PKCS1 RSA padding because OpenSSL only supports AlgorithmIdentifier's with explicit NULLs. Quoting https://datatracker.ietf.org/doc/html/rfc4055#page-6 :

>            There are two possible encodings for the AlgorithmIdentifier
>            parameters field associated with these object identifiers.  The two
>            alternatives arise from the loss of the OPTIONAL associated with the
>            algorithm identifier parameters when the 1988 syntax for
>            AlgorithmIdentifier was translated into the 1997 syntax.  Later the
>            OPTIONAL was recovered via a defect report, but by then many people
>            thought that algorithm parameters were mandatory.  Because of this
>            history some implementations encode parameters as a NULL element
>            while others omit them entirely.  The correct encoding is to omit the
>            parameters field; however, when RSASSA-PSS and RSAES-OAEP were
>            defined, it was done using the NULL parameters rather than absent
>            parameters.
> 
>            All implementations MUST accept both NULL and absent parameters as
>            legal and equivalent encodings.

OpenSSL does NOT accept both - it REQUIRES NULL be present. phpseclib, however, DOES accept both. At first, it didn't. At first, not knowing why some small number of PKCS1 signatures omitted NULL, I added the SIGNATURE_RELAXED_PKCS1 mode on 2015-08-26. https://phpseclib.com/docs/rsa#rsasignature_relaxed_pkcs1 talks more about that mode. Later, on 2021-04-05, there was CVE-2021-30130. Consequently, the SIGNATURE_PKCS1 mode was updated to accept either NULL or non-NULL.

See https://security.stackexchange.com/questions/110330/encoding-of-optional-null-in-der for an additional reference.

Other challenges I've run into (and solved):

- `openssl_pkey_new()` needs openssl.cnf. If the host OS has it, cool, but if not, it needs at least a dummy one. phpseclib includes a dummy one but that needed to be updated as did your calls to it. openssl.cnf also needed a few extra lines added to it for it to work (this is the first time that file was modified [in 13 years](https://github.com/phpseclib/phpseclib/blob/master/phpseclib/openssl.cnf)!). like on Linux you probably already have OpenSSL installed and an appropriately configured openssl.cnf, however, on Windows, there's a very good change you _don't_ have it already it "installed". Or rather, the odds are that you don't have an openssl.cnf in the correct directory location ([c]C:\Program Files (x86)\Common Files\SSL\openssl.cnf[/c]).
- on older versions of PHP `openssl_pkey_export()` would return PKCS1-style keys instead of PKCS8-style keys.
- for OpenSSL's new Curve25519 / Curve448 functionality to work with phpseclib, the PKCS8 serializer needed to be updated to accommodate those two curves
- all the unit tests need to be updated
- etc

Thus far I'm at 18 files changed, 1257 insertions(+) and 652 deletions(-). Sufficient to say, this is growing into quiet the beast but I'm plowing through.

terrafrost:
Closing this in favor of #2130, which features your two commits in this PR squashed and cherry picked.

Thank you for your effort on this!

PR #2122, opened 2026-01-20
Add `onFile` callback to SFTP rawlist

mnapoli:
This is another take at #1418 

Essentially listing files over FTP is blocking: if there are hundreds of thousands of files, we can't stop.

In our use case, we want to set a timeout after 60 seconds and a limit of 100 files fetched. This new parameter allows to solve for this by checking if we've reached the limit on each file received.

I understand it's not the absolute perfect solution. In an ideal world, we'd use generators. I did 3 different versions of this change:

1. I followed https://github.com/phpseclib/phpseclib/pull/1418#issuecomment-544261341 and tried to address this via a generator + iterator + arrayaccess to preserve backwards compatibility
    - The diff was quite decent, especially in critical parts of the code.
    - Since there are few unit tests, I felt very uncomfortable doing so many changes, I think this is risky.
    - Also there are flags like sorting that completely break lazy-loading results (we need all results to be retrieved to be able to sort), so it led to very complicated code
2. I tried to add second `rawlist()` method that would be generator-based, while preserving the existing one as-is
    - This was less worse than above, but with more duplicated code.
    - I still had a significant diff because I needed to change the private `rawlist` function, which is used in different places. I ended up using pretty advanced generator features (like `$generator->valid()` and `$generator->getReturn()`) to support the different return statuses (array OR int OR false), and again with so little tests I think those are very risky changes.
3. I ended up with this much much simpler approach: add a callback
    - This doesn't solve the problem of browsing thousands of files without filling the memory, but it allows for setting limits, and processing items as they are retrieved.
    - I think this is a good improvement considering how small the addition is.
    - My personal take, after having tried approach 1 and 2, is that going for a more complex (Generator) approach is attractive, but is actually MUCH riskier and challenging than expected.

Example usage:

```php
$startTime = microtime(true);

$files = $sftp->rawlist('/', true, function () {
    $currentTime = microtime(true);

    if (...timeout...) {
        throw new Exception('Reading files from FTP timed out');
    }
});
```

We could take this PR just a little bit further and:
- if `$onFile` is NOT provided, then the function returns the array as usual
- if `$onFile` IS provided, then the function does NOT store/return the array, so memory is efficient no matter how many files
But I didn't go that far to make the review and merge easier. Happy to revisit that if you prefer.

### DIFF
--- a/tests/Functional/Net/SFTPUserStoryTest.php
+++ b/tests/Functional/Net/SFTPUserStoryTest.php
@@ -792,6 +792,25 @@ class SFTPUserStoryTest extends PhpseclibFunctionalTestCase
     /**
      * @depends testRawlistDisabledStatCache
      */
+    public function testRawlistCallback(SFTP $sftp)
+    {
+        $files = [];
+        $callback = function ($dir, $filename, $attributes) use (&$files): void {
+            $files[] = $filename;
+        };
+
+        $sftp->rawlist(self::$scratchDir, true, $callback);
+
+        $this->assertContains('text.txt', $files);
+        $this->assertContains('subdir', $files);
+        $this->assertContains('leaf.txt', $files);
+
+        return $sftp;
+    }
+
+    /**
+     * @depends testRawlistCallback
+     */
     public function testChownChgrp(SFTP $sftp)
     {
         $stat = $sftp->stat(self::$scratchDir);


mnapoli:
The build fails on master.

terrafrost:
I like the idea! Could you add a unit test for this new feature? Prob would be best to go into https://github.com/phpseclib/phpseclib/blob/master/tests/Functional/Net/SFTPUserStoryTest.php

In-so-far as ideas #1 and #2 are concerned...  I'm a bit worried that generators would break things if multi channel support isn't implemented first.

Like what if you're in the middle of reading the contents of a directory with 100,000 entries and then you decide to open one of the entries. You'd think that you should be able to resume the directory listing from where you left off after having opened the file in question but my theory is that you would not be able to - not until multiple channel support has been implemented.

I had initially thought I'd do that in phpseclib 4.0 but imma push that off, now, to phpseclib 5.0. I'm working on some much needed X509 related changes for phpseclib 4.0 

> The build fails on master.

PHP 8.2 does. Come March, I'll start making the master branch use more fully use more modern PHP features. As part of my work on that it may well be the case that PHP 8.2 support is dropped idk. Rather than take the time to figure out what's up with the unit tests on PHP 8.2 right now I figure I'll deal with it later. My primary push right now (per the milestones that me and the Sovereign Tech Fund out of Germany came up with) is to add support for the [Cryptographic Message Syntax](https://en.wikipedia.org/wiki/Cryptographic_Message_Syntax).

Thanks!

mnapoli:
If I change the base branch of this PR to `3.0` would that allow releasing this feature sooner? Or is the 3.0 branch "closed" to features?

mnapoli:
Thank you for merging!

To clarify, will this be included in a new v3 release?

terrafrost:
The PR was made against the master branch - not the 3.0 branch. If you were to submit a new PR against the 3.0 branch I could consider it.

That said, nullable types couldn't be used in v3 because it wasn't until 7.1 that PHP started supporting them per https://www.php.net/manual/en/migration71.new-features.php .

It used to be the case that you could do `ObjectName $var = null` to make it so that `$var` could be of type ObjectName or null but that [went away in PHP 8.4](https://wiki.php.net/rfc/deprecate-implicitly-nullable-types) so for phpseclib 3.0 (which works on versions of PHP as low as 5.6) you'd have to drop type hinting altogether and instead do `$var instanceof ObjectName || is_null($var)` checks in the function itself or some such.

PR #2123, opened 2026-01-28
add exception SFPT::get

Yazzz28:
Add exceptions to SFTP::get() method instead of returning false

Changes
The `get()` method now throws exceptions instead of silently returning `false`:

| Scenario                                                                                  | Exception |
|------------------------------------------------------------------|-----------|
| Not connected or SFTP not initialized                                   |  `RuntimeException` |
| Remote file path cannot be resolved                                     | `FileNotFoundException` |
| Unable to open remote file (not found, permission denied) | `RuntimeException` |
| Unable to open local file for writing                                      | `RuntimeException`    |
| Connection closed during download                                     | `RuntimeException`   |
| Failed to close remote file handle                                          | `RuntimeException` |

### DIFF
--- a/phpseclib/Net/SFTP.php
+++ b/phpseclib/Net/SFTP.php
@@ -2100,25 +2100,26 @@ class SFTP extends SSH2
     /**
      * Downloads a file from the SFTP server.
      *
-     * Returns a string containing the contents of $remote_file if $local_file is left undefined or a boolean false if
-     * the operation was unsuccessful.  If $local_file is defined, returns true or false depending on the success of the
-     * operation.
+     * Returns a string containing the contents of $remote_file if $local_file is left undefined.
+     * If $local_file is defined, returns true on success.
      *
      * $offset and $length can be used to download files in chunks.
      *
      * @param  string|bool|resource|callable $local_file
-     * @return string|bool
+     * @return string|true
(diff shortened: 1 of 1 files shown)


terrafrost:
I like the idea of replacing `return false` instances with exceptions, however, the PR also includes CS changes that I don't agree with and that kinda dilute the main point of this PR. eg. replacing this:

```php
            $command = "test -x /usr/lib/sftp-server && exec /usr/lib/sftp-server\n" .
                       "test -x /usr/local/lib/sftp-server && exec /usr/local/lib/sftp-server\n" .
                       "exec sftp-server";
```
with this:

```php
            $command = "test -x /usr/lib/sftp-server && exec /usr/lib/sftp-server\n" .
                "test -x /usr/local/lib/sftp-server && exec /usr/local/lib/sftp-server\n" .
                "exec sftp-server";
```
I don't think that's an improvement.

Can you modify your PR to exclude those kinds of changes? If you still think that they ought to be made a separate PR that focuses on CS style changes and that presents a rationale for them can be made.

Thanks!

Yazzz28:
@terrafrost Thanks for the feedback!

I’ve updated the PR to focus only on replacing return false with exceptions and removed the unrelated CS-style changes you mentioned.
Thanks again for taking the time to review this.

PR #2142, opened 2026-06-01
Feature: SFTP Hardlink

srusche:
I would like to add this method to enable creating hardlinks via SFTP.

My use case is a remote storage server where I found it is possible to create hardlinks via SFTP extensions command, while ```ln``` command is missing and ```cp -l``` fails for non-admin accounts.

In the past I had a similar task in a Node backend and used [ext_openssh_hardlink](https://github.com/mscdex/ssh2/blob/318d447ce3aca26e1ac73b63767b82a29b02467b/lib/protocol/SFTP.js#L1426) from the [ssh2](https://www.npmjs.com/package/ssh2) node.js library, which is why I knew it is possible.

### DIFF
--- a/phpseclib/Net/SFTP.php
+++ b/phpseclib/Net/SFTP.php
@@ -3825,9 +3825,12 @@ class SFTP extends SSH2
             return false;
         }
 
-        // don't move the stat cache entry over since this operation could very well change the
-        // atime and mtime attributes
+        // this operation will change the ctime and link-count attributes
+        // which could be cached depending on sftp version
         $this->remove_from_stat_cache($oldpath);
+
+        // hardlink creation should fail if $newpath exists;
+        // removing it from the cache anyway, just to be sure
         $this->remove_from_stat_cache($newpath);
 
         return true;
--- a/phpseclib/Net/SFTP.php
+++ b/phpseclib/Net/SFTP.php
@@ -3789,4 +3789,47 @@ class SFTP extends SSH2
             Strings::unpackSSH2('QQQQQQQQQQQ', $response)
         );
     }
+
+    public function hardlink($oldpath, $newpath)
+    {
+        if (!$this->precheck()) {
+            return false;
+        }
+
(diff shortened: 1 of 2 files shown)


terrafrost:
So the change looks good altho I am curious about one thing:

```php
        // don't move the stat cache entry over since this operation could very well change the
        // atime and mtime attributes
        $this->remove_from_stat_cache($oldpath);
        $this->remove_from_stat_cache($newpath);
```
What's up with that? `symlink()` doesn't do that but `hardlink()` does?

Also, a unit test would be good! We have unit tests for other OpenSSH SFTP extensions, like testStatVfs and testPosixRename in tests/Functional/Net/SFTPUserStoryTest.php

srusche:
> So the change looks good altho I am curious about one thing:
> 
> ```
>         // don't move the stat cache entry over since this operation could very well change the
>         // atime and mtime attributes
>         $this->remove_from_stat_cache($oldpath);
>         $this->remove_from_stat_cache($newpath);
> ```
> 
> What's up with that? `symlink()` doesn't do that but `hardlink()` does?

Creating a hardlink changes the metadata of the file, as it increases the number of links and updates the "change" time ("change" refers to the metadata, "modify" to the file's content). Hardlink and original file share the same metadata, as it is stored in the common inode.

So I need to update the comment: the operation changes the *ctime* and *link-count* attributes. Depending on sftp version these may be included in stat cache and thus removing $oldpath is fine. That part originates from the copy method and I guess that copy will overwrite $newpath if it exists. I am not sure that hardlink creation always fails if $newpath exists; so the second line may not be necessary but also should not harm, that's why I kept it.

A symlink is a new separate file with its own metadata, which contains the linked file's path. Removing $oldpath should not be necessary here as it does not touch the original file. Again, if symlink creation always fails when $newpath exists, there is no need to remove $newpath from stat cache either.

> Also, a unit test would be good! We have unit tests for other OpenSSH SFTP extensions, like testStatVfs and testPosixRename in tests/Functional/Net/SFTPUserStoryTest.php

I will add that.

PR #2153, opened 2026-07-09
Don't delegate PSS signing to openssl_sign()

### LINKED ISSUE #2152: RSA::sign() with SIGNATURE_PSS produces an incompatible salt length on PHP >= 8.5 (via openssl_sign OpenSSL engine)
### Problem

On PHP >= 8.5, `openssl_sign()` gained support for `OPENSSL_PKCS1_PSS_PADDING`. `RSA::sign()` auto-detects this and delegates PSS signing to `openssl_sign()` instead of phpseclib's own `rsassa_pss_sign()`.

The problem: `openssl_sign()` has no parameter to control the PSS salt length. When phpseclib delegates to it for PSS, OpenSSL silently uses its own default salt length, which is the **maximum** possible, not `hLen` (i.e. not matching `$rsa->getSaltLength()`, which defaults to the hash length).

This produces a signature that is a mathematically valid RSASSA-PSS signature per RFC 8017 (any salt length is technically legal), but it silently breaks compatibility with any consumer that requires the salt length to equal the hash length - most notably RFC 7518 (JOSE) PS256/PS384/PS512, which mandate a fixed salt length equal to the hash output length. Every JOSE/JWT library (and phpseclib's own PHP-engine PSS verifier) rejects the resulting signature.

Note that `openssl_verify()` is not affected the same way: it auto-recovers whatever salt length is actually embedded in a signature, so verification via the OpenSSL engine happens to still succeed against these malformed signatures. This masks the bug in a phpseclib-only sign+verify round trip and only surfaces when verifying with a strict, spec-compliant PSS verifier that assumes a fixed salt length.

### How to reproduce

Uses only phpseclib. Signs with the default engine, then forces the pure-PHP engine for verification only, to show the signature fails phpseclib's own PSS verification:

```php
<?php

declare(strict_types=1);

require __DIR__.'/vendor/autoload.php';

use phpseclib3\Crypt\RSA;

$privateKey = RSA::createKey()
    ->withHash('sha256')
    ->withMGFHash('sha256')
    ->withPadding(RSA::SIGNATURE_PSS)
;
$publicKey = $privateKey->getPublicKey();
$message = 'Testing';

// On PHP >= 8.5 this silently uses openssl_sign(), which has no parameter to control the PSS
// salt length and defaults to the *maximum* salt length instead of hLen.
$signature = $privateKey->sign($message);

// Force the pure-PHP engine for verification. This is what any strict, spec-compliant
// PSS verifier does (fixed salt length, no auto-detection)
// e.g. RFC 7518 PS256 as implemented by every JOSE/JWT library.
RSA::forceEngine('PHP');

$verified = $publicKey->verify($message, $signature);

printf("PHP version: %s\n", PHP_VERSION);
printf("Configured salt length: %d\n", $privateKey->getSaltLength());
printf("Verified: %s\n", $verified ? 'true' : 'false');
```

**Actual result**

```
PHP version: 8.5.8
Configured salt length: 32
Verified: false
```

**Expected result**

`Verified: true` - a signature produced with `getSaltLength() === 32` should actually be encoded with a 32-byte salt, and should verify successfully against a verifier that enforces that salt length (as PS256/RFC 7518 requires).

This same script prints `Verified: true` when `RSA::forceEngine('PHP')` is called **before** `sign()` too (forcing the PHP engine for both operations), confirming the pure-PHP PSS implementation is correct and the bug is specific to the `openssl_sign()` delegation path.

### Environment

* phpseclib version: 3.0.55
* PHP version: 8.5.8
* OpenSSL version (`OPENSSL_VERSION_TEXT`): OpenSSL 3.0.2 15 Mar 2022
* OS: Linux

### Suggested fix

Exclude `openssl_sign` from the OpenSSL delegation path for `SIGNATURE_PSS` in `handleOpenSSL()`, since it provides no way to guarantee the salt length matches what was configured. `openssl_verify` does not need the same treatment since it auto-recovers the actual embedded salt length regardless of value.

```php
if ($this->$paddingType === self::SIGNATURE_PSS) {
    switch (true) {
        case !defined('OPENSSL_PKCS1_PSS_PADDING'):
            $error = 'Engine OpenSSL is forced but PSS encryption requires PHP >= 8.5.0';
            break;
        case $this->hash->getHash() !== $this->mgfHash->getHash():
            $error = 'Engine OpenSSL is forced but can\'t be used because the Hash and MGF Hash do not match';
            break;
        case $this->getSaltLength() !== $this->hLen:
            $error = 'Engine OpenSSL is forced but can\'t be used because the salt length doesn\'t match the hash length';
            break;
        // New code:
        case $func === 'openssl_sign':
            $error = 'Engine OpenSSL is forced but can\'t be used for PSS signing because openssl_sign() has no way to control the PSS salt length and defaults to the maximum length instead of the hash length';
            break;
    }
}
```

tribe-lis:
Fixes #2152.

## Changes

* `phpseclib/Crypt/RSA.php`: in `handleOpenSSL()`, reject delegating PSS signing to `openssl_sign()` in two cases:
   * the installed OpenSSL is older than 3.1.0.
   * the key is too small to fit the configured salt length.

### DIFF
--- a/phpseclib/Crypt/RSA.php
+++ b/phpseclib/Crypt/RSA.php
@@ -974,6 +974,13 @@ abstract class RSA extends AsymmetricKey
                         break;
                     case $this->getSaltLength() !== $this->hLen:
                         $error = 'Engine OpenSSL is forced but can\'t be used because the salt length doesn\'t match the hash length';
+                        break;
+                    case 'openssl_sign' && $this->getLength() < 8 * (2 * $this->getSaltLength() + 2):
+                        $error = 'Engine OpenSSL is forced but can\'t be used for PSS signing because the key is too small for OpenSSL to use the configured salt length';
+                        break;
+                    case 'openssl_sign' && OPENSSL_VERSION_NUMBER < 0x30100000:
+                        $error = 'Engine OpenSSL is forced but can\'t be used for PSS signing because OpenSSL < 3.1.0 defaults to the maximum salt length instead of the hash length';
+                        break;
                 }
             }
             /*
--- a/tests/Unit/Crypt/RSA/ModeTest.php
+++ b/tests/Unit/Crypt/RSA/ModeTest.php
@@ -145,6 +145,68 @@ p0GbMJDyR4e9T04ZZwIDAQAB
         RSA::forceEngine();
     }
 
+    public function testPSSSignAndVerifyAcrossEngines()
+    {
+        $privatekey = RSA::createKey(1024)
+            ->withHash('sha256')
+            ->withMGFHash('sha256')
+            ->withPadding(RSA::SIGNATURE_PSS);
+        $publickey = $privatekey->getPublicKey();
+
+        $message = 'zzzz';
+        $signature = $privatekey->sign($message);
+
+        RSA::forceEngine('PHP');
+        try {
+            $this->assertTrue($publickey->verify($message, $signature));
+        } finally {
+            RSA::forceEngine();
+        }
+    }
+
+    public function testPSSSignWithForcedOpenSSLEngineFailsOnSmallKey()
+    {
+        if (!defined('OPENSSL_PKCS1_PSS_PADDING')) {
+            $this->markTestSkipped('OPENSSL_PKCS1_PSS_PADDING is not defined (requires PHP >= 8.5)');
+        }
+
+        $privatekey = RSA::createKey(512)
+            ->withHash('sha256')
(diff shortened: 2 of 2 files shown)


terrafrost:
Could you update this to target the 3.0 branch instead of the master branch?

Thanks!

tribe-lis:
I changed the target branch from master to 3.0. Can you review changes?

PR #2160, opened 2026-08-06
X509: add support for the logotype extension (RFC 9399)

chaosben:
I came to this from BIMI: I needed to read Verified Mark Certificates, which carry their logo in the RFC 9399 logotype extension. phpseclib currently lists `id-pe-logotype` among the OIDs that are known but unsupported, so the extension comes back as an opaque `Element` and you are on your own from there. I had the maps written as a `registerExtension()` mapping in my own project for a while, and it seemed a shame to keep them there when they cover the whole RFC.

This is the follow-up to #2159. That fix is the reason this is possible at all — `issuerLogo [1]` and `subjectLogo [2]` share the same CHOICE definition, so nothing but the context tag distinguishes them, and before the fix every certificate I had was reported as `issuerLogo`.

### What is in it

- 13 maps under `phpseclib/File/ASN1/Maps/`, one per structure in Appendix A.1: `LogotypeExtn`, `LogotypeInfo`, `LogotypeData`, `LogotypeImage`, `LogotypeAudio`, `LogotypeDetails`, `LogotypeImageInfo`, `LogotypeImageType`, `LogotypeImageResolution`, `LogotypeAudioInfo`, `OtherLogotypeInfo`, `LogotypeReference`, `HashAlgAndValue`. I inlined the unnamed `SEQUENCE OF` wrappers with `min`/`max` rather than giving them classes, since the RFC does not name them and `GeneralNames` only has a class because RFC 5280 does.
- `id-logo-loyalty`, `id-logo-background` and `id-logo-certImage` in `OIDs/X509.php`, and `id-pe-logotype` moved out of the "not supported" block. Without those three, `otherLogos` carries a bare OID and the slot has no discernible purpose.
- `'id-pe-logotype' => Maps\LogotypeExtn::MAP` in the Extension trait.
- `tests/Unit/File/X509/LogotypeTest.php`.

### Bits I would look at twice

- **`type` is `DEFAULT color`.** It has to vanish from the encoding when it holds that value. `testImageTypeDefaultsToColor` pins both directions.
- **Tagging is mixed.** The module is `DEFINITIONS IMPLICIT TAGS`, yet all four `LogotypeExtn` slots are spelled `EXPLICIT`, and `communityLogos [0]` / `otherLogos [3]` wrap a `SEQUENCE OF` rather than a single logo. Easy to get subtly wrong, so it is worth a glance.
- **`resolution`** is an untagged CHOICE child whose alternatives carry `[1]`/`[2]`. Different case from #2159 and untouched by it, but I want to flag it so nobody assumes the fix is load-bearing there.

### Testing

Full `tests/Unit` on PHP 8.5: no failures, with deprecation and notice counts unchanged against the branch without this. phpcs and psalm are both clean.

Three real Verified Mark Certificates are included as fixtures — eBay, Badoo and Rabobank, from three different issuing CAs (DigiCert, Sectigo, and DigiCert's European CA) so the shape is not one vendor's habit. Each is checked for a non-critical extension resolving to `subjectLogo`, an `image/svg+xml` logo embedded as a gzipped `data:` URL, and a byte-identical round trip. The two inline DER vectors in the test file were cross-checked against `openssl asn1parse`, so the suite is not merely agreeing with itself.

As a sanity check on whether the tests actually bite, I reverted the #2159 fix locally: six of them go red, `subjectLogo` among them.

### What I left out, and three things I am happy to add

Maps and OIDs only. RFC 9399 is mostly a behavioural specification and a map cannot express any of it — hash computation and verification (§4.1), permitted URI schemes, the media type ABNF, RFC 5646 language tags, the SVG Tiny profile and the GZIP requirement (§7), "at most one background logotype" (§4.4.2), and the client-side rules in §6.

I left all of that alone. phpseclib does enforce extension semantics where a trust decision depends on it — `keyUsage` and `basicConstraints` during path validation — but RFC 9399 §6 rules that out here in as many words: "The relying party MUST NOT use the logotypes as part of certification path validation or automated trust decisions." So none of these rules can ever feed a decision the library makes, which puts them squarely in the application's court.

Also absent: resolving `indirect` references over the network, unpacking `data:` URLs, and attribute certificates — RFC 9399 allows the extension there (§4.1), but there is no attribute certificate entry point running `mapInExtensions`.

Three things I would genuinely like to contribute if you want them, in whatever combination suits you:

1. **The criticality rule.** "This extension MUST NOT be marked critical" (RFC 9399 §4.1) is one condition and the only hard requirement that is purely local to the extension.
2. **"At least one field present"** for `LogotypeExtn` and `LogotypeData`. Appendix A.2 states it formally as a `WITH COMPONENTS` constraint; A.1 only says it in prose.
3. **Convenience helpers, in both directions** — and I think the writing side is the more useful half.

   On the way in, something like `addSubjectLogo($mediaType, $uri, $imageData)`, alongside the issuer and community variants: build the nested structure, compute `logotypeHash`, and set the extension non-critical. The hash is the part worth taking off the caller: RFC 9399 §4.1 wants it computed over the whole referenced object using the one-way function associated with the certificate signature, and if it is wrong, §4.1 instructs clients to discard the logotype data. The certificate then looks perfectly healthy while the logo silently does not work. `createSubjectKeyIdentifier()` already has exactly this shape — hash something, write the result into an extension — so the pattern would not be new.

   On the way out, accessors that hand back a logo's media type, URIs and hashes without walking `['subjectLogo']['direct']['image'][0]['imageDetails']`. Less essential, since the mapped structure is navigable as it stands, but it is what I ended up writing on top in my own code.

   I left both out over a distinction I am not sure how you see it. There is plenty of precedent for extension-specific helpers — `addDomains()` and `addIPAddresses()` for `subjectAltName`, `makeCA()` for `keyUsage` and `basicConstraints`, `createKeyIdentifier()` for SKI and AKI, `validateURL()` reading `subjectAltName` on the way back out. But every one of them exists because phpseclib itself needs it, to build a usable certificate or to reach a validation decision. Logotype helpers would be the first that exist purely for the caller, since the library never decides anything based on a logo. That felt like your call rather than mine, and if the answer is no, the maps stand perfectly well on their own.

### Happy to do 3.0 as well

If you want this on 3.0, say so and I will open it there too. I checked the things that could have made it awkward, and none of them do:

- The map array format is identical, so the 13 maps carry over as they are. What changes is cosmetic — the `PHP version 5` header, `const` instead of `public const`, no trailing commas.
- `DEFAULT` behaves the same. `encode_der()` drops a value equal to `$mapping['default']` before it does anything else, and the strict comparison works fine against the mapped `'color'` string.
- #2158 already put the tagged CHOICE fix on that branch, so `issuerLogo` and `subjectLogo` are distinguishable there too.
- The wiring lands in different places: the OIDs in the `$oids` array in `X509.php` rather than `OIDs/X509.php`, and the mapping in the `getMapping()` switch rather than the trait. The tests assert on plain arrays out of `asn1map()` instead of the 4.0 types.

The `registerExtension()` note above does not apply on 3.0, incidentally — that version only compares against `self::$extensions`, so a built-in mapping does not get in the way.

One behavioural note for this branch: code that today calls `X509::registerExtension('id-pe-logotype', …)` will get an `InvalidStateException` afterwards, since a built-in mapping now exists. The call just becomes unnecessary — my own code was the main thing doing it.


terrafrost:
I like it!

Would it be possible to include a few full X509 certs that include the logotype extension? Like not generated ones but real world ones, even if they're expired? I've never seen the logotype extension used in a real world cert before and am intrigued by it!

chaosben:
Thank you — glad it lands well!

Three real ones are in now, and currently valid rather than expired, from three different issuing CAs so you can see the structure is not one vendor's habit: eBay (DigiCert), Badoo (Sectigo) and Rabobank (DigiCert's European CA). There is a test over them checking each resolves to `subjectLogo`, that the extension is not marked critical, and that the certificate re-encodes byte for byte. No assertion on the validity dates, so nothing goes red when they expire in 2027.

**Fetching your own** is the fun part, and it means you never have to wait on me for a sample. These are BIMI certificates — BIMI is what puts a sender's logo beside their message in a mail client, and the certificate is what backs that logo with a registered trademark. It is all published over DNS:

```
$ dig +short default._bimi.dhl.de TXT
"v=BIMI1; l=https://www.dhl.com/dhl-email-logo/dhl-bimi.svg; a=https://www.dhl.com/dhl-email-logo/dhl-vmc_20250916.pem"
```

The `a=` URL is the certificate, leaf plus chain:

```
$ curl -s https://www.dhl.com/dhl-email-logo/dhl-vmc_20250916.pem -o dhl.pem
$ openssl x509 -in dhl.pem -noout -ext 1.3.6.1.5.5.7.1.12
```

Domains with one today: `dhl.de`, `ebay.com`, `paypal.com`, `cnn.com`, `bankofamerica.com`, `adobe.com`, `linkedin.com`, `bbc.co.uk`, `rabobank.nl`, `badoo.com`. A `default._bimi` TXT record is the tell, and plenty of large senders have one.

Three things I found genuinely surprising while working with these:

**The logo lives inside the certificate rather than being linked from it.** Every one I have seen uses a `data:image/svg+xml;base64,…` URI instead of an external URL, and the payload is gzipped — RFC 9399 §7 requires that for embedded SVG. On the DHL certificate it is 750 bytes compressed and 1406 bytes of SVG Tiny once inflated; Rabobank's inflates to 31 KB. With this PR, getting it out is two lines:

```php
$details = $ext['extnValue']['subjectLogo']['direct']['image'][0]['imageDetails'];
[, $b64] = explode(',', (string) $details['logotypeURI'][0], 2);
file_put_contents('logo.svg', gzdecode(base64_decode($b64)));
```

**The subject DN carries the trademark registration itself.** Those `1.3.6.1.4.1.53087.x` attributes are BIMI's: `.1.13` is `Registered Mark`, `.1.2` names the office that granted it (`European Union Intellectual Property Office`, `UK Intellectual Property Office`, `United States Patent and Trademark Office`), and `.1.4` is the registration number. phpseclib prints them as raw OIDs today. Happy to add them too, though that is a different spec and probably belongs in its own PR.

**`logotypeHash` is not decorative.** RFC 9399 §4.1 tells clients to discard the logotype data outright if the hash does not match what was fetched — so a certificate can look perfectly healthy while the logo silently never appears. That is the reason the write-side helper I offered in the description would compute the hash rather than leave it to the caller.

The offers in the description all still stand, including the 3.0 backport — no hurry on any of them.

chaosben:
Thank you for the time you put into this one — the review, the merge, and the psalm annotation you tidied up on my helper afterwards. I appreciate it.

If there is appetite for more in this corner of the library, I would be glad to pick up any of these:

1. **A 3.0 backport.** The maps carry over unchanged; only the wiring moves — OIDs into the `$oids` array, the mapping into the `getMapping()` switch. #2158 already put the tagged CHOICE fix on that branch, so the groundwork is there.
2. **A write-side helper**, something like `addSubjectLogo($mediaType, $uri, $imageData)`. The useful part is `logotypeHash`: RFC 9399 §4.1 wants it computed with the hash function tied to the certificate's signature, and if it is wrong, clients drop the logo without complaining. `createSubjectKeyIdentifier()` already has that shape.
3. **Read-side accessors**, so callers do not have to walk `['subjectLogo']['direct']['image'][0]['imageDetails']` by hand.
4. **The two rules that are actually enforceable**: the extension must not be marked critical (RFC 9399 §4.1), and at least one field has to be present in `LogotypeExtn` and `LogotypeData` (Appendix A.2 states it as a `WITH COMPONENTS` constraint).
5. **The BIMI mark OIDs** — `1.3.6.1.4.1.53087.x`, the trademark registration attributes in these certificates' subject DNs. They print as raw OIDs today.

Whatever is useful to you — and if none of it is, that is a perfectly good answer too.

terrafrost:
> **The two rules that are actually enforceable:** the extension must not be marked critical (RFC 9399 §4.1)

In v4 the extension critical thing should be as simple as modifying `Extension::getExtensionCriticalValue()`.

In v3, if this were to be backported to that, I'd say there's not a ton of point because it doesn't even attempt to auto set the criticality level.

> at least one field has to be present in LogotypeExtn and LogotypeData (Appendix A.2 states it as a WITH COMPONENTS constraint).

phpseclib doesn't really enforce that for distinguished names. Like quoting https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.4 :

```
   The issuer field identifies the entity that has signed and issued the
   certificate.  The issuer field MUST contain a non-empty distinguished
   name (DN).
```
v4, in particular, does that by default. eg. `echo new X509();` will have an empty subject DN. And idk that I want to make them required, either, because then that makes the API more, overall, complicated to use.

> **The BIMI mark OIDs** — 1.3.6.1.4.1.53087.x, the trademark registration attributes in these certificates' subject DNs. They print as raw OIDs today.

How many OIDs are we talking about?

> **A write-side helper,** something like addSubjectLogo($mediaType, $uri, $imageData). The useful part is logotypeHash: RFC 9399 §4.1 wants it computed with the hash function tied to the certificate's signature, and if it is wrong, clients drop the logo without complaining. createSubjectKeyIdentifier() already has that shape.
>
> **Read-side accessors,** so callers do not have to walk ['subjectLogo']['direct']['image'][0]['imageDetails'] by hand.

4.0.0 is due to be released in a few weeks. Any new feature like this would benefit from documentation and the documentation I have for v4 isn't in a public git repo atm so altho I do think that'd be neat it's a little premature I'd say. Like ideally you'd add it to the documentation so that I have more time to focus on the 4.0.0 release but right now you can't because the repo isn't public.

Also, those two prob ought to go hand in hand. Like if you do one you ought to do the other.

> **A 3.0 backport.** The maps carry over unchanged; only the wiring moves — OIDs into the $oids array, the mapping into the getMapping() switch. https://github.com/phpseclib/phpseclib/pull/2158 already put the tagged CHOICE fix on that branch, so the groundwork is there.

You're call on that but I'd say let's hold off on that, too, for the time being. Like if that feature is added to v3 I'd need to update phpseclib3_compat. Like just as how phpseclib_compat let you emulate phpseclib v2 with v3, phpseclib3_compat will let you emulate v3 with v4.

That said, I don't have an ETA for phpseclib3_compat's release, yet. Like I do have an ETA for phpseclib 4.0.0 but not phpseclib3_compat 1.0.0.

I mean, if someone comes along and asks for that feature to be incorporated into 3.0 then it can be done, regardless of phpseclib3_compat's status and I'll just deal with it, but baring that, I see no need to rush!

chaosben:
Thanks for going through all of it point by point — that was more than I expected.

### The OIDs

Twelve under `1.3.6.1.4.1.53087`, and the arc looks settled rather than growing. Eleven are subject DN attributes, one is a policy identifier.

Registered marks, which is what the certificates now in the test suite use:

| OID | Attribute |
|---|---|
| `.1.2` | `trademarkOfficeName` |
| `.1.3` | `trademarkCountryOrRegionName` |
| `.1.4` | `trademarkRegistration` |
| `.1.5` | `legalEntityIdentifier` |
| `.1.6` | `wordMark` |
| `.1.13` | `markType` |

Government marks, used by the statute-based certificates:

| OID | Attribute |
|---|---|
| `.3.2` | `statuteCountryName` |
| `.3.3` | `statuteStateOrProvinceName` |
| `.3.4` | `statuteLocalityName` |
| `.3.5` | `statuteCitation` |
| `.3.6` | `statuteURL` |

Plus `.1.1`, the VMC policy identifier, which is a certificate policy rather than a DN attribute.

These are less of a special case than they look: `.1.2` through `.1.5` are the successors to `1.3.6.1.4.1.311.60.2.1.4` through `.7`, and you already carry `.1` through `.3` of that same arc as `jurisdictionLocalityName` and its two siblings. Adding these would round that out.

Scope is small. Strictly it is twelve lines in `OIDs/X509.php` and nothing else — `translateDNProp()` passes raw OIDs through, and the short-name `match` in `DN.php` already falls back to the trailing segment, so both cope without changes. Input aliases and short display names there would be polish rather than necessity. It also needs no documentation, so it would not add to the 4.0.0 pile — though I am happy to hold it until after the release if you prefer.

### Criticality

Let me make sure I am aiming at the right target here. `getExtensionCriticalValue()` already returns `false` for anything unlisted, so `id-pe-logotype` comes out non-critical today; an entry would state the intent without changing behaviour. What it would not do is stop `setExtension('id-pe-logotype', $value, true)`, because the default only applies when the third argument is null.

So: do you want just the explicit entry, or a guard that makes the RFC's "MUST NOT" actually unoverridable? The first is a one-liner. The second touches `setExtension()`, which I would not go near without you asking for it.

### At least one field

Withdrawn — your DN comparison settles it. If an empty subject DN is acceptable output despite RFC 5280, then a required-field check on this one extension would be an odd exception, and not worth making the API more awkward for.

### Helpers and the 3.0 backport

Understood on both, and the release should come first. Once the v4 documentation repo is public I would be glad to write the docs for the read and write helpers as part of the same piece of work, rather than handing you a feature and leaving the documentation behind. Same for 3.0 — I will leave it be unless someone actually asks, and `phpseclib3_compat` is a good reason not to create work for yourself now.

terrafrost:
A PR to add the OIDs would be great!

> So: do you want just the explicit entry, or a guard that makes the RFC's "MUST NOT" actually unoverridable? The first is a one-liner. The second touches setExtension(), which I would not go near without you asking for it.

Just the explicit entry. Like you said, "_an entry would state the intent_".

Thanks!

PR #2164, opened 2026-08-16
fix: pss_verify fail from go signed data

deatil:
some pss verify fail from other lang sign, so this is fix verify.

### DIFF
--- a/phpseclib/Crypt/RSA/PublicKey.php
+++ b/phpseclib/Crypt/RSA/PublicKey.php
@@ -33,6 +33,18 @@ final class PublicKey extends RSA implements Common\PublicKey
     use Common\Traits\Fingerprint;
 
     /**
+     * PSS find salt
+     */
+    protected bool $pssSaltLengthAuto = true;
+
+    public function withPssSaltLengthAuto(bool $status = true)
+    {
+        $this->pssSaltLengthAuto = $status;
+
+        return $this;
+    }
+
+    /**
      * Exponentiate
      */
     private function exponentiate(BigInteger $x): BigInteger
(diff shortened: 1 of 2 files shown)


phpseclib/Crypt/RSA/PublicKey.php
```diff
@@ -137,6 +153,13 @@ private function emsa_pss_verify(string $m, string $em, int $emBits): bool
         $dbMask = $this->mgf1($h, $emLen - $this->hLen - 1);
         $db = $maskedDB ^ $dbMask;
         $db[0] = ~chr(256 - (1 << ($emBits & 7))) & $db[0];
+        if ($this->pssSaltLengthAuto) {
+            $pslen = mb_stripos($db, chr(0x01));
```
terrafrost: Just do `strpos` vs `mb_stripos`.

First, `mb_stripos` is case insensitive, which isn't needed, here, since `chr(0x01)` most likely doesn't have a case in any language.

Second, even `mb_strpos` is overkill. Like even within this function, it's doing `substr` vs `mb_substr`. You do the `mb_*` functions when you need stuff that works on characters vs bytes. Consider this bit of code:

```php
$a = '淀川水系';
echo strlen(substr($a, 1, 1)) . ',';
echo strlen(mb_substr($a, 1, 1));
```
That outputs `1,3`. ie. the answers don't match because `$a` is a unicode string and substr() is working at the byte level whereas mb_substr() is working at the character level.

For cryptography purposes, it's the byte level we want to be working at.

terrafrost: Also, do `$psLen` vs `$pslen`. Every other length variable in this method is camel case so this one might as well be camel case as well.

terrafrost:
Thanks for this - you found a real bug. The PHP and OpenSSL engines should either have matching output or throw an Exception and your work showed that the existing implementation wasn't doing that. Like the PHP engine _can_ validate the signature in your unit test but it can only do so if you manually call `withSaltLength()` with a non-intuitive value of 222, which isn't at all obvious (`($private->getLength() - $private->getHash()->getLength() - 16) >> 3` === 222).

I wound up going in a different direction with the fix:

https://github.com/phpseclib/phpseclib/commit/f88b1052146e3370bcdab76ce6db17beba23b004

Two things to be mindful of for future PRs:

1. Don't use mbstring functions in crypto code. Consider the following:

    ```php
    $str = '淀川水系';
    echo strlen($str);     // outputs 12
    echo mb_strlen($str);  // outputs 4
    ```

    The take away here is that mbstring operates on characters whereas PHP's regular string functions operate on bytes. This is important because crypto always operates on bytes. `mb_strpos()` on binary data will silently return the wrong offset. Use `strpos()`. phpseclib also deliberately avoids depending on ext-mbstring.

2. If you could squash before opening a PR, that'd help — sixteen commits all named "fix" makes the history hard to read later.

Thanks!

deatil:
@terrafrost good.

PR #2169, opened 2026-08-27
Fix typo

touchweb-vincent:
Hello,

There is probably a typo here.

Thanks

### DIFF
--- a/phpseclib/Crypt/Common/Formats/Keys/PKCS8.php
+++ b/phpseclib/Crypt/Common/Formats/Keys/PKCS8.php
@@ -171,7 +171,7 @@ abstract class PKCS8 extends PKCS
                 throw new UnexpectedValueException('Human readable string claims private key but DER encoded string claims public key');
             }
 
-            if ("$public[publicKey]"[0] != "\0") {
+            if ($public['publicKey'][0] != "\0") {
                 throw new UnexpectedValueException('The first byte of the public key should be null - not ' . bin2hex($public['publicKey'][0]));
             }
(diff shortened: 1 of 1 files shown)


terrafrost:
So that's actually intentional. If you do `print_r($public)` just before that line you'll get this:

```
Array
(
    [publicKeyAlgorithm] => Array
        (
            [algorithm] => phpseclib4\File\ASN1\Types\OID Object
                (
                    [value] => 1.2.840.10045.2.1
                    [name] => id-ecPublicKey
                )

            [parameters] => phpseclib4\File\ASN1\Types\OID Object
                (
                    [value] => 1.2.840.10045.3.1.7
                    [name] => secp256r1
                )

        )

    [publicKey] => phpseclib4\File\ASN1\Types\BitString Object
        (
            [value] => 0004942e71839ff875a2b56e5190216cecd7e26b31af423d08b22a03d023862ee9a73f85e68583faa18c2d6678196b0868cf45925ef31472a9a047988c5c96f11aba
        )

)
```
Note how `$public['publicKey']` isn't a string but, instead, a BitString object. In theory, the code change you made could still work if BitString implemented ArrayAccess, which it does, but if you're gonna access it as an array then the mapped value needs to be set, as it is for [KeyUsage](https://github.com/phpseclib/phpseclib/blob/master/phpseclib/File/ASN1/Maps/KeyUsage.php).

So like what the code is doing, as is, is forcing `__toString()` to be called on the object, producing a string the first character of which can then be accessed.

